PR #182, opened 2014-09-24
Add Benchmarks & Split/Unroll-scopes optimizations (x1.6 faster)

ankostis:
For issue #158:

1. ``benchmark`` branch: Added a benchmark with meta_schemas cheking & extensive sample-schema with references.
2. Implemented two optimizations:
    2.a. ``split_scopes`` branch: Keep URIs in 2 parts (url, fragment) to avoid fragging/defragging since fragments do not join, they always override (~1.1 faster)
    2.b. ``unroll_scopes`` branch:  Replace resolver.in_scope() context-manager with a push/pop stack of scopes, that does nothing it when no ``id`` property exists (~1.5 faster).
3. ``split_unroll_scopes`` branch: Combined both optimizations (~1.6 faster).
3. For future reference, the timed-results are engraved on the respective methods of ``wltp/test/test_benchmarks.py`` TC (unfortunately they are host-specific).

ALL TCs RUN AS OK (apart from split_frags which was post-edited on the combined branch)

Issue #158, opened 2014-04-14
Performance Problems


wizzat:
Hey,

So I've been using jsonschema for basic json validation of events before processing.  However, one thing that I've noticed is that urlparse and urljoin are consistently among the hot spots in my code base.  After tracing it down I've found that a lot of it comes from RefResolver.resolving and RefResolver.resolving.

It seems like a couple of optimizations could be really handy:

    @contextlib.contextmanager
    def resolving(self, ref):

        if ref == '#':
            yield self.store[self.base_uri]
        else:
            ...

I'm not entirely sure how to go about handling in_scope, but in my application it seems like resolution scope always becomes the raw scope.

    @contextlib.contextmanager
    def in_scope(self, scope):
        old_scope = self.resolution_scope
        self.resolution_scope = scope
        try:
            yield
        finally:
            self.resolution_scope = old_scope

The next spot that seems to take up a bunch of time is in iter_errors.  There aren't any validation errors, but this seems like it's just a matter of applying the schema to the object and detecting any errors.  However, in several places it seems like it's wrapped in a list, which means you lose the ability to grab the first error and bail out.

PR #184, opened 2014-09-24
Optimize: Allow rules to break error-loop (x2 faster)

ankostis:
An improvements ON TOP of #182 optimizations currently underway, all dealing with the #158 issue.
Added a new _BreakLoopException that is thrown by the ``$ref``` rule, and removed the ad-hoc logic at the beggining of the error-loop.

* Not only faster, but better standard-deviation --> probably more memory-efficient.
* Expecting the usual TravisCI problem with python-2.7.6 (pypy)


Julian:
Think the parts of this that were mergeable were covered already -- feel free to reopen if there's more that needs looking at and thanks again, much appreciated.

PR #192, opened 2015-01-07
Fix Draft4Validator example code in validate.rst

bennoleslie:


Julian:
Ugh, it's really terrible how hard it is to get basic automated validation on this stuff. Thanks for catching, looks great.

PR #202, opened 2015-02-28
Performance - reduce urljoin/urldefrag overhead

dnephin:
Based on the work in #182 (I cherry-picked a few commits) for issue #158. Also happens to fix the problem described in #201 at the same time.

Rebased on master, fixed tests and style checks.

Removed base_uri since it was not really being used now that there is a stack.

I hope this is a smaller chunk to review.

### DIFF
--- a/bench.py
+++ b/bench.py
@@ -1,0 +1,74 @@
+#!/usr/env/bin python
+"""
+Benchmark the performance of jsonschema.
+
+Example benchmark:
+
+    wget http://swagger.io/v2/schema.json
+    wget http://petstore.swagger.io/v2/swagger.json
+    python bench.py -r 5 schema.json swagger.json
+
+"""
+from __future__ import print_function
+import argparse
+import cProfile
+import json
+import time
+
+import jsonschema
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('schema', help="path to a schema used to benchmark")
+    parser.add_argument('document', help="document to validate with schema")
+    parser.add_argument('-r', '--repeat', type=int, help="number of iterations")
+    parser.add_argument('--profile',
+                        help="Enable profiling, write profile to this filepath")
+    return parser.parse_args()
+
+
+def run(filename, schema, document):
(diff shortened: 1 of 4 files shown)


Julian:
Awesome, definitely helps, I'll give a shot at looking this over.

I mentioned in the other ticket I would want to maintain backwards compatibility for APIs even if they're not used internally anymore, I haven't checked carefully but I expect that to be fairly easy to add back for this subset?

dnephin:
> I mentioned in the other ticket I would want to maintain backwards compatibility for APIs even if they're not used internally anymore

Cool, I believe the only thing that was removed was `RefResolver.in_scope()`. It is easy enough to restore that with the stack implementation.

> A benchmark, even a simplistic one, would also be lovely for anything performance related 

I was using a very simple one myself to profile the changes, so I it should be no problem to include it.


I actually have another branch which uses a cache around some of the slower operations and is showing another significant performance increase ontop of these changes.  I'll see about getting all of these changes together into a single PR.

dnephin:
I've restored `in_scope()` which I believe is the only breaking API change, and added `bench.py`.

Using the example schemas in the bench.py docstring,

master: ~56ms
this branch: ~44ms

I'm going to experiment more with the caching branch.

dnephin:
With caching this is down to ~21ms. I'm going to get that branch cleaned up and replace this review I think.

ankostis:
A great thank you @dnephin  for getting up to this!

Julian:
Definitely, appreciated.

Running your benchmark on a warmed up PyPy basically yields equivalent times on my machine, which means this doesn't really speed up anything there (probably expectedly). I'm generally OK with CPython-only speed improvements (which I assume is where you're seeing improvement) as long as it doesn't *slow down* things on PyPy or introduce other complexity, so seems like we're on the right track here. I'll play a bit more with this until you've got your caching branch as well, this is small enough for me to have a shot at reviewing right now :)

dnephin:
Fair point, I wouldn't expect to see much improvement in pypy.  I've opened #203 since I rebased over a bunch of these smaller commits.

PR #203, opened 2015-03-02
Perfornance - Cache expensive url operations

dnephin:
Based on the work in #182 (I cherry-picked a few commits) for issue #158
Replaces #202
Also fixes the problem described in #201

I rebased over many of my changes in #202. Many of them were unnecessary after adding caching.

On cpython 2.7 I'm seeing the benchmark runtime go from ~56ms to ~22ms.  I don't expect any significant improvement in pypy, since it's already going to have this cached.

The primary change here is to add two caches to `RefResolver`, one for `urljoin`, and another for resolving a reference. 

Removing the context manger from `ref()` validation (the latest commit), gets this down to ~19ms


jsonschema/tests/test_validators.py
```diff
@@ -775,11 +775,11 @@ def test_it_resolves_local_refs(self):
             self.assertEqual(resolved, self.referrer["properties"]["foo"])
 
     def test_it_resolves_local_refs_with_id(self):
-        schema = {"id": "foo://bar/schema#", "a": {"foo": "bar"}}
+        schema = {"id": "http://bar/schema#", "a": {"foo": "bar"}}
         resolver = RefResolver.from_schema(schema)
         with resolver.resolving("#/a") as resolved:
             self.assertEqual(resolved, schema["a"])
-        with resolver.resolving("foo://bar/schema#/a") as resolved:
+        with resolver.resolving("http://bar/schema#/a") as resolved:
```
dnephin: I had to change these to `http` because (at least on py3) the behaviour of `urljoin` changes based on the scheme.  With an unknown scheme it was not replacing the fragment at all.

jsonschema/tests/test_validators.py
```diff
@@ -863,9 +863,7 @@ def test_cache_remote_off(self):
         )
         with resolver.resolving(ref):
             pass
-        with resolver.resolving(ref):
-            pass
-        self.assertEqual(foo_handler.call_count, 2)
+        self.assertEqual(foo_handler.call_count, 1)
```
dnephin: With caching this is no longer called twice, so I had to set the count to 1.

jsonschema/_utils.py
```diff
@@ -38,6 +38,22 @@ def __repr__(self):
         return repr(self.store)
 
 
+class Cache(object):
```
Julian: I'd prefer to use something pre-baked for caching. `functools.lrucache` with the backport module, say?

jsonschema/compat.py
```diff
@@ -40,6 +43,9 @@ def urlsplit(url):
     return SplitResult(scheme, netloc, path, query, fragment)
 
 
+DefragResult = namedtuple('DefragResult', 'url fragment')
```
Julian: Is this to just gain attribute access? If so I'd prefer an actual class, namedtuple is a plague.

jsonschema/tests/test_validators.py
Julian: Yeah there's some global that defines weirdly which schemes it's fine with. I don't recall why this worked before then?

jsonschema/_utils.py
dnephin: That should work. This doesn't really need the "lru" portion, but with a sufficiently large value for maxsize, that should be fine.

Apparently both `functools32` and `repoze.lru` have this. I'll look into which ones work on py2.6 

On a related note, do you think I should add a new kwarg to `RefResolver` to disable caching entirely (since there is already an option to disable caching of remote refs) ?

jsonschema/tests/test_validators.py
dnephin: I believe it worked before because it wasn't joining the ref to the rest of the url. The url was being fetched first, then the ref was being used to locate the section in the document. Now the full url is being stored as the scope instead.

jsonschema/compat.py
dnephin: Previously `urldefrag` was being used in a few places, I believe it's only being used in one place now, so this is less necessary (and I believe it is only for attribute access). I can remove it (it was actually pulled in from the cherry-pick of the previous changes from ankostis).

I must say I'm surprised by this response. I consider namedtuple one of the nice parts of python, and one of the most underused.  Replacing this with a class would be pretty verbose and lots of boilerplate (slots=(...), assignment in the construction, equality and comparison functions). In my experience keeping data objects separate from functionality that operations on them also makes code a lot more reasonable (which is why I find `namedtuple` to be so valuable, it lets you create data objects very easily).

What's wrong with namedtuple?

jsonschema/_utils.py
Julian: Maybe a similar solution to what we do for remote cache would work, with the `cache` just being an optional argument which constructs an appropriately configured `lrucache`, but disabling caching would be providing a `functools.lrucache` with maxsize=0 or whatever.

But yeah, sounds like a decent idea. Nice catch.

jsonschema/compat.py
Julian: Ah OK, understood, up to you on whether to remove or not.

Basically I agree with all of the nice things you mentioned about why data objects are nice, but namedtuple's biggest issue is that it's... a `tuple`. That introduces all sorts of unintended consequences that people don't realize when they use it. Specifically to give just one example, it means that two namedtuples of *entirely different types* are comparable, and will compare equal if they contain the same positional elements. Really really really bad. I think you're right about namedtuple being convenient though, FWIW that's why something like `characteristic` (http://pypi.python.org/pypi/characteristic) is an attempt bridge the gap there.

jsonschema/tests/test_validators.py
Julian: Ah, OK, makes sense.

jsonschema/validators.py
```diff
@@ -79,7 +79,10 @@ def iter_errors(self, instance, _schema=None):
             if _schema is None:
                 _schema = self.schema
 
-            with self.resolver.in_scope(_schema.get(u"id", u"")):
+            scope = _schema.get(u"id")
+            if scope:
```
Julian: Probably want `is not None`? Which might mean a missing test case? (an `id` of "" is probably meaningless, but I guess it means something reasonable relative to some other base URI.)

jsonschema/validators.py
```diff
@@ -233,13 +239,12 @@ class RefResolver(object):
...
         self, base_uri, referrer, store=(), cache_remote=True, handlers=(),
     ):
-        self.base_uri = base_uri
-        self.resolution_scope = base_uri
         # This attribute is not used, it is for backwards compatibility
         self.referrer = referrer
         self.cache_remote = cache_remote
         self.handlers = dict(handlers)
 
+        self.scopes_stack = [base_uri]
```
Julian: Probably makes sense to make this private (same for all of the below).

jsonschema/_validators.py
```diff
@@ -190,9 +190,14 @@ def enum(validator, enums, instance, schema):
 
 
 def ref(validator, ref, instance, schema):
-    with validator.resolver.resolving(ref) as resolved:
+    scope, resolved = validator.resolver.resolve(ref)
+    validator.resolver.push_scope(scope)
```
Julian: There's another backwards incompatibility here, which is that these are newly introduced methods on the `RefResolver` interface that people who might have implemented them did not need until now.

To resolve that, I think we'd either need to fall back to `resolving` if there was no `push_scope` attribute, or perhaps you might consider an interface that composes `RefResolver` rather than modifying `RefResolver` itself. Lemme know if this makes sense.

jsonschema/validators.py
```diff
@@ -259,17 +267,40 @@ def from_schema(cls, schema, *args, **kwargs):
 
         return cls(schema.get(u"id", u""), schema, *args, **kwargs)
 
+    def push_scope(self, scope):
+        self.scopes_stack.append(
+            self.urljoin_cache(self.resolution_scope, scope))
+
+    def pop_scope(self):
```
Julian: Same comment about making these methods private, unless there's a reason you can think of to not (although I thikn this question is moot if we consider the composition route I mentioned above) -- certainly if it's not private we probably should handle errors a bit nicer, which might make sense regardless.

jsonschema/validators.py
```diff
@@ -277,26 +308,20 @@ def resolving(self, ref):
...
-        if not uri:
-            uri = self.base_uri
-
-        if uri in self.store:
-            document = self.store[uri]
-        else:
+    def resolve_from_url(self, url):
+        ref = urldefrag(url)
+        try:
+            document = self.store[ref.url]
```
Julian: Is the `KeyError` vs using `.get` intentional here?

Julian:
Left some more comments inline as I'm trying to review this in the quick couple of minutes I can grab :D

Appreciated again, thanks a lot for pushing this forward.

jsonschema/_utils.py
dnephin: So it looks like `functools32` is py2.7 only (I actually need 2.6 still).

`repos.lru` works in 2.6, but it's not small (https://github.com/repoze/repoze.lru/blob/master/repoze/lru/__init__.py ~373 lines). I notice that currently there are no `install_requires` dependencies.

I could add an install_requires for `version < 3.2` or I could vendor it in `jsonschema/vendor/repos/lru.py`. I usually prefer to add the dependency instead of vendoring it.  Do you have a preference here?

jsonschema/_validators.py
dnephin: The performance improvement here was to remove the contextmanager.  I'm not sure how to compose over that change. The interfaces seem incompatible to me (but I could be missing something).

Falling back to the old interface would be easy enough:
```python

if hasattr(validator.resolver, 'resolve'):
    ... # do what this does now
    return

... # Use the old contextmanager way

```

jsonschema/compat.py
dnephin: That is a good point about tuple comparison, I hadn't considered that.

I think I will remove this. It's only used locally in a single function now, so the attributes are a lot less necessary.

jsonschema/validators.py
dnephin: I don't think "" is relevant, here's my simple test cases:

```python

In [7]: urljoin("", "#/path")
Out[7]: '#/path'

In [8]: urljoin("", "http://example.com/#/path")
Out[8]: 'http://example.com/#/path'

In [9]: urljoin("#/path", "")
Out[9]: '#/path'

In [10]: urljoin("http://example.com/#/path", "")
Out[10]: 'http://example.com/#/path'

In [11]: urljoin("http://example.com/", "")
Out[11]: 'http://example.com/'

```

jsonschema/validators.py
dnephin: Since these (pop/push) are being called from `ref()` I think they need to remain public.  I'm still not sure about the composition route.

The errors here would be for `pop()` on an empty stack?  Theoretically that should never happen because we always match push/pop, but a nicer error won't hurt either.

jsonschema/validators.py
dnephin: Yup, it is intentional. Credit goes to ankostis for this one (this was from the cherry-pick).

While raised exceptions are slower, we actually expect the common case to be retrieving the document succesfully.  I believe in the success case it's actually faster to use `__getitem__` directly and try/except.

dnephin:
Thanks for the review and feedback!  I've got a couple open questions (how to include the lrucache dep, and how composition would work), then I'll make the changes.

jsonschema/_utils.py
Julian: Ech, annoying. OK, yeah adding the dependency is fine with me.

jsonschema/_validators.py
Julian: Just curious -- did we benchmark whether the reason it's slow is a context manager vs specifically `functools.contextmanager`?

But yeah, sorry, you're right -- although you don't need to do that check here per se every time you hit this function, you can check to see if you have a new or old `RefResolver` in `__init__` for the validator and bind accordingly?

jsonschema/validators.py
Julian: Ech yep, right again. OK.

And yes to the error -- it's true it's unlikely, but if it's public API, someone else can use these, and do so incorrectly. Not a huge deal though.

jsonschema/validators.py
Julian: K, makes sense.

jsonschema/_validators.py
dnephin: I did experiment with using a class with `__enter__` and `__exit__` and it was about the same as `contextlib.contextmanager` (possibly even a little slower).  It's always possible I was doing something non-optimal.

Julian: Got it. OK, I'd assume there's some cost in generator-nonsense with `contextmanager`, but I'm not sure I'd have guessed it'd be insignificant to switch to the regular protocol. Regardless, this way is fine with me, but if I have an overabundance of time (ha) I'll double check.

jsonschema/_utils.py
dnephin: Cool, the interface is identical, I'll only install `repoze.lru` for python2.

Adding a dependency required a small change to where `__version__` is defined.  

When trying to run setup, `setup.py` imports `__init__.py` imports `validators.py`  imports `comat.py` tries to import `repoze.lru` which will fail before the virtualenv doesn't exist yet.  So I created a `version.py`, moved `__version__` there, imported it from both `setup.py` and `__init__.py` and use `exec()` to read it without importing it.  

I believe this is a pretty common way of handling this problem.

dnephin: Sadly with `lru_cache` the benchmark is back to ~21ms (it's ~19.5ms with my custom Cache object).  I guess that is good enough though. I understand wanting to use something that already exists.

Julian: I've been meaning to switch to http://vcversioner.readthedocs.org/en/latest/ which handles that process for us and isn't overcomplex like some other tools that do the same sort of thing.

Good to hear that we can swap those.

Julian: We're making the cache a parameter though right? So you should be able to drop in something faster if the 2ms are relevant for your app?

dnephin: Ah yes, good point. I was considering just making the maxsize a parameter, but making the decorator function itself a parameter makes sense.

dnephin:
Ok, I think I have all the changes in.

Julian:
Awesome. I'm out on business, but I'll have another look and try to merge on Sunday, thanks again!

Julian:
OK! Think this roughly lgtm -- some of the assertions that were removed I'd like to add back since they test backwards compatible things we're forced to leave, but I can add them back in after merging, which now I hopefully will have time for tomorrow.

Thakns again!

dnephin:
Awesome, looking forward to it.  Thank you!

Julian:
Started merging (in https://github.com/Julian/jsonschema/tree/perf_cache_resolving) but adding back in the assertions I was referring to actually fails, so there's some more backwards incompat. I'll try to give it a bit more time early this week, unless you see immediately what the correct way to add `base_uri` back in is.

dnephin:
Ah, I see what you were saying. My latest commit 7241db0 fixes it.

Julian:
OK managed to find slightly more time, and made a few small changes -- needed to add back backwards compatibility for RefResolvers without the new methods, and I changed the parameters to take caches, rather than parameters they can use to make caches.

I think I'd like to move `RefResolver.resolve` into a function that handles the backwards compat (and leave it off the interface), and possibly combine it with `push_scope` since that's how you're using it (so, say `resolve_into` -> `resolve` + `push_scope`, similar to the context manager, which was sort of what I was going with before with suggesting composition but I think it can work as a function, and then after all that we'll probably have to run your benchmark again to make sure we've still gotten whatever performance gains we're looking for.

So we're close :), sorry it's taking so long.

dnephin:
What would call `resolve_into()`? `_validators.ref()` ? Would it be added to the list of validator directly instead of `ref()`?

Julian:
The same place (no not added to validator) -- I'm imagining just putting the caches on Validator (in `create`) -- it's probably just easier to show you the change and see if it makes sense, sorry this is taking so long, just trying to avoid digging ourselves into a hole by changing interfaces we have little control over (RefResolver) when we possibly can introduce a new one (or at worst slightly modify one we do control easiliy, the Validator interface) -- going to do that I hope by the end of the week and ping you.

Julian:
OK never mind, I tried it and like this way better :)

Merged. Sorry again that this took so long, much appreciated!

dnephin:
That's for all the work getting this merged!

Since there is a bug fix included in this change (for python3), would it be possible to do a bug fix release (v2.4.1) ?

Julian:
I'd like to cut 2.5, but Travis is using a build of ubuntu that's failing the test suite on PyPy, which I have to figure out.

I can certainly try to cut a bugfix release, but which bug fix is in here again?

dnephin:
I guess I never opened a ticket for it. The issue is described in #201 (which I discarded because this fixed it in a better way).

Julian:
Ah the test flakiness... OK lemme see if I can make a branch with just the fix for that on it, hopefully tomorrow, if not sooner if I can just get 2.5 out as well.

bpicolo:
@Julian Think you'll have a chance to cut the bug fix release? (Or is 2.5 close)?

Julian:
I hope so, sorry :( I'm back from my trip, so hoping to have some spare
bandwidth over the weekend.

prat0318:
@Julian wanted to follow up if the patch could go in any immediate releases?

Julian:
OK, release is out. Sorry for the delay here all. Feel free to follow up with any issues.

PR #211, opened 2015-03-24
CLI: Read JSON instances one to a line from stdin.

jkleint:
This lets you mass-validate JSON by piping instances into stdin, instead of having to put them in separate files and say `-i file.json`... for every one.  Each instance has to be on one line.  You use it like this:

```
python -m jsonschema schema.json < instances.json
```

You may want to do things differently, but this shows how it could be done.

### DIFF
--- a/jsonschema/cli.py
+++ b/jsonschema/cli.py
@@ -19,15 +19,7 @@ def _json_file(path):
 
 
 parser = argparse.ArgumentParser(
-    description="JSON Schema Validation CLI",
-)
-parser.add_argument(
-    "-i", "--instance",
-    action="append",
-    dest="instances",
-    type=_json_file,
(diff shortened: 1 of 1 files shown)


Julian:
Hi!

Thanks so much for the PR.

This certainly seems useful -- as implemented though it breaks backwards compatibility -- maybe it'd be useful to have as a new command line flag (or alternatively, read from stdin if no instances were provided).

It'd also be nice to have a test for the new functionality.

Does that sound reasonable?

jkleint:
You're welcome.  I was thinking a `--filter` option that passed through valid instances and logged invalid ones to stderr would be useful.  May be a while before I can get to it.

jkleint:
Implemented this in #213, let me know what you think.

PR #212, opened 2015-04-01
Validate function now supports filepaths in addition to dictionaries.

SamuelMarks:


Digenis:
I think 2 calls `json.load(open())` don't take up much space to worth a shortcut.

Julian:
This change isn't particularly safe, it leaks file objects, but also what's the advantage of doing this? Why can't the caller pass in what they want? Functions shouldn't take inputs of multiple types generally.

Julian:
Going to pass on this but feel free to follow up if you'd like. Thanks!

PR #213, opened 2015-04-02
CLI option --filter: validate multiple instances from stdin.

jkleint:
Reads JSON instances one to a line from stdin, outputs those
that validate in compact form (one to a line).  Print errors
to stderr for those that do not validate.

### DIFF
--- a/jsonschema/cli.py
+++ b/jsonschema/cli.py
@@ -21,13 +21,20 @@ def _json_file(path):
 parser = argparse.ArgumentParser(
     description="JSON Schema Validation CLI",
 )
-parser.add_argument(
+input_group = parser.add_mutually_exclusive_group()
+input_group.add_argument(
     "-i", "--instance",
     action="append",
     dest="instances",
     type=_json_file,
     help="a path to a JSON instance to validate "
          "(may be specified multiple times)",
+)
+input_group.add_argument(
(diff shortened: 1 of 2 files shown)


Julian:
Cool! Thanks so much.

So this seems useful too, but isn't there a way to make it non-exclusive with the old way?

E.g., we could add a `-f` / `--instance-file` which takes line-delimited JSON instances, and then use `-` for stdin to handle your case, and then a `--filter=valid` that shows valid instances on stdout in the collected instances passed either via `-i` or `-f`?

jkleint:
Or you can use `cat` to handle all of those cases, and [keep it simple](http://www.faqs.org/docs/artu/ch01s06.html).

Julian:
Well yes or you can use a for loop with -i and keep it even simpler :).

KISS is always relative: to existing functionality, and to how flexible you want to be, and to forward compatibility.

I'm not saying I've gotten those three exactly correct by throwing that out there, but it'd seem slightly more so than what's here, unless you feel differently.

jkleint:
I like it as is, but of course you're free to embellish.

Julian:
I doubt I'll personally ever hit this, so I'm going to close for now, but if anyone else ever feels like discussing further feel free.

Thanks certainly for the submission.

Issue #228, opened 2015-06-05
jsonschema 2.5.0 not installable on Python 3


wichert:
You use a universal wheel for jsonschema, which means the dependencies are forced to be identical on Python 2 and 3. This breaks the handling of functools32, which is only required (and installable) on Python 2. That results in this error:

```
$ bin/pip install --upgrade jsonschema
Collecting jsonschema
  Downloading jsonschema-2.5.0-py2.py3-none-any.whl
Collecting functools32 (from jsonschema)
  Downloading functools32-3.2.3-1.tar.gz
    Complete output from command python setup.py egg_info:
    This backport is for Python 2.7 only.
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/p_/p3cbpfr96sb7h5gpq6vfvwcc0000gn/T/pip-build-d1dmcog1/functools32
```

SylvainCorlay:
Yeah, apparently the pypi release is broken. The egg.info includes a requires.txt that lists `functools32` which is Python 2.7 only.

wichert:
It's specifically only the wheel which is broken. The source releases are fine.

SylvainCorlay:
I think that it depends on how you install it.

asottile:
This is also broken in py26 (I imagine broken in any version that is not 2.7).  The wheel is not universal and probably shouldn't be marked as such.

Julian:
Thanks for the report, sorry that things are broken -- I'm trying to get some tests around this so that I can fix it and push a bugfix. (https://github.com/Julian/jsonschema/tree/install-checks has the idea, I'm just gonna use travis to make sure that what a release generates is installable).

I'll update as soon as I figure things out.

asottile:
@Julian I believe simply deleting the .whl from pypi will solve this in the short term (forcing installation from sdist)

SylvainCorlay:
And requires.txt in egg.info?

mriedem:
Confirmed in openstack py34 jobs today:

https://jenkins03.openstack.org/job/gate-nova-python34/82/console

dims:
And some of openstack py26 jobs as well:
[...]

PR #230, opened 2015-06-06
Use PEP426 environment markers for extras_require.

mdomke:
This pull request tries to address the issue #228 and introduces PEP426 compliant environment markers for the `extras_require` specification as proposed by @takluyver.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,14 +22,18 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: PyPy",
 ]
 
+extras_require = {
+    'format': ["rfc3987", "strict-rfc3339", "webcolors"],
+    ':python_version == "2.6"': ["argparse", "repoze.lru"],
+    ':python_version == "2.7"': ["functools32"]
+}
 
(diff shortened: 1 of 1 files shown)


takluyver:
This looks correct to me.

Julian:
Thanks so much, this looks something like what I expected too, although I really would like to get some automation before merging to make sure that this doesn't happen again. I can give that another shot hopefully tomorrow, unless you've got some ideas.

Also CC @dstufft just in case I can bother you into a consult for this, context is in #228.

rbtcollins:
I don't think the checking for bdist_ in args is needed: setuptools should handle the markers fine for egg_info, sdist and install too. I put up #231 before I knew this was here, it doesn't have the bdist_ checks.

The basic approach is here is fine, its exactly what the wheel package does.

Julian:
So thanks again! I'm going to merge #231 only because the way that that looping over argv looks makes me uneasy. But I really appreciate everyone pitching in especially with the amount of time I have to think about this specifically :/

PR #231, opened 2015-06-08
Fix universal wheels

rbtcollins:
The current universal wheels have static install_requires matching the python version they were built on: so a wheel built on 2.6 claims it needs argparse and repoze.lru on 2.7 and 3.4, etc. This clearly doesn't work :)

environment markers are however honoured by wheels and pip, using this somewhat ugly syntax in extras (I'm looking to improve that, but this works - see the distutils-sig list for some discussion on it recently).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,20 +23,15 @@ classifiers = [
 ]
 
 
-extras_require = {"format" : ["rfc3987", "strict-rfc3339", "webcolors"]}
-if sys.version_info[:2] == (2, 6):
-    install_requires = ["argparse", "repoze.lru"]
-elif sys.version_info[:2] == (2, 7):
-    install_requires = ["functools32"]
-else:
-    install_requires = []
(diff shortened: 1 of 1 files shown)


FlaPer87:
This should also fix #228

mdomke:
This is basically the same as #230

rbtcollins:
Yeah it is; it doesn't have the cruft for old setuptools; I'll leave it up to the jsonschema maintainers to decide which they want.

Julian:
Hey! Thanks. I'm going to merge this even though there isn't anything testing this and that makes me pretty unhappy, but I don't have the bandwidth to get that in place at this very second, and I know people are waiting, so it'll have to be circled back for.

Very much appreciated, I'll try to get another release out with this change.

Carreau:
+1 for when possible, it's Monday morning and the CI test are starting to fail in a couple of repos.

PR #232, opened 2015-06-08
Speedup Validator.is_type

dtheodor:
Hi,

I made a few changes in the `Validator.is_type`, this results in a total speedup of 5% as reported by running `python bench.py -r 10000 schema.json swagger.json`, which gives a `Mean: 53.578ms` compared to the `Mean: 56.780ms` of the current master branch. `bench.py` was run with the swagger files downloaded as specified in said file. I ran the bench in an ubuntu vm with python2.7, feel free to run it in your own setup.

### DIFF
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -65,7 +65,7 @@ def create(meta_schema, validators=(), version=None, default_types=None):  # noq
 
             # prepare types
             self._types = {}
-            for type_key, pytypes in _types.iteritems():
+            for type_key, pytypes in _types.items():
                 pytypes = _utils.flatten(pytypes)
                 self._types[type_key] = {
                     "pytypes": pytypes,
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -60,8 +60,19 @@ def create(meta_schema, validators=(), version=None, default_types=None):  # noq
         def __init__(
             self, schema, types=(), resolver=None, format_checker=None,
         ):
-            self._types = dict(self.DEFAULT_TYPES)
-            self._types.update(types)
+            _types = dict(self.DEFAULT_TYPES)
+            _types.update(types)
+
+            # prepare types
+            self._types = {}
+            for type_key, types in _types.iteritems():
+                types = _utils.flatten(types)
+                self._types[type_key] = {
+                    "types": types,
+                    "is_non_bool_number": any(
+                        issubclass(type_, numbers.Number) for type_ in types
+                    ) and bool not in types
+                }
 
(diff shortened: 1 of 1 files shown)


Julian:
Hey, awesome, thanks so much -- sorry I missed this during the release, I'll give it some time to review this week.

Julian:
(Just to be clear for anyone finding this, I'm really sorry this fell for so long, but what it's waiting for is the addition of a programmatic benchmark that can be added, so that we're sure this change actually makes positive impact but so that it also continues to exist as more changes are made. I think looking at https://github.com/haypo/perf is probably a reasonable place to start, even though jsonschema unfortunately already has a bunch of ad-hoc benchmarks.)

pgjones:
This definitely makes a positive impact as you can see from [before](http://vmprof.com/#/ffe52dbe-22a3-4b09-8de2-9fe40deb0894) and [after](http://vmprof.com/#/e0845617-908d-4460-8c54-51f55d3f36e8) vmprof profiles of a simple setup. 
How do you expect the programatic benchmark to work? 
Could you merge this on the data I've provided above?

Julian:
@pgjones sorry for the delay here, will elaborate in a second I hope, but what I'd expect would basically be an automatable thing that would answer "after someone makes a change, has validator.is_type gotten slower?" and in particular for both CPython and PyPy.

I'll have a look at the vmprof output you've got, but I suspect that while they might confirm that stuff is slow here, we still would want to be able to automate a benchmark that gives an absolute answer to "how much faster or slower is the codebase on some benchmark that uses .is_type, or .validate" and then have that benchmark run on each commit (in CI)

I also vaguely recall that this patchset breaks backwards compatibility -- I might be misremembering, but if it does we'd need to address that as well.

Sorry again for the delay here, please do nag me (or even better help out with the above :) if you think that this really is in a close state to being worthwhile.

Julian:
It seems like https://github.com/ebdrup/json-schema-benchmark is a thing in the node.js JSON Schema community -- seems like it'd be a decent idea to have a look at integrating with that / those benchmarks as well.

pgjones:
Using the code in #355 I get the following locally,
```
+-----------------------------------+---------+-----------------------------+
| Benchmark                         | master  | speedup_is_type             |
+===================================+=========+=============================+
| standard Petstore Example petsore | 29.7 ms | 27.1 ms: 1.09x faster (-9%) |
+-----------------------------------+---------+-----------------------------+
```

Julian:
Thanks (and apologies again) to all here, I merged the one from the other PR, we're all good here on benchmarks! Hooray!

PR #242, opened 2015-07-24
Set cause to the correct value during FormatError.

arianvp:
When reading the documentation and reading the source code of jsonschema
I found an inconsitency in the code.

Basically, when a FormatError occurs, it's re-raised into a
ValidationError with its cause set to the original FormatError according
to the documentation. Which makes a lot of sense.

But in the code, the cause of the ValidationError is set to the cause of
the FormatError. Which is not the documented behaviour nor very useful
(in my opinion); So I consider this as a bug.

See the documentation here:
https://python-jsonschema.readthedocs.org/en/latest/errors/#jsonschema.exceptions.ValidationError.cause

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -149,7 +149,7 @@ def format(validator, format, instance, schema):
         try:
             validator.format_checker.check(instance, format)
         except FormatError as error:
-            yield ValidationError(error.message, cause=error.cause)
+            yield ValidationError(error.message, cause=error)
 
 
 def minLength(validator, mL, instance, schema):


Julian:
Hi! Thanks.

Can you clarify why you think this isn't useful? What is it you want off of a `FormatError` object?

arianvp:
Well i want to be able to tell if an error happened due to the validator failing or due to a format checker failing.  The docs say I can do this with e.cause, but because we create a new ValidationError and set its cause to format_error.cause, ValidationError's cause will always be None because usually FormatErrors don't have cause set

arianvp:
So I think the ValidationError should have its cause set to the FormatError

Julian:
Which validator?

Do you just mean you want `if error.validator != "format"`? `cause` is supposed to mean whatever non-validation error caused the failure, e.g. if you used a hostname format then it will be some `ValueError` that was raised with information about why the hostname is invalid. `FormatError`s should generally have cause set, see the tests that fail after this change is made.

If you have a case where one isn't being set I'd love to help, could you provide some details on your instance and schema?

arianvp:
So a `FormatError` is not considered a non-validation error then?  I thought the docs meant: "Because  `FormatErrors` that can occur in formatters are not `ValidationError`s, we return a `ValidationError` in the `format` validator with its `cause` set to `FormatError`.

The use-case we had was: We wanted to print the reason of a failed schema check only if a format error occured. Namely, we used a formatter to check if a password string was valid,  and instead of leaking the password back to the user or into the logs, we wanted to do a custom error message instead.

How would I do that in this case? check `if error.validator != 'format'`  ?

Julian:
FormatError is just a part of the format API that rolls up other Exceptions
for Validators so that they can not know what specifically they need to
catch. It'd be private except that you can implement custom formats.

But otherwise there's nothing useful for library users in them. The docs
mean "with the cause being whatever the actual exception was, which was
something before the format error".

Presumably yes, you'd just want

message = error.message if error.validator != " format" or
error.schema_path[-1] != "password" else "Invalid Password"

arianvp:
Alright thanks. That makes sense. I guess we can close the pull request then

Julian:
Cool, lemme know if you have any other questions.

Issue #119, opened 2013-08-16
required, additionalProperties, propertyNames, dependentRequired and any other validators that reference property *names* need a way to indicate the erroring names in ValidationErrors


martinpengellyphillips:
Trying to figure out if there is a way to retrieve the name of a missing required property from an error (without hacking it out of error.message).

For example:

``` python
from jsonschema import Draft4Validator as Validator

schema = {
    'type': 'object',
    'properties': {
        'a': {
            'type': 'string'
        },
        'b': {
            'type': 'string',
            'pattern': '^hello$'
        }
    },
    'required': ['a', 'b']
}

errors = Validator(schema).iter_errors({'b': 'world'})
for error in errors:
    print error.path, error.schema_path, error.message

# deque([]) deque(['required']) 'a' is a required property
# deque(['b']) deque(['properties', 'b', 'pattern']) 'world' does not match '^hello$'
```

In error.path and error.schema_path there is no mention of 'a'. Is there a way to determine that the property missing was 'a' without using error.message?

leplatrem:
We also bumped in a similar limitation, accessing the list of fields for errors like `Additional properties are not allowed ('field1', 'field2' were unexpected)`.

As said earlier, with a minimum of indication on the expected strategy we can submit a PR :)

Thanks !

GMLudo:
+1, for now we show directly the error message via our API,  but we'll need soon to able to translate and to give more human compliant error messages ;-)

Julian:
@leplatrem the expected strategy I think is still to just introduce ``extra_info``, which is the most reasonable thing I think unfortunately of the current options.

fre-sch:
As I've commented [on 13 Mar 2014](https://github.com/Julian/jsonschema/issues/119#issuecomment-37510798) a simple change in `jsonschema/_validators.py` would make the name of the required property available in the ValidationError instance as `path` as well as `relative_path`:

[jsonschema/_validators.py](https://github.com/Julian/jsonschema/blob/19a37febe8a3eb36d3fdd5cd5a46d47cfa8251ae/jsonschema/_validators.py#L296)
```diff
 def required_draft4(validator, required, instance, schema):
     if not validator.is_type(instance, "object"):
         return
     for property in required:
         if property not in instance:
-            yield ValidationError("%r is a required property" % property)
+            yield ValidationError(
+                "%r is a required property" % property,
+                path=[property],
+            )
```

[...]

PR #254, opened 2015-10-25
Add path into extras

gdassori:
Invalidating extras doesn't provide the involved property name into ValidationError.path, as other validators seems do.


Julian:
Hey! Thanks so much for the PR.

The `path` attribute is supposed to be the path into the instance where the error occurred. That isn't the case here, you're passing a set of property names to it (not a tuple, by the way, in case that was what you meant).

There's an issue that's been open to add support for an `extra_information` field on `ValidationErrors` in #119  which you might be interested in, that would cover this case. I'm going to close this as a duplicate but feel free to follow up over there if you're interested. Thanks again!

PR #255, opened 2015-10-27
docs: faq: Updates to extend_with_default() example

stuarteberg:
The FAQ explains that one could (in theory) write a schema in which default values aren't valid under their own schema, and then goes on to present some example code that would allow such a use-case.

Fair enough, but I think most users would rather have an example that actually *does* validate the defaults.  Plus, if the default values are added to the instance *before* validation, then entire objects can be provided with default contents, even if the object was missing from the instance entirely.  (The updated FAQ text in this PR provides an example.)

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -19,7 +19,9 @@ would pass validation the first time, but fail the second!
 Still, filling in defaults is a thing that is useful. :mod:`jsonschema`
 allows you to :doc:`define your own validator classes and callables
 <creating>`, so you can easily create a :class:`IValidator` that does do
-default setting. Here's some code to get you started:
+default setting. Here's some code to get you started. (In this code, we add
+the default properties to each object *before* the properties are validated,
+so the default values themselves will need to be valid under the schema.)
 
     .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


stuarteberg:
In case it's helpful, you can view a pretty version of the new doc text here:
http://stuarteberg.github.io/jsonschema/faq.html#why-doesn-t-my-schema-that-has-a-default-property-actually-set-the-default-on-my-instance

Julian:
Cool, thanks! Doc improvements are very much appreciated. I haven't read through this yet, but FWIW, I really don't recommend doing `default` setting during validation at all -- I forget whether that's mentioned in that part of the FAQ.

But I'll have a read through, thanks again!

stuarteberg:
>FWIW, I really don't recommend doing default setting during validation at all

Indeed, your text mentions that, and I didn't change it (I only added text).  I know default-setting isn't great for all use-cases, but I still *really* appreciate that you gave an example in the FAQ.  My only problem with the example code was that it didn't work with nested objects out-of-the-box.  By swapping a few lines, now it does.

Julian:
Cool, looks good, appreciated.

PR #261, opened 2015-11-20
Replace rfc3987 (GPL) with rfc3986 (Apache)

shakeelmohamed:
We'd like to consider using the jsonschema package which is licensed under an MIT-like license, but cannot take a dependency on rfc3987 because it is distributed under GPL.

This was the minimum amount of changes required to make the tests pass, we're open to any feedback on these changes!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ _static
 _templates
 
 TODO
+*.pyc
+.eggs/
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ TODO
 *.pyc
 .eggs/
 .idea/
+.tox/
+jsonschema.egg-info/
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -158,7 +158,7 @@ def is_host_name(instance):
 
 
 try:
-    import rfc3987
+    import rfc3986
 except ImportError:
     pass
 else:
@@ -166,8 +166,10 @@ else:
     def is_uri(instance):
         if not isinstance(instance, str_types):
             return True
-        return rfc3987.parse(instance, rule="URI")
-
+        elif not rfc3986.is_valid_uri(instance, require_scheme=True):
+            return False
+        else:
+            return rfc3986.normalize_uri(instance)
 
 try:
     import strict_rfc3339
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ classifiers = [
 ]
 
 extras_require = {
-    "format" : ["rfc3987", "strict-rfc3339", "webcolors"],
+    "format" : ["rfc3986", "strict-rfc3339", "webcolors"],
     ":python_version=='2.6'": ["argparse", "repoze.lru"],
     ":python_version=='2.7'": ["functools32"],
 }


shakeelmohamed:
Just realized the docs will need an update, I will make that change if you're open to merging this https://github.com/Julian/jsonschema/blob/master/docs/validate.rst

Julian:
Hi! Thanks for the patch.

I have no issue with this in theory, supporting non-GPL is certainly something I would love to do. And yeah would need a doc update. Leaving some comments in-line on the changeset.

Thanks again!

.gitignore
```diff
@@ -6,3 +6,8 @@ _static
 _templates
 
 TODO
+*.pyc
```
Julian: IMHO these all belong in your global gitignore, they're not specific to this project (which is why they weren't here before).

In case you don't already have one here's a sample of mine: https://github.com/Julian/dotfiles/blob/master/.config/git/ignore

jsonschema/_format.py
```diff
@@ -158,7 +158,7 @@ def is_host_name(instance):
 
 
 try:
-    import rfc3987
+    import rfc3986
```
Julian: We'd need to preserve backwards compatibility here.

So we'd need to first try to import `rfc3987.parse` and then fall back on `rfc3986.is_valid_uri`. The tests also should be updated to run under both modules (isolated from each other) to ensure that they always both pass the suite. This can be done fairly easily with a tox matrix I think. I can give some more guidance if need be.

setup.py
```diff
@@ -22,7 +22,7 @@
 ]
 
 extras_require = {
-    "format" : ["rfc3987", "strict-rfc3339", "webcolors"],
```
Julian: I think probably we'd need to support backwards compat here too. I assume you've already verified that strict-rfc3339 and webcolors are non-GPL?

If so I think we'd need to add a format-nongpl extra, otherwise we'll potentially change the deps being installed for people.

.gitignore
shakeelmohamed: That's fair, I actually didn't know a global gitignore was possible :smile:. I will revert those changes

jsonschema/_format.py
shakeelmohamed: Great, I can manage the code change - but I'll probably need guidance around tox since I'm not familiar with it

setup.py
shakeelmohamed: I didn't think of checking the other deps... `strict-rfc3339` is GPL as well so we'll need to find an alternative there, `webcolors` is BSD.

Totally reasonable for the backwards compatibility perspective.

shakeelmohamed:
Thanks for the quick feedback @Julian, looks like this will be a non-trivial set of changes - I appreciate your perspective on backwards compatibility and we should definitely strive for that.

shakeelmohamed:
We've decided not to use this project, so I'm closing this PR

angelcaido19:
@shakeelmohamed I would like to work on this issue. Shall I reopen this ticket or create a new one?

Julian:
A new PR is fine!

naimetti:
@Julian I have created a new (draft yet) PR about this issue: #619 
I would like to know your opinion about it.

Issue #263, opened 2015-11-30
Compatibility issue with cx_Freeze with Python 2.7


PierreRaybaut:
Currently, `jsonschema` breaks any attempt to deploy a Python script using `cx_Freeze`, a common tool for deploying self-consistent Python applications, (not only, but for example) on Windows platforms.

The source of the problem is in [jsonschema/compat.py](https://github.com/Julian/jsonschema/blob/master/jsonschema/compat.py#L5):
```python
try:
    from collections import MutableMapping, Sequence  # noqa
except ImportError:
    from collections.abc import MutableMapping, Sequence  # noqa
```

When commenting those lines, `cx_Freeze` works as expected.

The fact is `jsonschema` is part of the huge dependency list of `IPython`. `IPython` is very invasive when installed: even if not needed by your application, deploying with `cx_Freeze` will take all `IPython` related modules with it (and `jsonschema` is among them).
In other words, building the most insignificant Python script will fail only because `IPython` is installed (which happens a lot) and it depends on `jsonschema`.

See PR #264

Julian:
For anyone finding this: see the discussion on #264, which was a proposed fix (and a probably working one I assume :), for what seems like an upstream issue, but it'd need a test to be mergeable.

Julian:
It seems like the upstream issue was fixed, so I'm going to close this as "unsupported, but willing to support if the above is addressed."

PR #264, opened 2015-11-30
Fix cx_Freeze compatibility issue (Python 2.7)

PierreRaybaut:
See Issue #263

### DIFF
--- a/jsonschema/compat.py
+++ b/jsonschema/compat.py
@@ -1,16 +1,16 @@
 import operator
 import sys
 
-
-try:
-    from collections import MutableMapping, Sequence  # noqa
-except ImportError:
-    from collections.abc import MutableMapping, Sequence  # noqa
-
 PY3 = sys.version_info[0] >= 3
 PY26 = sys.version_info[:2] == (2, 6)
 
 if PY3:
+    try:
+        # Python <=3.2
+        from collections import MutableMapping, Sequence  # noqa
+    except ImportError:
+        # Python >=3.3
+        from collections.abc import MutableMapping, Sequence  # noqa
     zip = zip
     from functools import lru_cache
     from io import StringIO
@@ -22,6 +22,7 @@ if PY3:
     int_types = int,
     iteritems = operator.methodcaller("items")
 else:
+    from collections import MutableMapping, Sequence  # noqa
     from itertools import izip as zip  # noqa
     from StringIO import StringIO
     from urlparse import (


Julian:
Hi!

Thanks a lot for the PR, appreciate it.

I'm not sure I understand the specifics of why this should break cx_Freeze -- I have no problem supporting cx_Freeze I don't think, it's come up once before, but to do so we'd need to have tests in place to ensure that it continues to work. Do you have any ideas for how to do so?

PierreRaybaut:
The most basic test I ran to "debug" this is the following:
https://github.com/PierreRaybaut/guidata/blob/master/guidata/tests/disthelpers.py

But the problem is that cx_Freeze behavior depends on your Python installation: this bug occurs only with Python 2.7 and if the latest IPython is installed with all its dependencies.

The test above was executed from a newly installed WinPython 2.7.10.2 or 2.7.10.3.

So... not the most simple thing to test...

I'm not the only one having issues with this, see for example:
https://bitbucket.org/anthony_tuininga/cx_freeze/issues/127/collectionssys-error#comment-15021373

Anyway, the fact is that this PR seems quite transparent for the code (it's seems cleaner to me to separate the import statements between Python 2 and Python 3, and even clearer with the associated comments) and should not have any side effect except fixing issue #263.

Julian:
The issue you linked seems to say it's an upstream issue with cx_Freeze then?

> Anyway, the fact is that this PR seems quite transparent for the code (it's seems cleaner to me to separate the import statements between Python 2 and Python 3, and even clearer with the associated comments) and should not have any side effect except fixing issue #263.

That's somewhat less the point unfortunately -- the whole reason this change is being made is to support a thing that isn't "officially supported" because no tests ensure it works. If we can't get some testing to make sure it does, there's nothing preventing it from breaking tomorrow, in which case merging this PR doesn't really have much value IMHO.

Perhaps something as simple as spinning up a venv, installing jsonschema and ipython, and then running cx_Freeze would be enough, but there's clearly going to be some simpler way to boil down the bug, although it's true it might not be easy to find :(

PierreRaybaut:
Ok, I understand your reluctance. I already had a clean workaround when submitting this is issue, so this is not for me... I was just taking the time to ensure other people would not suffer from this IPython-jsonschema-cx_Freeze bug. And the fix seemed to be cleaner with the original implementation (the try...except statement was cleary there to make the distinction between Python 2 and Python 3, whereas there is an if...else statement just below doing the job nicely... I already tend to prefer explicit to implicit, but that's a personnal choice).

Julian:
Well it wasn't exactly :) as you can tell by virtue of you needing to replicate it even after you moved the import. jsonschema doesn't support 3.2 anymore, so technically it's not needed anymore, but that's why it was there. I don't care *too* much about minor style one way or the other, but yeah I am reluctant to merge patches that make it *look* like cx_Freeze is supported, without it actually being so. I'll leave this open in case anyone else manages to come up with a minimal test case.

Thanks again for the report.

PierreRaybaut:
No problem.

This bug is really weird anyway because the workaround consists in excluding "collections.abc" when calling cx_Freeze setup. In other words, to make it work with your current source code, the end user trying to deploy an application with Python *2.7* has to exclude a Python *3* builtin module ("collections.abc"). Very weird.

PR #272, opened 2016-01-29
re.compile can also raise other kinds of errors

Youwotma:
Wonder if it wouldn't be safer to just use Exception here

### DIFF
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -190,7 +190,7 @@ else:
             return strict_rfc3339.validate_rfc3339(instance)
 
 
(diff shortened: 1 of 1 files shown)


Youwotma:
Maybe should just use `Exception` since it might raise other different exceptions in the future and this is breaking the style checker. What do you think, @Julian ?

Julian:
Hi!

I'd prefer not to catch `Exception ` for the same reason as not really wanting to `except Exception `, which is basically what this turns into.

Would you be able to include test cases that exercise the other exceptions here?

Thanks!

Issue #283, opened 2016-05-17
ErrorTree TypeError: list indices must be integers or slices, not str


joepvandijken:
I have a problem using the ErrorTree with the attached example (second code piece).

I get a "TypeError: list indices must be integers or slices, not str".
When I change the ErrorTree.__init__ to (first code piece) it gives me the correct result, is this a bug or am I missing something, if it is a bug I'll gladly send a pull request as soon as I figure out how.

```python
   def __init__(self, errors=()):
        self.errors = {}
        self._contents = defaultdict(self.__class__)

        for error in errors:
            container = self
            for element in error.path:
                container = container[element]
            container.errors[error.validator] = error

            container._instance = error.instance  # was self._instance = error.instance
```

```python
import json
from jsonschema.exceptions import ValidationError, ErrorTree
from jsonschema.validators import validator_for, validate, Draft4Validator

schema = """\
{
    "$schema": "http://json-schema.org/draft-04/hyper-schema#",
    "type": "object",
    "properties": {
        "username": {
            "type": "string"
        },
        "secret": {
            "type": "object",
            "properties": {
                "password": {
                    "type": "string",
                    "minLength": 10,
                    "pattern": "^[abcdefgh]*$"
                }
            },
            "required": [
                "password"
                ]
        },
        "favorite_numbers": {
            "type": "array",
            "items": {
                "type": "integer"
            },
            "maxItems": 3
        }
    },
    "additionalProperties": false,
    "required": [
        "username",
        "secret"
    ]
}
"""


def errortree_to_dict(et):
    res = {k: errortree_to_dict(et[k]) for k in et}
    if len(res) == 0:
        return et.errors
    return res


if __name__ == "__main__":
    schema = json.loads(schema)
    instance = {"username": 'Henk', 'secret': {'password': 'gg'}, 'favorite_numbers': [1, 2, 1, 3]}
    
    v = Draft4Validator(schema)
    et = ErrorTree(v.iter_errors(instance))
    print(errortree_to_dict(et))
```

PR #288, opened 2016-06-14
Minor fix in ErrorTree for multiple errors on different levels.

joepvandijken:
https://github.com/Julian/jsonschema/issues/283

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -196,7 +196,7 @@ class ErrorTree(object):
                 container = container[element]
             container.errors[error.validator] = error
 
-            self._instance = error.instance
+            container._instance = error.instance
 
     def __contains__(self, index):
         """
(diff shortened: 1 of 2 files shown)


joepvandijken:
If anyone will verify this is an actual bug, I will also try to add a regression test for this case. Currently both versions pass all tests.

Julian:
Great, thanks! This does seem likely to be correct. A test would be great.

joepvandijken:
If you have any comments on the code let me know, I'm happy to address those.

Julian:
Thanks so much! I'll try to have a look in the next day or two and leave any comments.

Julian:
So sorry that this slipped so long. I have a comment or two, mainly revolving around that it'd be nice to try to use "real" looking data rather than "i1" and "i2", but despite that, I think the real solution is going to be whenever I have the time to make jsonschema's test suite use hypothesis (which at the current pace is far off :().

But it's certainly good enough to merge, and I'm sorry again about the wait.

Issue #208, opened 2015-03-16
cli.py should check_schema before using it.


Parakleta:
The documentation suggests "first verify that the provided schema is itself valid, since not doing so can lead to less obvious error messages and fail in less obvious or consistent ways" but the command line utility does not follow this advice.

If it is desirable to not recheck the schema on every run for performance reasons related to scripting it would be possible to instead check the schema just when the cli tool is called with no instances (which currently appears to be effectively a no-op), but I would prefer it on every run (since multiple instances can be provided in a single execution).

Julian:
Yup, agreed. Any chance you'd be interested in a PR?

Parakleta:
Not really.  There are several different ways the issue could be resolved and I don't think it's my place to determine which of those methods is correct, especially since I have no particular understanding of this project, or more than a passing familiarity with Python.

For my purpose because I am calling the validator from other code as well I have changed the `Validator.__init__` function, adding the line `schema is not self.META_SCHEMA and self.check_schema(schema)`.  I chose this because it is more convenient for me to know that any validator object has a legal schema attached, but I don't know how this fits into the scope of your project.

Going forward I think I am likely to abandon JSON schema in general anyway since I need more expressive power, meaning either XSD1.1 or a Lua DSL.

wspeirs:
Fixed in #297

PR #297, opened 2016-09-28
Made required_draft4 more robust when used with Draft3

wspeirs:
When I try to validate a Draft3 schema using the Draft4 validator, you get the following error:

```
  File "/usr/lib/python3/dist-packages/jsonschema/_validators.py", line 299, in required_draft4
    for property in required:
TypeError: 'bool' object is not iterable
```

My change _fixes_ this issue by alerting the use to the mis-match in draft versions.

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -296,6 +296,11 @@ def properties_draft4(validator, properties, instance, schema):
 def required_draft4(validator, required, instance, schema):
     if not validator.is_type(instance, "object"):
         return
+
+    if not isinstance(required, list):
+        yield ValidationError("Found Draft3 style 'required' restriction")
+        return
+
     for property in required:
         if property not in instance:
             yield ValidationError("%r is a required property" % property)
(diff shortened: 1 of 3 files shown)


Julian:
Hi! Thanks.

So, this would only happen if you didn't first verify you had a valid schema by using `Draft4Validator.check_schema` before you called `Draft4Validator(schema).validate` -- if you did, you'd see that the meta schema complains about having the wrong type there (which is why `jsonschema.validate calls that first before calling the method on the validator).

Generally, all the validators ask for and document that they need to be given valid schemas for their errors to be decipherable.

wspeirs:
I was presented this error when using the command line tool. So maybe the
real fix is to change that to call `check_schema` first?

Julian:
Yes! Definitely, that is issue #208 -- would definitely appreciate a patch for that.

wspeirs:
@Julian updated my PR to include a fix for #208 as well. If you want, you can cherry pick b44c1a2 but I think it's safe to pull in both changes.

Julian:
Seems like it's in the right place, can you add a test?

wspeirs:
@Julian added a unit test in 788ea8e

Julian:
Merged (at least the part that's correct here). Thanks -- really appreciate it.

PR #300, opened 2016-10-07
fix travis build failing as per https://github.com/travis-ci/travis-c…

simevo:


Julian:
Thanks, really appreciate you looking into this, I had it on my list to track it down at some point.

It does still look like the build is failing though. I think the change you added here added another travis build, but the actual change needed is to install the package into the existing one possibly.

simevo:
I have committed to my branch the fixes to the additional travis test failures that surfaced after the pull request - the price to pay is a dependency on the future library (https://python-future.org/)

I run tox on that and all tests pass (except the ones on pypy3 which is not installed here)

Julian:
Sorry. I guess I'm being unclear. I don't think any changes to the code should actually be necessary, though I haven't reviewed yours carefully. But the build is still failing, and it's because the one old travis build, which wants all interpreters to be present, still does not have access to a python 3.5. What I was trying to say was that it looks like you're adding a second independent build by adding `- 3.5` to the travis.yml, when the right solution is to install 3.5 into the existing build, which `tox` is already parametrizing.

Lemme know if this makes sense.

simevo:
About the code changes, those are possibly related to the particular environment here (debian 9 stretch amd64).
About travis, what about keeping the two independent travis builds, and running the tests with some other test runner, that only runs against the platform python ? In this way the 2.7 travis build would run against python 2.7 and on the 3.5 build against python 3.5. Tox would still be useful during local development.

Julian:
> About the code changes, those are possibly related to the particular environment here (debian 9 stretch amd64).
Are you saying you're getting test failures when running locally? If you are please open a (separate) issue with traceback + details. But it doesn't look like from those changes that you should be, those blocks of code shouldn't run on Py3.

> About travis, what about keeping the two independent travis builds, and running the tests with some other test runner, that only runs against the platform python ? In this way the 2.7 travis build would run against python 2.7 and on the 3.5 build against python 3.5. Tox would still be useful during local development.
I don't care too much one way or the other here, it's more important that things are consistent -- right now, tox is responsible for parametrization. Using travis for parametrization is another strategy (I generally prefer this way, but it doesn't matter too much if you use TOXENV) but one way or the other all the builds should parametrize the same way I think. Which means either all the builds should change to use TOXENV and travis Python versions, or things should stay as-is, and like I said, there's probably just like a `aptitude install python3.5` or something that you need to tell travis to install one additional interpreter in `travis.yml`.

simevo:
seems to be fixed using tox-travis: https://github.com/ryanhiebert/tox-travis

Julian:
That change now appears to make it so that the *only* toxenvs that run are the ones using those two interpreters.

Specifically for example, you've now disabled the pypy builds.

simevo:
there is definitely a duplication of work in travis and tox but it works

Julian:
This still doesn't run all of the builds anymore. I appreciate the work definitely! But I'm just going to apply the solution [I mentioned](https://github.com/Julian/jsonschema/commit/e5738b3c398f11866717d1bde8bf5297f8442e92) and close this out.

simevo:
You're right ! with my approach the docs and style tests are never executed. Your solution is better.

PR #301, opened 2016-10-11
python 3 compatibility in the example code

strets123:


Julian:
I don't think compatibility code should make it into docs, it's distracting. If we really wanted to write compatible code in this particular case it seems better to just use `.items` , though as a general rule personally I think it's fine to just write code in one version.

Julian:
Gonna close this out, but if you'd like to discuss further definitely lemme know.

PR #303, opened 2016-11-03
Fix indentation

drpoggi:
If it was intentional to line up the functions for readability then disregard, but this actually hurt readability for me as the double indentation caught me off guard and my immediate thought was "that can't be right." Well TIL, just thought I'd propose this change in case this was an accident.


Julian:
Ah no :) wasn't intentional, although yeah Python doesn't care about how far you indent a block as long as you are consistent -- but yes, thanks, this is better!

PR #306, opened 2016-11-20
Fix relative references

mathewmarcus:
Currently, calling the resolve_fragment method of the RefResolver class to return the JSON referenced by the pointer of a $ref key fails if the pointer is preceded by a `#` character. The JSON schema specs specify this format, e.g. { "$ref": "#/definitions/address" }. Specifically, the following exception is raised: jsonschema.exceptions.RefResolutionError: Unresolvable JSON pointer. This happens even if the pointer is valid.  

I made changes to the resolve_fragment method of the RefResolver class in jsonschema/validators.py to strip away this character, and now the pointer resolves correctly. I also added two tests to jsonschema/tests/test_validators.py to demonstrate this behavior of the old method and to show the results of the revised method. Let me know if you have any questions.

### DIFF
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -6,8 +6,9 @@ from jsonschema import FormatChecker, ValidationError
 from jsonschema.tests.compat import mock, unittest
 from jsonschema.validators import (
     RefResolutionError, UnknownType, Draft3Validator,
-    Draft4Validator, RefResolver, create, extend, validator_for, validate,
+    Draft4Validator, RefResolver, create, extend, validator_for, validate
 )
+from jsonschema.compat import Sequence, unquote
 
 
 class TestCreateAndExtend(unittest.TestCase):
@@ -771,6 +772,28 @@ class TestRefResolver(unittest.TestCase):
     stored_uri = "foo://stored"
     stored_schema = {"stored" : "schema"}
 
+    def prev_resolve_fragment(self, document, fragment):
+        fragment = fragment.lstrip(u"/")
+        parts = unquote(fragment).split(u"/") if fragment else []
+
+        for part in parts:
+            part = part.replace(u"~1", u"/").replace(u"~0", u"~")
+
+            if isinstance(document, Sequence):
+                # Array indexes should be turned into integers
+                try:
+                    part = int(part)
+                except ValueError:
+                    pass
+            try:
+                document = document[part]
+            except (TypeError, LookupError):
+                raise RefResolutionError(
+                    "Unresolvable JSON pointer: %r" % fragment
+                )
+
+        return document
+
     def setUp(self):
         self.referrer = {}
         self.store = {self.stored_uri : self.stored_schema}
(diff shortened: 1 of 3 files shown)


Julian:
Awesome, really appreciate a patch here, will try to have a look and leave some initial comments -- will try to read this a bit more carefully as well in a bit.

jsonschema/tests/test_validators.py
```diff
@@ -771,6 +772,28 @@ class TestRefResolver(unittest.TestCase):
     stored_uri = "foo://stored"
     stored_schema = {"stored" : "schema"}
 
+    def prev_resolve_fragment(self, document, fragment):
+        fragment = fragment.lstrip(u"/")
```
Julian: This seems like quite a lot of not-testing-specific-looking code making an appearance in a test. I don't have a specific recommendation yet on how to remove it, but it doesn't look right to me, especially not in combination with being patched into place (I'd like to sometime soon remove the `mock` dependency entirely, but alas).

jsonschema/tests/test_validators.py
```diff
@@ -898,6 +921,77 @@ def test_helpful_error_message_on_failed_pop_scope(self):
             resolver.pop_scope()
         self.assertIn("Failed to pop the scope", str(exc.exception))
 
+    def test_prev_resolve_fragment_raises_exception(self):
+        schema = {
+            "definitions": {
```
Julian: This schema seems like it should be easy to simplify.

jsonschema/tests/test_validators.py
```diff
@@ -898,6 +921,77 @@ def test_helpful_error_message_on_failed_pop_scope(self):
...
+                        "id": {
+                            "type": "integer",
+                            "format": "int32"
+                        }
+                    }
+                }
+            }
+        }
+        resolver = RefResolver.from_schema(schema)
+        self.assertEqual(resolver.resolve_fragment(schema, "#/definitions/Category"), schema['definitions']['Category'])
```
Julian: I assume that what this is trying to fix is probably validation related, so would be nice to have at least one actual integration test (that uses a relative reference in a schema for validation) -- especially so, since if there's a bug here, it means we're missing a test in the [official test suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite)

jsonschema/tests/test_validators.py
mathewmarcus: Ok sorry about that, I can remove some of the unnecessary parts of that method. I only put them in there to replicate the original method as closely as possible.

jsonschema/validators.py
```diff
@@ -356,7 +356,7 @@ def resolve_fragment(self, document, fragment):
 
         """
 
-        fragment = fragment.lstrip(u"/")
+        fragment = fragment.lstrip(u"#").lstrip(u"/")
```
Julian: This seems likely to be misplaced overall to me (which maybe is another reason why an integration test would be good).

This method is about resolving fragments -- fragments don't start with `#`, `#` delimits a fragment, so whoever called this should have removed the `#`

mathewmarcus: Got it, my mistake

mathewmarcus: Would you want functionality, perhaps in another method, to convert a pointer into a fragment, by stripping the `#`, etc, or do you want to close the request?

PR #309, opened 2016-11-27
Fix "tox -e coverage" and add .gitignore

pawl:
The coverage testenv uses green, so I added it to the dependencies for that testenv. I also added the basics to the .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,15 @@ _cache
 _static
 _templates
 
-TODO
+# Byte-compiled / optimized / DLL files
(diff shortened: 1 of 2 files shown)


Julian:
Er, thanks! Fixed in a slightly different way -- you can tell I don't really use `tox -e coverage` much :P, I just generally run coverage myself.

As for the `.gitignore` changes, I'd prefer keeping that in your global gitignore -- they're things that should be ignored always, so it's tiring to copy that across every (Python) repository.

PR #317, opened 2017-01-26
Better error message for additionalProperties: false when there are patternProperties

### LINKED ISSUE #316: Error message on patternProperties could be more informative
Consider the following schema snippet:

```yaml
     templates:
        type: object
        additionalProperties: false
        patternProperties:
          "^[a-zA-Z][-a-zA-Z0-9]{0,29}$":
            $ref: "#/definitions/docker-component"
```

If there is a key in the `templates` JSON object that doesn't match that regex the error message looks like this (I'm formatting this with `error.path` + `error.message`):

```
/templates: Additional properties are not allowed ("postgresql-v9'6" was unexpected)
```

This is not very informative to the reader. Here's my proposal for a better error message:

```
/templates: "postgresql-v9'6" does not match any of these regexs: ^[a-zA-Z][-a-zA-Z0-9]{0,29}$.
```

itamarst:
Fixes #316

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -29,8 +29,18 @@ def additionalProperties(validator, aP, instance, schema):
             for error in validator.descend(instance[extra], aP, path=extra):
                 yield error
     elif not aP and extras:
-        error = "Additional properties are not allowed (%s %s unexpected)"
-        yield ValidationError(error % _utils.extras_msg(extras))
+        if "patternProperties" in schema:
+            patterns = sorted(schema["patternProperties"].keys())
+            if len(extras) == 1:
+                verb = "does"
+            else:
+                verb = "do"
+            error = "%s %s not match any of the regexs: %s" % (
+                ", ".join(map(repr, sorted(extras))), verb, ", ".join(patterns))
+            yield ValidationError(error)
+        else:
+            error = "Additional properties are not allowed (%s %s unexpected)"
+            yield ValidationError(error % _utils.extras_msg(extras))
 
 
 def items(validator, items, instance, schema):
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -30,13 +30,16 @@ def additionalProperties(validator, aP, instance, schema):
                 yield error
     elif not aP and extras:
         if "patternProperties" in schema:
-            patterns = sorted(schema["patternProperties"].keys())
+            patterns = sorted(schema["patternProperties"])
             if len(extras) == 1:
                 verb = "does"
             else:
                 verb = "do"
-            error = "%s %s not match any of the regexs: %s" % (
-                ", ".join(map(repr, sorted(extras))), verb, ", ".join(patterns))
+            error = "%s %s not match any of the regexes: %s" % (
+                ", ".join(map(repr, sorted(extras))),
+                verb,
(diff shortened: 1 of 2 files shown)


jsonschema/_validators.py
```diff
@@ -29,8 +29,18 @@ def additionalProperties(validator, aP, instance, schema):
             for error in validator.descend(instance[extra], aP, path=extra):
                 yield error
     elif not aP and extras:
-        error = "Additional properties are not allowed (%s %s unexpected)"
-        yield ValidationError(error % _utils.extras_msg(extras))
+        if "patternProperties" in schema:
+            patterns = sorted(schema["patternProperties"].keys())
```
Julian: Can leave off the `.keys()`

jsonschema/_validators.py
```diff
@@ -29,8 +29,18 @@ def additionalProperties(validator, aP, instance, schema):
...
+            if len(extras) == 1:
+                verb = "does"
+            else:
+                verb = "do"
+            error = "%s %s not match any of the regexs: %s" % (
```
Julian: I think it might be `regexes` :/?

jsonschema/tests/test_validators.py
```diff
@@ -191,6 +191,26 @@ def test_invalid_format_default_message(self):
         self.assertIn(repr("thing"), message)
         self.assertIn("is not a", message)
 
+    def test_additionalProperties_false_patternProperties(self):
+        schema = {u"type": u"object",
+                  u"additionalProperties": False,
+                  u"patternProperties": {
+                      u"$abc^": {u"type": u"string"},
```
Julian: Nitpick: can you change these to slightly more realistic regexes, e.g. `^abc$` :)?

jsonschema/tests/test_validators.py
```diff
@@ -191,6 +191,26 @@ def test_invalid_format_default_message(self):
...
+                      u"$def^": {u"type": u"string"}
+                  }}
+        message = self.message_for({u"zebra": 123}, schema,
+                                   cls=Draft4Validator)
+        self.assertEqual(
+            message,
+            "{} does not match any of the regexs: $abc^, $def^".format(
+                repr(u"zebra")))
```
Julian: Nitpick: can you change the formatting here to not combine parens, e.g.

```
self.assertEqual(
    message,
    "whawefiuhweifuhiweuhfiwuehfiuwheftever".format(
        repr(stuff),
    ),
)
```

so that each one gets a trailing comma?

jsonschema/_validators.py
```diff
@@ -29,8 +29,18 @@ def additionalProperties(validator, aP, instance, schema):
...
+                ", ".join(map(repr, sorted(extras))), verb, ", ".join(patterns))
```
Julian: Any reason to `map` `repr` for the first argument but not for the patterns?

Still would be a bit confusing I think if patterns ended up having the empty string in it.

itamarst:
OK, updated.

itamarst:
Travis CI issue appears to be unrelated at this point.

Julian:
Bleh, yeah, OK, should have Travis CI fixed now, PyPy3 fun.

Thanks though! Looks good, merging, appreciated.

Now for a release...

PR #330, opened 2017-04-20
Allowed function-like type validators

KOLANICH:
Note ```type_backup```. Your tests are messing with ```type``` somewhere, so I had to do this. I won't fix that, so don't forget to replace type_backup with type when you fix that.

### DIFF
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -17,6 +17,7 @@ from jsonschema.compat import (
 from jsonschema.exceptions import ErrorTree  # Backwards compat  # noqa: F401
 from jsonschema.exceptions import RefResolutionError, SchemaError, UnknownType
 
+type_backup = type
 
 _unset = _utils.Unset()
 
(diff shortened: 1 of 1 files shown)


Julian:
Hi. Thanks.

This change is missing context, tests, and documentation :).

Context is probably most important. Can you describe what functionality this changeset is trying to add, or since I can mostly tell, why adding that functionality in this way is a good idea?

My first inclination is that there are other, better ways to implement this without overloading the type validator.

KOLANICH:
20 апреля 2017 г. 14:47:46 GMT+03:00, Julian Berman <notifications@github.com> пишет:
>Hi. Thanks.
>
>This change is missing context, tests, and documentation :).
>
>Context is probably most important. Can you describe what functionality
>this changeset is trying to add, or since I can mostly tell, why adding
>that functionality in this way is a good idea?
>
>My first inclination is that there are other, better ways to implement
>this without overloading the type validator.

> Can you describe what functionality this changeset is trying to add

In fact I only wanted to add a type for functions and function-like objects into a schema. To check if a type is callable you have to call ```callable()```. But current code uses instanceof only. I guess there can be more types checked with functions.

Julian:
Have you considered using the `format` validator instead, which already supports this functionality?

The `type` validator maps very "naturally" onto, well, types :)

KOLANICH:
No, haven't considered that.

Julian:
Gonna close this out, give that a shot -- feel free to reopen if you'd like to discuss further.

PR #338, opened 2017-06-08
Dedent examples in validate.rst to fix sphinx doctest

dlax:


Julian:
Thanks for sending this over.

I still haven't bothered researching what the heck Sphinx did in the last release, they seem to have split a bunch of things out and made some backwards incompatible changes and it all seems kind of weird and like I should go hunt down the rationales, but this does appear to fix the build, appreciated!

Will merge as soon as I get to a computer.

Thanks again!

PR #342, opened 2017-07-03
Improved email validation

Nberezhnoy:
Hey. I improved the email validation method. It is based on the regular expression from [Ajv](https://github.com/epoberezkin/ajv)

### DIFF
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -147,11 +147,16 @@ def _checks_drafts(both=None, draft3=None, draft4=None, raises=()):
     return wrap
 
 
+_email_re = re.compile(r"^[a-z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)*$")
+
+
 @_checks_drafts("email")
 def is_email(instance):
     if not isinstance(instance, str_types):
         return True
-    return "@" in instance
+    if not _email_re.match(instance):
+        return False
+    return True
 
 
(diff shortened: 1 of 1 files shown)


Julian:
@Nberezhnoy I could have sworn I replied to this, really sorry I apparently haven't :/

So, I'm slightly conflicted here. The current implementation, which yeah literally is just `@ in s`, is intentionally simplistic -- as an end-user of The Internet, email validation is a blight -- it doesn't work -- both false positives and false negatives -- and no one does it correctly, because the full implementation of it is extremely complex and has lots of seemingly invalid emails that are actually totally OK, plus it doesn't accomplish what people often actually want, which is "will this email get to the human I want it to get to", for which you just, well, send the email and cross your fingers.

So this was basically the one deviation from the spec in this entire library.

That regex you've pasted, which ajv looks like it gets from a w3 document, itself acknowledges that it also contains false negatives and positives, so I worry about users' expectations. My position until now has been that an easily explainable thing was better than the complex thing, given that neither particularly accomplish much. (I see ajv has a fuller form of the actual regex too)

Lemme think a bit more about it, but I welcome comments from anyone with an opinion swayed one way or the other.

CC... @handrews / @epoberezkin possibly.

handrews:
@Julian I've wondered before if we should change the `"format"` implementation requirements from "if you implement it, you should implement these formats" which is a boolean requirement (do you or don't you validate formats?) to more of a best effort thing.

"email" may be the worst of the lot, but I know ajv offers various speed/completeness tradeoffs and I'm pretty sure that you're right that no one fully implements "email" correctly.  Not all of these can be done with regexes at all.

Julian:
@handrews yep agreed

Julian:
@Nberezhnoy I'm going to go with my instinct and close this, but happy to discuss further if you disagree.

And yeah @handrews I think we do need to address this at the spec level.

PR #353, opened 2017-08-26
Improve travis and tox build

pgjones:
This will hopefully add Python 3.3, 3.4, and 3.6 testing as well as
simplify the travis management via the tox-travis tool.


Julian:
Nice, thanks!

PR #354, opened 2017-08-28
Include the JSON-Schema-Test-Suite as a submodule

pgjones:
This is instead of merging the repositories together as it makes the
test suite easier to manage (no merging) and follows the convention
of many of the other validators that use it.


Julian:
Hey! Thanks!

I prefer `git subtree` to `git submodule` -- it's easier for end users (and I've found it equal for developers), who don't need to know that the subrepo is external (FWIW I was the original author of the test suite, and I'd recommend other validators to follow *this* convention :D although it doesn't make too big of a difference)

Can you explain a bit more what you mean about making the test suite easier to manage without merging? Happy to be persuaded certainly.

pgjones:
No problem, sounds like you have a preference. I saw it in a number of the JS validators and thought it was helpful.

PR #355, opened 2017-08-28
Change benchmark scripts to use perf

pgjones:
This includes adding a benchmark suite that benchmarks every valid
test in the JSON-Schema-Test-Suite. This should make the benchmark
scripts more useful.

Not sure if this is what you are looking for, but hopefully it serves as a starting point.


Julian:
This looks likely to be a fantastic start, really appreciated!

I'm gonna play around a bit with what you have here and leave a few comments!

Julian:
Hey @pgjones I wanted to thank you (and apologize for this taking so long!)

I just got off a very long plane ride, which means I had some time to make progress here :)

I made some (significant) changes to you had, mostly revolving around factoring out the loading of test suite cases so that the code wasn't duplicated here as you had it, but I used all of this as the first iteration.

It's on a branch called `perf`, which I hope to merge as soon as I figure out what we should be doing to publish those results / run them on pull requests. Certainly suggestions welcome, and again thanks!

Julian:
OK, this finally should be merged, even though I don't have a good answer to the above. Will have to keep thinking about it, but for now at least we can move forward. Thanks again!

pgjones:
Thanks @Julian sadly I've no suggestions about where to publish the results as I couldn't find a good solution.

PR #356, opened 2017-08-28
Use ujson if it installed.

AndrewBurdyug:
Hi! 

I think we can improve speed of json load/dump operations by using ujson, please see: https://github.com/esnme/ultrajson

### DIFF
--- a/benchmarks/bench.py
+++ b/benchmarks/bench.py
@@ -12,7 +12,10 @@ Example benchmark:
 from __future__ import print_function
 import argparse
 import cProfile
-import json
+try:
+    import ujson as json
+except:
+    import json
 import time
 
 import jsonschema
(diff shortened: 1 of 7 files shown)


Julian:
Hi, thanks for sending this over!

I have a few concerns, most importantly that jsonschema does not really do any performance-critical json loading or dumping. Client code is where json loading should happen, as you can see from your patchset, pretty much all of what was changed was in tests or the CLI, which isn't too performance-sensitive, or when loading a metaschema, which should only happen once per process really.

More minor-ly, ujson is likely to only be faster on CPython, not PyPy, and it'd also be nice to have benchmarks proving it to be faster for typical jsonschema workloads before merging.

But due to the above, I'm not sure it's really any gain -- lemme know if you disagree though, certainly appreciate the PR.

Julian:
Gonna close this, but lemme know if you disagree, happy to discuss further!

Issue #364, opened 2017-10-09
Support namedtuple


ejconlon:
Though `json` doesn't treat `namedtuple` instances as objects, `simplejson` does.  It would be useful to be able to validate object graphs with `namedtuple` instances like `dict` instances.

The main complication here is that `Validator`'s `is_type` and `DEFAULT_TYPES` and very subclass-oriented, and detecting a `namedtuple` is a matter of duck-typing (`hasattr(instance.__class__, '_fields')` or the less-restrictive `hasattr(instance, '_asdict')`).  From there I'm not quite sure where to make the change to traverse object items.

Thanks for your consideration.

Julian:
:/ that makes me a bit sad to see, but I've already been open to the idea that is_type's interface should be wider (it started narrow to drive out needing it to be wide, but it's likely getting to the point where now it needs to be -- draft 6 already poses some issues there).

Will have to think about possible ways to widen it but I think this sounds reasonable. Suggestions welcome obviously, as well as certainly patches.

bsmithers:
Just moving the conversation from the PR (https://github.com/Julian/jsonschema/pull/370#discussion_r149834435) here to keep things more contained.

I've had a little look at your suggestions [1]. I can certainly see the benefit in immutability and a `TypeChecker` object feels like a good approach. I think the interface cannot be quite as simple as it first seems though.

Although `type_checker.redefine(**types)` looks nice, it can't handle non-string types [2]. Another minor issue is that it doesn't allow for the removal of a type check (e.g. `any` was a valid type in draft3). So I've implemented it as `type_checker.update(redefine=(), remove=())` where `redefine` is a dictionary of checks to add or update and `remove` is any iterable.

 The `namedtuple` case is then the slightly more verbose:
```
type_checker = Draft6Validator.TYPE_CHECKER.update(redefine={
            u"object": object_or_named_tuple
        })
v = Draft6Validator(schema, type_checker=type_checker)
```

Some additional complications:
1. Since the old type checking was via `isinstance`, `type_checker` tests if has been given a `type` (or tuple of types) and invokes the legacy type checking in this case. This is at least self-contained and could be removed in future
2. `type_checker` cannot raise a proper `UnknownType` exception as it doesn't have the schema. Currently I've handled this by raising a secondary exception, which is re-raised as an `UnknownType` exception by the Validator
[...]

Issue #346, opened 2017-07-19
Ref v2 Spec


Julian:
*This is some random notes on a new version of the ref resolution API. Feel free to leave comments, although some of the below might be overly terse chicken scratch*

- [ ] It should probably be easier to pre-resolve all refs (maybe that should even be the default) rather than lazy/forced resolution
- [ ] Keep in mind async use cases
- [ ] Too many methods bruh. Just 2? `Resolver().resolve(ref)` + `Resolver().for_scope(url)`?
- [ ] Take hyperlink.URLs, not strs

bsmithers:
Hey. I was looking at some of the tests that are currently skipped as known failures [1]

I poked this a little and realised that the id field isn't handled when it is is encountered in fragment resolution. E.g. in [2], there is a self-reference `{"$ref": "#/definitions/baz/definitions/bar"}` where under `/definitions/baz` the scope is updated. 

I haven't had time to work out if there are any ill-effects but one possible solution could be to return a new scope during fragment resolution if an id field is present. Is this something that is worth exploring separately or do you still plan to rework the resolver? 

[1] https://github.com/Julian/jsonschema/blob/master/jsonschema/tests/test_jsonschema_test_suite.py#L311-L318

[2] https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/master/tests/draft4/refRemote.json#L103-L135

Julian:
I am all for an incremental fix if it's small and self contained! Would definitely be appreciated. Just have to be careful to preserve backwards compatibility for any custom defined RefResolvers, but assuming that's not an issue, patch would be great! And we can worry about making things cleaner as a separate item (this ticket).

agoose77:
I've just been thinking about the problem of following $id declarations during ref resolution and came across this issue.

I don't think that we should recursively dereference schemas. Whilst it's certainly a simpler approach (at least, conceptually), I think the extra cost should be avoided (especially if there are schemas with lazy reference paths (though I'm not sure if currently jsonschema does this, as even in the case of oneOf we need to ensure _only_ one subschema validates (and hence we have to dereference it anyway)).

I think a better approach would be to follow the $id uri from within the resolution process, such as the approach (taken here)[https://github.com/gazpachoking/jsonref]. I'm not convinced that we could (or should) just use this repository (for a start, it doesn't support the newer $id keyword, but the approach is valid enough (we already use context managers to track the current scope). 

I'll give it some thought.

PR #370, opened 2017-11-05
Fixes the currently failing json-schema-test-suite tests

bsmithers:
Currently using jsonschema at work and interested in getting draft6 support finalised. To jump in, I've provided a couple of quick fixes for the previously failing tests. 

Before:
`FAILED (skips=47, failures=3, errors=10, successes=1209)`
After:
`PASSED (skips=47, successes=1254)`

Note that I pulled in the latest tests from the JSON schema test suit repo. 

I was hesitant to have `RefResolver` needing to know about id vs $id, but I could see no other way without removing `RefResolver.from_schema()` from the interface. I also briefly looked at the tests being skipped, which appear to require `RefResolver` to be able to parse the id attribute, but I'll comment on that separately.

### DIFF
--- a/json/README.md
+++ b/json/README.md
@@ -155,6 +155,7 @@ for more information.
 ### PostgreSQL ###
 
 * [postgres-json-schema](https://github.com/gavinwahl/postgres-json-schema)
+* [is_jsonb_valid](https://github.com/furstenheim/is_jsonb_valid)
 
 If you use it as well, please fork and send a pull request adding yourself to
 the list :).
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -296,6 +296,8 @@ else:
         draft6="json-pointer", raises=jsonpointer.JsonPointerException,
     )
     def is_json_pointer(instance):
+        if not isinstance(instance, str_types):
+            return True
         return jsonpointer.JsonPointer(instance)
 
 
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -391,6 +391,15 @@ def type(validator, types, instance, schema):
         yield ValidationError(_utils.types_msg(instance, types))
 
 
+def type_draft6(validator, types, instance, schema):
+    # Draft 6 changes integer type to include floats with no fractional part
+    if isinstance(instance, float) and "integer" in types and instance.is_integer():
+        return
+
+    for e in type(validator, types, instance, schema):
+        yield e
+
+
 def properties(validator, properties, instance, schema):
     if not validator.is_type(instance, "object"):
         return
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -56,7 +56,7 @@ def validates(version):
     return _validates
 
 
-def create(meta_schema, validators=(), version=None, default_types=None):
+def create(meta_schema, validators=(), version=None, default_types=None, scope_key="id"):
     if default_types is None:
         default_types = {
             u"array": list, u"boolean": bool, u"integer": int_types,
@@ -68,6 +68,7 @@ def create(meta_schema, validators=(), version=None, default_types=None):
         VALIDATORS = dict(validators)
         META_SCHEMA = dict(meta_schema)
         DEFAULT_TYPES = dict(default_types)
+        SCOPE_KEY = scope_key
 
         def __init__(
             self, schema, types=(), resolver=None, format_checker=None,
@@ -76,12 +77,13 @@ def create(meta_schema, validators=(), version=None, default_types=None):
             self._types.update(types)
 
             if resolver is None:
+                _schema = schema
                 if schema is True:
-                    resolver = RefResolver.from_schema({})
+                    _schema = {}
                 elif schema is False:
-                    resolver = RefResolver.from_schema({"not": {}})
-                else:
-                    resolver = RefResolver.from_schema(schema)
+                    _schema = {"not": {}}
+
+                resolver = RefResolver(_schema.get(scope_key, u""), _schema)
 
             self.resolver = resolver
             self.format_checker = format_checker
@@ -101,7 +103,8 @@ def create(meta_schema, validators=(), version=None, default_types=None):
             elif _schema is False:
                 _schema = {"not": {}}
 
-            scope = _schema.get(u"$id")
+            scope = _schema.get(scope_key)
+
             if scope:
                 self.resolver.push_scope(scope)
             try:
@@ -149,11 +152,8 @@ def create(meta_schema, validators=(), version=None, default_types=None):
                 raise UnknownType(type, instance, self.schema)
             pytypes = self._types[type]
 
-            # FIXME: draft < 6
-            if isinstance(instance, float) and type == "integer":
-                return instance.is_integer()
             # bool inherits from int, so ensure bools aren't reported as ints
-            elif isinstance(instance, bool):
+            if isinstance(instance, bool):
                 pytypes = _utils.flatten(pytypes)
                 is_number = any(
                     issubclass(pytype, numbers.Number) for pytype in pytypes
@@ -181,6 +181,7 @@ def extend(validator, validators, version=None):
         validators=all_validators,
         version=version,
         default_types=validator.DEFAULT_TYPES,
+        scope_key=validator.SCOPE_KEY
     )
 
 
@@ -279,10 +280,11 @@ Draft6Validator = create(
         u"properties": _validators.properties,
         u"propertyNames": _validators.propertyNames,
         u"required": _validators.required,
-        u"type": _validators.type,
+        u"type": _validators.type_draft6,
         u"uniqueItems": _validators.uniqueItems,
     },
     version="draft6",
+    scope_key="$id",
 )
 
 
@@ -356,7 +358,7 @@ class RefResolver(object):
         self._remote_cache = remote_cache
 
     @classmethod
-    def from_schema(cls, schema, *args, **kwargs):
+    def from_schema(cls, schema, scope_key="id", *args, **kwargs):
         """
         Construct a resolver from a JSON schema object.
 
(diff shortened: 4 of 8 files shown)


Julian:
Hey! Awesome, thanks so much for trying to move this along, these 3 things (draft 6, ref v2, and perf) have all hit little walls which has left them hanging there :/, so definitely definitely appreciate some help moving them to completion.

Gonna have to leave some detailed comments, for now just doing a first pass.

json/README.md
```diff
@@ -155,6 +155,7 @@ for more information.
 ### PostgreSQL ###
 
 * [postgres-json-schema](https://github.com/gavinwahl/postgres-json-schema)
+* [is_jsonb_valid](https://github.com/furstenheim/is_jsonb_valid)
```
Julian: I'm gonna just merge this immediately (the squash of the upstream changes)

jsonschema/validators.py
```diff
@@ -56,7 +56,7 @@ def _validates(cls):
     return _validates
 
 
-def create(meta_schema, validators=(), version=None, default_types=None):
+def create(meta_schema, validators=(), version=None, default_types=None, scope_key="id"):
```
Julian: So, this makes me a bit uneasy (which is why I didn't do it immediately myself).

I think to be safe (future-proof) we need to make the interface a bit wider -- i.e., rather than just specifying the name of a key to use for scoping, we should provide something like a function here that the validator will use to detect scope (and yeah to make the current implementations just be lookups). I'm worried about the potential for some more complicated process, and once we already need an argument to do it, might as well have it be a generic one. Lemme know if that makes sense.

We also need to have direct tests for this (i.e. separate from any specific draft support).

jsonschema/validators.py
```diff
@@ -76,12 +77,13 @@ def __init__(
             self._types.update(types)
 
             if resolver is None:
+                _schema = schema
                 if schema is True:
-                    resolver = RefResolver.from_schema({})
+                    _schema = {}
```
Julian: The problem here is that now you will mis-report schemas.

One major gap in draft 6 support is writing some tests for the new schema coersion behavior (of bools) -- specifically, this will misreport what the schema was in `ValidationError.schema`.

jsonschema/_validators.py
```diff
@@ -391,6 +391,15 @@ def type(validator, types, instance, schema):
         yield ValidationError(_utils.types_msg(instance, types))
 
 
+def type_draft6(validator, types, instance, schema):
```
Julian: Here too I think draft 6 taught us we need to widen the type interface. I put some notes in #364, maybe have a look there if you don't mind?

jsonschema/validators.py
```diff
@@ -356,7 +358,7 @@ def __init__(
         self._remote_cache = remote_cache
 
     @classmethod
-    def from_schema(cls, schema, *args, **kwargs):
+    def from_schema(cls, schema, scope_key="id", *args, **kwargs):
```
Julian: So, this also makes me uneasy -- up until now, ref resolvers were completely separate from drafts (because the behavior was the same across them).

I don't love this idea -- now a caller needs to know implementation details of the draft they're using.

I'd much rather encapsulate that, even if that means that we now need to have separate ref resolvers per draft (and we'd also need to preserve exactly the old interface for a class called just `RefResolver`). Annoying, but I think the whole change is annoying :/

setup.py
```diff
@@ -28,7 +28,7 @@
         "jsonpointer>1.13",
         "rfc3987",
         "strict-rfc3339",
-        "uritemplate>3.0.0",
+        "uritemplate>=3.0.0",
```
Julian: I know that the > was failing here, but that's because the patch with this wasn't merged  yet (and in fact 3.0.0 doesn't have what we need).

So I think it was correct as-is [and probably even needs some more work once the upstream PR is merged].

Maybe you can chime in upstream with any comments? The PR is https://github.com/python-hyper/uritemplate/pull/36

jsonschema/_format.py
```diff
@@ -296,6 +296,8 @@ def is_css3_color(instance):
         draft6="json-pointer", raises=jsonpointer.JsonPointerException,
     )
     def is_json_pointer(instance):
+        if not isinstance(instance, str_types):
```
Julian: This needs a test :)

jsonschema/validators.py
bsmithers: > So, this also makes me uneasy -- up until now, ref resolvers were completely separate from drafts (because the behavior was the same across them).

The problem is that they're not, because that `from_schema` method needs to understand id. My initial thought was actually to drop that method completely, but of course that's extremely unpleasant for users in the wild.

jsonschema/_format.py
bsmithers: It already does - the test fail was what caused me to fix :)

Julian: Ah cool. Hm, I thought I'd ran these after updating. Thanks for fixing.

jsonschema/validators.py
Julian: Right but what I mean is -- as far as `RefResolver` is concerned, when presented with any schema, there is an algorithm you can use to unambiguously know this (but now this introduces the possibility of error).

Even now, you have this algorithm:

* Look for a `$schema` key
* If you find one, you (RefResolver) know what the correct scope key is
* If not, you use the "package" default (which I'm not too happy with how we're managing currently :/, but it's a thing)

You can take that algorithm, and layer it on top of an "explicit" one -- say a `Draft3Draft4RefResolver` and a `Draft6RefResolver` alongside a `resolver_for_schema` that calls the right thing as above.

(I know I'm being terse still, lemme know if that's unclear)

jsonschema/validators.py
bsmithers: > Lemme know if that makes sense.
You mean specifying a validator for `id` or `$id` to handle the scope change?

I did consider that, however I wasn't sure the corresponding architectural change would be nice and wondered if a simpler solution is better at least until such time as there are future differences. But I see your point.

> We also need to have direct tests for this (i.e. separate from any specific draft support).
Agreed on that.

Julian: I think it would be slightly narrower than a validator -- basically it is a function to replace the calls just to `someschema.get("id")` -- so, `scope_of(schema)`, say.

Julian: And really, yeah this is tightly related to #346 unfortunately, which is to say, I think the even better thing is to come up with the right interface for *that*, provide the *RefResolver* class you want to use to this function, and have a function on that interface that is "scope of". I just can't decide whether it's obvious enough how that looks quite yet, and how much work it is to change it :/

bsmithers: > provide the RefResolver class you want to use to this function

That was kinda where my initial thought was with this too. Unless I'm misunderstood, my concern is that a user may specify a `RefResolver` instance when calling `.validate()` -- this would mean a user is then specifying the draft number twice, i.e. `Draft4Validator` and `Draft4RefResovler`, which seemed unpleasant? I'm not up to speed enough to determine if that's a narrow enough issue though.

Julian: Right so I think the idea there would be to stash what is passed in here as the resolver callable onto the validator interface -- i.e., the newer way to construct resolvers would become `DraftSomethingValidator(resolver=DraftSomethingValidator.Resolver(...))`

bsmithers: Alright well I think I've got a bit of a better picture of the direction you'd like this to go in. I'll try and take a look at the related issues and then see if I can work towards something. Of course I shan't be offended if you're actively working on these things and prefer effort elsewhere.

Julian: Ideas I have, time, much less so :(, especially given that the ideas are often only partially baked and need refinement or experimentation to see whether they fully fit -- so definitely definitely would appreciate the help!

I think the one thing that seems most clear at the moment is the stuff I mentioned about `type`, so I might recommend that one as the first self-contained area to poke at -- widen the type interface, get that merged, and then come back here and wiggle this onto the new one.

But there are many many paths forward and yeah I super appreciate any help.

bsmithers: I've taken a look at the `type` stuff [1] - it's WIP and not yet ready for a PR, but if you have the time to give some feedback before I take it further it'd be really appreciated.

In particular:
- Rather than the single `is_type` injected function discussed in #365, I've taken the approach of a function per type. I figure this is more composable for extensions
- I presume a need to maintain backwards compatibility with the old `default_types` and per-instance `types` properties. I believe this does so (tests using this pass)  - but ideas appreciated
- I feel this works well for 'simple' types, for example handling the change in integer in draft6 and e.g.  a hypothetical scenario where strings encoding ints are valid under `integer` [2]. However, some difficulty remains for something like #364 to extend `object` to allow `namedtuple` - a user would have to provide custom validators for all `object`-based checks, e.g. in my test case [3]. Wether or not this is an acceptable level of work for a custom type or we should look to facilitate this I'm not sure. 

[1] https://github.com/Julian/jsonschema/compare/draft6...bsmithers:types
[2] https://github.com/bsmithers/jsonschema/blob/types/jsonschema/tests/test_types.py#L56
[3] https://github.com/bsmithers/jsonschema/blob/types/jsonschema/tests/test_types.py#L95

Julian: So! I think that looks like a great start -- if I can inject one idea there (which is only 80% baked I think, so lemme know if I go off a rail) -- I think the right thing here is to take a cue from what's happening on the ref resolver and format checker sides -- to introduce a `TypeChecker` class, one that you associate an instance of with a validator class in create and extend.

The one major mistake I'd love to correct would be to not have mutable APIs for that -- so, to talk about the specific `namedtuple` case, the API I'd expect would look roughly like:

```
def object_or_namedtuple(value):
    return Draft4Validator.type_checker.is_type("object") or getattr(value, "_replace", None) is not None
```

say, and then `Draft4Validator.type_checker.redefine(object=object_or_namedtuple)` which should return a new instance (which you'd use via `Draft4Validator(type_checker=that)` where the appropriate redefinition has happened (`attr.evolve` + a `pyrsistent.pmap` of checkers are the pieces we'd use to make that happen).

You can then also have `type_checker.redefine(**types)` to convert from the old format, and I think the above does seem like a reasonable amount of work to do the namedtuple case.

How does that sound to you (and as usual apologies for spitting information out slowly here after you've already gotten started :), as usual with these things what seems right only starts to come out and evolve after progress has begun).

manycoding:
@Julian Hi, could this be merged? It should help to resolve other pull requests, I reckon.

Julian:
@manycoding will have another look now but I'm not even sure anything in this PR is relevant anymore, all the tests now pass already and a lot of the specific fixes here went in separately.

What specifically from this one are you looking at/for?

Issue #341, opened 2017-06-22
Support Location-Independent Identifier $refs


hooke:
With jsonschema v.2.6.0, I couldn't manage to use id's of subschemas in $ref in the same file.

`test.json` with schema:
```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "title": "Test Schema",
  "description": "A test for subschemes' relative URIs",
  "type": "object",
"additionalProperties": false,
  "properties": {
    "prop1": { "$ref": "#smth"}
  },
  "definitions": {
    "smth": {
      "$id": "#smth",
      "type": "object",
      "properties": {
        "some_field": {
          "type": "string"
        }
      }
    }
  }
}
```

`test_sample.json` with JSON to test:
```
{
  "prop1": {
    "some_field": "a text"
  }
}
```
The python code:
```
import os
import jsonschema, json
from pprint import pprint

schema_absolute_path = "/home/hooke/prj/MK/python-test/example/"
schema_file_name = "test.json"

with open(schema_absolute_path + schema_file_name) as schema_file:    
    schema = json.load(schema_file)

try:
    jsonschema.Draft4Validator.check_schema(schema)
    print "Valid draft 4 schema"
except jsonschema.exceptions.SchemaError: 
    print "Invalid schema"

schema_absolute_uri = 'file://' + schema_absolute_path + schema_file_name

resolver = jsonschema.RefResolver(base_uri = schema_absolute_uri, referrer = schema)



validator = jsonschema.Draft4Validator(
        schema, resolver=resolver)

instance_file_name = "test_sample.json"
with open(schema_absolute_path + instance_file_name) as data_file:    
    data = json.load(data_file)

try:
    validator.validate(data, schema)
    print "Properly accepted object"
except jsonschema.exceptions.RefResolutionError as e: 
    print e
except jsonschema.exceptions.ValidationError: 
    print "Failed to accept object"
```

Result:
```
$ python example_test.py 
Valid draft 4 schema
Unresolvable JSON pointer: u'smth'
```
The aforementioned schema was validated successfully on jsonschemalint.com with draft-04, and with https://github.com/epoberezkin/ajv validator.

willson-chen:
@hooke Hi, I have tested #717 could solve this problem.

Julian:
This appears to work at this point (the code you pasted correctly validates the instance). Closing, but feel free to follow up if there are still any issues.

PR #371, opened 2017-11-17
Add support for referencing subschemas by id.

dfinlay:
Add support for `$ref` elements using a subschema's declared `id` value. This is supported by a an additional store in the `RefResolver` class that stores all `id` declared subschemas.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -215,3 +215,42 @@ def uniq(container):
                     return False
                 seen.append(e)
     return True
+
+
+def subschemas(schema, key=u'id'):
+    """
+        Finds all explicitly identified (sub)schemas contained in the provided ``schema``, including the root schema.
+        The optional ``key`` parameter is to allow for easier interop between draft-03/04 and draft-06+.
+
+        Arguments:
+
+            schema (dict):
+
+                The schema to extract all constituent (sub)schemas from.
+
+            key (str):
+
+                The keyword used to define a URI reference for a schema. id for draft-03/04 and $id for draft-06+.
+
+        Returns:
+
+            list: A list of the (sub)schemas defined within the provided ``schema``.
+
+        """
+    schemas = list()
+    if not isinstance(schema, dict):
+        return schemas
+    schema_id = schema.get(key, None)
+    if schema_id is not None and isinstance(schema_id, str_types):
+        schemas.append(schema)
+    for key, value in schema.items():
+        # Recurse if the value is a dictionary.
+        if isinstance(value, dict):
+            schemas.extend(subschemas(value))
+        # Check all list members. We don't have to worry about iterables due to raw schema type mapping.
+        if isinstance(value, list):
+            for item in value:
+                # Recurse if the value is a dictionary.
+                if isinstance(item, dict):
+                    schemas.extend(subschemas(item))
+    return schemas
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -897,7 +897,8 @@ class TestRefResolver(TestCase):
 
     def test_cache_remote_on(self):
         ref = "foo://bar"
-        foo_handler = mock.Mock()
+        foo_handler = mock.MagicMock(spec=dict())
+        foo_handler.__getitem__.side_effect = ''
         resolver = RefResolver(
             "", {}, cache_remote=True, handlers={"foo": foo_handler},
         )
@@ -909,7 +910,8 @@ class TestRefResolver(TestCase):
 
     def test_cache_remote_off(self):
         ref = "foo://bar"
-        foo_handler = mock.Mock()
+        foo_handler = mock.MagicMock(spec=dict())
+        foo_handler.__getitem__.side_effect = ''
         resolver = RefResolver(
             "", {}, cache_remote=False, handlers={"foo": foo_handler},
         )
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -299,6 +299,7 @@ class RefResolver(object):
         )
         self.store.update(store)
         self.store[base_uri] = referrer
+        self.subschema_store = self.extract_store_subschemas()
 
         self._urljoin_cache = urljoin_cache
         self._remote_cache = remote_cache
@@ -321,6 +322,14 @@ class RefResolver(object):
         """
 
         return cls(schema.get(u"id", u""), schema, *args, **kwargs)
+
+    def extract_store_subschemas(self):
+        subschemas = dict()
+        for base_id, schema in self.store.items():
+            subschemas[base_id] = {
+                subschema[u"id"].lstrip(u'#'): subschema for subschema in _utils.subschemas(schema)
+            }
+        return subschemas
 
     def push_scope(self, scope):
         self._scopes_stack.append(
@@ -407,6 +416,12 @@ class RefResolver(object):
 
         """
 
+        # First try looking up the fragment schema reference in the subschema_store.
+        try:
+            return self.subschema_store[document[u'id']][fragment]
+        except (KeyError, LookupError, TypeError):
+            pass
+
         fragment = fragment.lstrip(u"/")
         parts = unquote(fragment).split(u"/") if fragment else []
 
@@ -480,6 +495,9 @@ class RefResolver(object):
 
         if self.cache_remote:
             self.store[uri] = result
+            self.subschema_store[uri] = {
+                subschema[u"id"].lstrip(u'#'): subschema for subschema in _utils.subschemas(result)
+            }
         return result
 
 


Julian:
Hey @dfinlay thanks so much!

Would you mind elaborating a bit on what this is doing? Is this functionality you think is in the spec that's missing from jsonschema, or functionality you wish existed, or a convenience helper for functionality that exists already?

dfinlay:
Sorry, I should have added some more commentary. This pull request adds support accessing subschemas by their `id`/`$id` [keyword](https://tools.ietf.org/html/draft-zyp-json-schema-04#section-7.2). This specifically addresses #341 as well.

A note on the implementation, I added the `store_subschema` to `RefResolver`, because I didn't want to be too intrusive in the changes.  I _think_ it would be better to make `RefResolver.store` a `dict` of `dicts`, and use `#` to identify the root schema. Thoughts?

I'd like to add some more tests, but I was a bit lost in the the test structure. Guidance on where / how to add them would be appreciated.

Julian:
Aha, awesome!

I have to definitely read this carefully -- the one general thing I'd say though is that I definitely want to restrict public changes to the `$ref` implementation (to `RefResolver`), because I have strong suspicion that it needs replacing entirely ( #346 ). So definitely if this can be done in a way that doesn't even *add* to the public interface, we should do that I think (so, e.g., is there actually a need to add the additional `subschema_store` parameter, or can it just be created by not overridable externally?)

>  I think it would be better to make RefResolver.store a dict of dicts, and use # to identify the root schema.

This sounds likely to be backwards incompatible with what the current thing asks for. But I probably have to read the changes here carefully to understand exactly.

> I'd like to add some more tests, but I was a bit lost in the the test structure.

Yeah, sorry bout that :/. The tests for this seem very likely like they'd go in `test_validators`, alongside the 2 I see you modified, but I can follow up with a specific recommendation after I read the code.

Julian:
Hey, haven't forgotten this, just still having some trouble figuring out how much of this is a bug fix and how much is new :/.

It's likely my own fuzzy knowledge of `$ref` that I never properly sat down to think about.

Still definitely hoping to find some time to think about this PR though. Sorry for the wait.

elipapa:
any news on this?

dfinlay:
I should probably have been more proactive with this.  I needed to add more tests!  Also, looking back at this without rose-tinted glasses, there is as a problem with the implementation.  You cannot pass an object for validation against a defined subschema.  Instead, you can only validate subschema elements if they are children of root schema instance. This should be resolved before this change is merged.

Ideally, one should be able to pass a string id to the validator with an instance and have it validated against the schema w/ that ID, regardless of whether it is a subschema and root.

Here is an example of how you have to validate instances against subschemas now (Ignore the versions here, just an example).

```
schema = {
  "$schema": "http://json-schema.org/draft-06/schema#",
  "$id": "awesome.json",
  "title": "Awesome",
  "type": "object",
  "properties": {
    "name": {
      "title": "Awesome Name",
      "type": "string",
      "maxLength": 256,
      "default": ""
    },
    "sauce": {
      "$ref": "#awesome-sauce"
    }
  },
  "required": [
    "name",
    "sauce"
  ],
  "definitions": {
    "awesome-sauce": {
      "$id": "#awesome-sauce",
      "title": "Awesome Sauce",
      "description": "Sauce for your food that is quite awesome.",
      "enum": [
        "ranch",
        "french-onion",
        "au-jus",
        "misc",
        "recycled"
      ]
    }
  }
}

validator = jsonschema.Draft4Validator(schema)

data = dict(name="Dave's Homemade Ranch", sauce="ranch")

sauce = dict(sauce="ranch")

validator.validate(data)
> True

validator.validate(sauce)
> False

validator.validate(sauce, schema['definitions']['awesome-sauce'])
> True
```

DeepSpace2:
Hi guys. Any news on this? I have a schema that validates on an online validator but is invalid when validated with jsonschema

taichi-jp:
Any updates on this?
I really need to refer to other json files!

Julian:
So I thought I'd left this comment somewhere but apparently I never did, so apologies @dfinlay :/ (And of course again thanks so much for even sending the PR)

Two "approach"-y comments though for you or anyone else who's finding this and trying to take it the last mile:

First to fix the $ref issue I'd want to see at least one test in json-schema-org/JSON-Schema-Test-Suite being fixed by the patch. Right now there is one specific bug which leads to ~10 tests being skipped there across all drafts. If that's the bug this fixes, this should unskip those and show that they're passing.

And second, unless it is physically impossible because of the way the behavior needs to work, I don't want to change the current approach of ref resolution, whereby no pre-resolutin of references happens until they're encountered. That I'm pretty sure this approach changes, so yeah, if that's necessary due to the bug, I'd like to discuss that specifically (and understand why that's the case), but otherwise would want to see the scope fix happen lazily.

handrews:
@Julian I think at most you can only defer a resolution pass until the first plain-name fragment reference.  At that point, you need to walk potentially all of the schema as you cannot predict where the plain name was defined.  And ideally, if someone re-defined it, throw an error (although that's not required IIRC, the spec just says such behavior is undefined).

If you try to avoid walking the whole schema at that point, you either have to remember which parts you have not walked, or start over from the beginning when you find the next reference to a plain name fragment that has not already been resolved.

And actually, it's more complicated than this- if the reference resolves to something where the non-fragment part of the URI is not the same as the root schema's `$id`, then you have to go find the schema that declares that base URI.

I can see avoiding it up front as a useful optimization since many schemas never do either of these things, but once you need to calculate stuff, it seems easier to just do it all in one pass rather than try to minimize each lookup.

Julian:
@handrews thanks, probably eventually I'll understand how this needs to look :)

So, I guess at a basic level you're telling me that it's not necessarily the case that resolution at a given level involves only levels you've already traversed? That's kind of hard for me to picture, any chance you could give a simple example? When you say "plain name" you mean e.g. `$ref: "foo"`? Is the right mental model not that you then just apply the base URI that you've seen along the way to finding that schema?

handrews:
@Julian by plain name I mean the plain name fragment syntax rather than JSON Pointer fragments.  So:

```JSON
{
    "definitions": {
        "foo": {
            "$id": "#foo",
            "type": "string"
        }
    },
    "type": "object",
    "properties": {
        "fooByName": {"$ref": "#foo"},
        "fooByPointer": {"$ref": "#/definitions/foo"}
    }
}
```

Both `fooByName` and `fooByPointer` resolve to the same thing.  If you haven't look at `definitions` yet, you won't know that the URI fragment `"#foo"` appears there.  And you might have a more complicated situation like `definitions` nested in a schema under `definitions`.  Or someone could use `$id` to define a plain name fragment in a schema under `properties` or `items` or `allOf`.  I would not do that- I would put all re-usable schemas under `definitions`- but it is legal.

See https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.2.4 for a fairly comprehensive list of examples of how `$id` can be used to create various identifiers that could be referenced with `$ref`.

The way I handled it the one time I implemented it (sadly that code was left behind at a past job), I basically did a tree walk, keeping a stack of relevant base URIs (because as you see in the spec examples, multiple can be usable at the same time), and computed all possible full URI identifiers for each schema.  I kept a map of full URIs  to schemas so I never had to do this again once a schema was loaded- I think I internally replaced all `$ref`s that had relative URI references with fully resolved URIs, because otherwise you have to keep track of `$id` again while validating, and that's really annoying.

Julian:
Will have a look at the spec section you linked but man that seems really unfortunate to me :/

Would seem way cleaner to just allow one section where you define all your "variables" in, not litter every part of the document with a possible name. Maybe I'll come around as I see more examples on why that's a good idea :(

dfinlay:
Sorry I ignored following up on the additional tests for this PR, got pulled into other directions and it languished. If there is anything I can do to help close this out and get this part of the spec support I'm happy to. If the design approach I took is suboptimal, I'm not attached to it in anyway. I went for what I saw as the minimally invasive approach.

Julian:
@dfinlay deeeeeeefinitely not your fault, apologies for this taking so long with all the other things going on.

>  If there is anything I can do to help close this out and get this part of the spec support I'm happy to. If the design approach I took is suboptimal, I'm not attached to it in anyway. I went for what I saw as the minimally invasive approach.

Awesome!

From what @handrews said, I might be out of luck on what I want, so I might have to deal with this approach as a general strategy :/

I'm still triple checking whether there's no other way to make this happen, but if we do go this way, basically the main thing here would be to just make everything private (so no additional public API methods introduced, we do the schema traversal in some private method, store off all the mappings between simple names and schemas, and then reference them).

The two other minor things is that there's some small things to update here that now exist now that draft 6/7 support is merged (like for example the `key` argument you had to `subschemas` is now a thing that is passed in on `jsonschema.validators.create`, so you should be able to reuse that rather than taking the argument a second time).

And the last one that yeah I'd love to see the currently-skipped tests in JSON-Schema-Test-Suite be unskipped and confirmed to pass after this change.

If you make progress on any of those do lemme know, otherwise thanks again!

alvarogzp:
Hello, what is the status of this?

It has been more than one year since the PR was created and jsonschema still fails parsing a schema with `$ref`s pointing to custom `$id`s in the same schema (issue #341). It's sad that I'll need to modify my schema to be less expresive in order to verify it on Python (is there any other Python implementation with proper `$ref` support?).

Sorry for being hard, but it's really sad to see such a basic feature stalled for years, it makes me feel the project is dead! It is even referenced in the docs that jsonschema don't support this: https://json-schema.org/understanding-json-schema/structuring.html#using-id-with-ref.
Imagine how rare it is that this feature isn't supported to be mentioned on the official json schema guide!

Julian:
@alvarogzp please feel free to read the comments on this ticket -- notably, the need for an upstream test case to be failing and fixed by this (if the existing ones do so, great).

> It's sad that I'll need to modify my schema to be less expresive in order to use it on Python

Considering how composable jsonschema is, you absolutely do not, since you've got about 5 other options to choose from, not least of which being transforming your schema just before handing it to `jsonschema`, but please do save the guilt trip. It's a bit ridiculous to say such a thing, or create FUD around the project being dead which clearly can be seen not to be the case by just looking at the commit history.

jsonschema is an open source project.

alvarogzp:
Sorry about my previous post and language, it was totally inappropriate.
I'll try to be more proactive instead of just complaining.

I've been looking in the JSON-Schema-Test-Suite for any test fixed by this (as you suggested), and certainly there isn't. However, I found https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/225 which adds tests for it (the technical name is "Location-independent identifiers", and they are defined in https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.2.3). I'll try to help with that PR too.

Also, I have been testing this branch and found some issues regarding to the handling of the 'id' property (also mentioned on the comments above). I have fixed it, merged with master (there was a merge conflict) and added a small test to TestRefResolver (you can check the new test would fail against master and pass against this branch). The result is [here](https://github.com/alvarogzp/jsonschema/tree/ref-by-id), and the diff of my changes (without the merge commit) can be found [here](https://github.com/alvarogzp/jsonschema/compare/82ab075..e411906). Feel free to sync this branch with mine, or if you want I can create a new PR.

I tested the branch with the following schema and it validates properly (it doesn't on master):
```json
{
  "type": "object",
  "title": "Test",
  "properties": {
    "test": { "$ref": "#test" }
  },
  "definitions": {
    "test": {
      "$id": "#test",
      "title": "A test string",
      "type": "string"
    }
  }
}
```

If there is more I can do, just tell me.

Thanks for the project, and sorry again!

Julian:
Thanks, no hard feelings, and definitely thanks for the offer to help it is very much appreciated.

I think based on what you found upstream, yeah let's get that merged in there then first. I can review that upstream. Can you maybe have a look at that PR in the test suite, leave any comments there that you see, and then I can merge it (and possibly double check with @handrews to make sure he agrees, if he's available)?

Once that's done, we can come back here and have a look at the fix you put in as well, and maybe get this merged!

jjmurre:
Any change that this PR will be merged one day?

Julian:
Hey @jjmurre  -- do you think you'd be willing to help address the issues remaining above?

jjmurre:
I'll have a look, see if I can find my way around the codebase.

PR #374, opened 2017-11-27
Widened type interface

bsmithers:
Following on from discussions in #364 I think this is pretty much ready to be merged in, though of course let me know if there are any issues. I think the only bit that is really 'new' is the conversion from the old `default_types` & `types` to functions that are redefined in the `TypeChecker`

- Rebased the changes off of master removing any use of Draft6 (squashed into one commit)
- No changes to `TypeChecker` interface since latest discussions in the issue
- Expanded test coverage. Since the JSON schema test suite tests will now use the new `TypeChecker`, I've added tests that re-run the `types` tests explicitly on the old interface.
- Updated docs - hopefully is all clear.


Julian:
Woohoo! Awesome.

Will leave some comments but probably will be a trickle until I get a full review done.

Really appreciated!

docs/creating.rst
```diff
@@ -29,10 +29,16 @@ Creating or Extending Validator Classes
         will have :func:`validates` automatically called for the given
         version.
 
-    :argument dict default_types: a default mapping to use for instances
-        of the validator class when mapping between JSON types to Python
-        types. The default for this argument is probably fine. Instances
-        can still have their types customized on a per-instance basis.
+    :argument dict default_types: Deprecated. Please use the type_checker
```
Julian: Might be nice to use a [sphinx deprecated directive](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-deprecated) even though we haven't been very careful with `versionchanged` either...

docs/creating.rst
```diff
@@ -29,10 +29,16 @@ Creating or Extending Validator Classes
...
+        argument instead.
+
+        If set, it provides mappings of JSON types to Python types that will
+        be converted to functions and redefined in this object's TypeChecker
```
Julian: Period at the end please, and either "type checker" (i.e. not the class name) or ":class:`TypeChecker`" at the end.

docs/creating.rst
```diff
@@ -29,10 +29,16 @@ Creating or Extending Validator Classes
...
+
+    :argument jsonschema.TypeChecker type_checker: an instance
+        of :class:`TypeChecker`, whose :meth:`is_type` will be called to
```
Julian: It won't insert the word "method", so you have to say ":meth:`is_type` method"

docs/creating.rst
```diff
@@ -29,10 +29,16 @@ Creating or Extending Validator Classes
...
+        validate the :validator:`type` property If unprovided, a default
```
Julian: property. If

docs/creating.rst
```diff
@@ -29,10 +29,16 @@ Creating or Extending Validator Classes
...
+        :class:`TypeChecker` will be created, with no support types.
```
Julian: I think actually we can leave off the "if unprovided" entirely here and just rely on people looking at the newer property (which we've already pointed them to) for info on default behavior.

docs/creating.rst
```diff
@@ -59,6 +65,10 @@ Creating or Extending Validator Classes
 
     :argument str version: a version for the new validator class
 
+    :argument jsonschema.TypeChecker type_checker: an instance
+        of :class:`TypeChecker`. If unprovided, the existing
```
Julian: I try not to repeat instance-of-class kind of info in the actual prose (because it's already in the `:argument:` role). It's easy to get around with just "a type checker.", even though the class name is pretty similar in this caes.

docs/index.rst
```diff
@@ -47,6 +47,7 @@ Contents:
     references
     creating
     faq
+    types
```
Julian: Any reason to add this as an entirely new document? If not might just be reasonable to put this in the `validate.rst` doc, I think even format is already just in that same doc.

docs/types.rst
```diff
@@ -0,0 +1,23 @@
+.. currentmodule:: jsonschema
+
+=============
+Type Checking
+=============
+
+Each :class:`IValidator` has an associated :class:`TypeChecker`. The
```
Julian: So, half baked thought, but I guess in retrospect it's not even obvious that every validator *has* the `type` validator. In terms of JSON Schema it seems pretty likely that it will stay forever, but in the more generic notion, a bit less so... Maybe someone creates a validator that doesn't have that property (http://github.com/Julian/Seep is already such an example I think), in which case providing a type_checker to create is a bit random...

Definitely not saying to change anything, but feel the need to note that uncomfortableness somewhere.

docs/types.rst
```diff
@@ -0,0 +1,23 @@
...
+TypeChecker provides an immutable mapping between names of types and
```
Julian: Would recommend "type checker" again here.

docs/types.rst
```diff
@@ -0,0 +1,23 @@
...
+
+See :ref:`validating-types` for an example of providing a custom type check.
+
+.. autoclass:: TypeChecker
+    :members:
+
+.. autoexception:: jsonschema.exceptions.UndefinedTypeCheck
+
+    Raised when trying to remove a type check that is not known to this
+    TypeChecker. Internally this is also raised when calling
```
Julian: We don't really get a choice on the "internally" I think.

`TypeChecker.is_type` is public, so everyone needs to know it raises this exception there too, so we can never really consider it internal anymore.

docs/validate.rst
```diff
@@ -48,8 +50,10 @@ classes should adhere to.
 
     .. attribute:: DEFAULT_TYPES
 
-        The default mapping of JSON types to Python types used when validating
-        :validator:`type` properties in JSON schemas.
+        Deprecated. Under normal usage, this will be an empty dictionary.
```
Julian: We can't make this an empty dictionary, it has to preserve the existing behavior at least for `Draft3Validator` and `Draft4Validator`, and also for anyone who is using the `default_types` argument to `create`. If tests are missing for that we probably need to add them, it's existing behavior that someone could be relying on, so we can deprecate it but can't change it until we remove all those arguments.

We also need to have accessing the attribute warn with one of the warnings you've added (thanks!)

We can either use a property for that, or look at one of the libraries that are focused around "deprecation" itself if it starts to get annoying.

docs/validate.rst
```diff
@@ -152,24 +157,24 @@ more general instance checks can introduce significant slowdown, especially
...
+existing :class:`TypeChecker` or create a new one. You may then create a new
+:class:`IValidator` via :meth:`extend`.
 
 .. code-block:: python
 
     class MyInteger(object):
-        ...
+        pass
+
+    def is_my_int(instance):
```
Julian: Question which maybe was more obvious when writing this: maybe these functions should take both the instance and also the checker itself, rather than needing to "manually" pull that off?

(E.g., that'd allow for more easily reusing functions across drafts)

docs/validate.rst
```diff
@@ -152,24 +157,24 @@ more general instance checks can introduce significant slowdown, especially
...
+        return Draft3Validator.TYPE_CHECKER.is_type(instance, "number") or \
```
Julian: Parens instead of `\` please.

jsonschema/_types.py
```diff
@@ -0,0 +1,206 @@
...
+    :meth:`TypeChecker.redefine` or :meth:`TypeChecker.redefine_many` and
+    removed via :meth:`TypeChecker.remove` or
+    :meth:`TypeChecker.remove_many`. Each of these return a new
+    :class:`TypeChecker` object.
+
+    Arguments:
+
+        type_checkers (pyrsistent.pmap):
+
+            It is recommend to set type checkers through
```
Julian: Hm, is there a reason to recommend that over the param to `__init__`? If there is, I think it'd be better to remove the possibility entirely and stick to just that. (But if we can safely support both, seems fine to do so and to drop the recommendation).

jsonschema/_types.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    _type_checkers = attr.ib(default=pyrsistent.pmap({}))
+
+    def is_type(self, instance, type):
+        """
+        Check if the instance is of the appropriate type.
+
+        Arguments:
+
+            instance (any primitive type, i.e. str, number, bool):
```
Julian: Hm, it's not just primitive types is it? It's any JSON-deserialized thing?

jsonschema/_types.py
```diff
@@ -0,0 +1,206 @@
...
+
+        Raises:
+
+            :exc:`jsonschema.exceptions.UndefinedTypeCheck`:
+                if type is unknown to this object.
+        """
+        try:
+            return self._type_checkers[type](instance)
+        except KeyError:
+            raise UndefinedTypeCheck
```
Julian: Should include the name of the type here (and ideally add that to the test that checks this branch)

jsonschema/_types.py
```diff
@@ -0,0 +1,206 @@
...
+            bool: Whether it conformed.
+
+
+        Raises:
+
+            :exc:`jsonschema.exceptions.UndefinedTypeCheck`:
+                if type is unknown to this object.
+        """
+        try:
+            return self._type_checkers[type](instance)
```
Julian: Pedantry: in theory someone can add a type checker that itself raises `KeyError`, so a bit safer to do this in 2 steps and only catch the first one (and might be worth adding a test for that even though it's an edge case).

jsonschema/_types.py
```diff
@@ -0,0 +1,206 @@
...
+        except KeyError:
+            raise UndefinedTypeCheck
+
+    def redefine(self, type, fn):
+        """
+        Redefine the checker for type to the function fn.
```
Julian: ``fn``

jsonschema/_types.py
```diff
@@ -0,0 +1,206 @@
...
+
+                A dictionary mapping types to their checking functions.
+
+        Returns:
+
+            A new :class:`TypeChecker` instance.
+
+        """
+        definitions = dict(definitions)
+        evolver = self._type_checkers.evolver()
```
Julian: Isn't this just `self._type_checkers.update(definitions)`?

jsonschema/compat.py
```diff
@@ -28,7 +28,7 @@
     )
     from urllib import unquote  # noqa
     from urllib2 import urlopen  # noqa
-    str_types = basestring
+    str_types = basestring,
```
Julian: Shouldn't need this to be a tuple, do you?

jsonschema/exceptions.py
```diff
@@ -140,6 +140,9 @@ class RefResolutionError(Exception):
     pass
 
 
+class UndefinedTypeCheck(Exception):
+    pass
+
```
Julian: 2 blank lines please.

Julian:
Left a bunch of comments, most of which are nitpicks, still need to review the tests, but lemme know what you think.

docs/creating.rst
bsmithers: Hah - I nearly mentioned that in the PR. I'll double check, but the deprecated directive appears to require a version that something was deprecated in and I wasn't sure of your approach to versioning & releases.

docs/validate.rst
bsmithers: Good catch. This is partly a wording issue but I'm also going to revisit that in the code. Currently `default_types` is set to the old default if no `TypeChecker` is provided. However, you made me realise that this always needs to be true, because of direct access to the property.

jsonschema/_types.py
bsmithers: Yeah, this is kind-of an awkwardness with the attrs `frozen` class. In order for `evolve` to work, the attribute must be able to be set via `init`. This probably means exposing the underlying type (i.e. `pyrsistent.pmap`), which doesn't seem right.

I suppose that we could have init take in a standard dictionary instead, which is then converted to the pmap by the `TypeChecker` (and then a little juggling in the update methods). This doesn't feel massively clean either but is perhaps the preferred solution.

Thoughts? Sorry this didn't come up earlier, I overlooked this before as the init is generated by attrs.

bsmithers:
Thanks for reviewing, I'll look at this fully when I get some time but I've left a couple of replies.

jsonschema/_types.py
bsmithers: Yeah so this is the juggle to make it work:

```
    _type_checkers = attr.ib(default=())

    def __attrs_post_init__(self):
        object.__setattr__(self, "_type_checkers",
                           pyrsistent.pmap(dict(self._type_checkers)))
```
As it happens, this is the same as the attrs issue you open a few months back: https://github.com/python-attrs/attrs/issues/207

I think this is probably the right solution, however. It means that checkers can be defined on init without users needing to supply a `pmap`. Given this, we could then drop the `redefine_many` method (since the Draft3 `TypeChecker` initialisation was the main internal use of this) but I'm inclined to keep it around.

jsonschema/compat.py
bsmithers: Actually no, I didn't need this to change in the end  - at some point it was looking like I did. 

Isn't it odd that str_types is a tuple in PY3 and not the PY2 though?

Julian: I mean, mostly it's `isinstance` that's the odd thing and accepts either types or tuples of types, and since this is essentially only used for that, it's "fine" to mix those types.

Doesn't matter too much one way or the other though.

docs/validate.rst
bsmithers: > We also need to have accessing the attribute warn with one of the warnings you've added (thanks!)

This isn't quite a trivial as first sounds - `DEFAULT_TYPES` is a class property and access to it is very likely to be done so on the class rather than a concrete instance.

So we _could_ do one of the fudges for class properties, but is it worth it? Any access to `DEFAULT_TYPES` is likely (and arguably only a problem) to be followed by some derived use (i.e. `create(default_types=...)` or `validator(types=...)`), both of which would trigger a deprecation warning.

jsonschema/_types.py
Julian: What's making you reluctant to ask for a `pmap` and expect callers to provide one out of curiosity?

(If you did have such a reservation, you can use the `coerce` argument to `attr.ib`, but my personal position happens to be that in 2017, essentially make believe the stdlib types don't exist :)

bsmithers: It feels like an implementation detail. To me, the `TypeChecker` should provide an immutable interface but the way in which it does so shouldn't need to be public. Secondarily, it's simple, familiar and more concise for the caller if they can provide a vanilla dict. 

I'm not really too fussy, however.   

(Thanks for the pointer, that's a much cleaner way of doing it - looks like it's `convert` to `attr.ib`)

docs/creating.rst
Julian: Yeah -- I think if it helps that you can assume the next version will be 2.7.0 (but lemme know if there's anything else making that harder).

Julian:
Hey! Should I be looking at this again yet?

docs/validate.rst
bsmithers: I'm not sure I follow this one - seems it will end up quite circular as each of these functions is already attached to a checker. I also wonder if it would be a performance hit, since each type check would then also be doing a comparison on type checker object? Or maybe I misunderstood. 

For reusing across drafts - that's already handled through redefinition of types->functions, no? As in, most of the type checks remain the same and a new draft only redefines a subset.

bsmithers:
Hey, sorry for the slow down on this, I haven't quite had time to finish things off. Mostly just need to review the documentation changes. I'll try to get that done in the next day or so.

I've switched to using `convert` for the type_checker init as per your suggestion and fixed the the deprecation of `DEFAULT_TYPES` - the attribute will always be available but any access will trigger a deprecation warning (added test coverage for this also).

Julian:
Trusssst me I get the not having enough time part :) was just checking if you meant to resubmit it yet, but no rush at all.

docs/validate.rst
Julian: They're attached to a checker, but they don't know what the "current" checker knows.

E.g., imagine someone wants to write a checker that satisfies "number or string" -- they can't do that under the "current" definition of number of a type checker. I.e., compare:

```
def number_or_string(instance):
    return isinstance(instance, int) or isinstance(instance, str)
```

vs.

```
def number_or_string(checker, instance):
    return checker.is_type(instance, "number") or checker.is_type(instance, "string")
```

(This is a toy example, but hopefully the point is clear enough now at least? If not lemme know)

bsmithers:
Think this is ready for a second look. 

One thing I'm unsure of it's what's happening with the `bool` checks and perf - I've seen there's been a few changes but I'm not exactly sure what's been going on or how the performance tests are setup (I think it's ad-hoc at the minute?)

docs/validate.rst
bsmithers: Ah yeah I get it now. I think that's a worthwhile change. It's particularly useful for custom types as when you're overwriting the checking function for an existing type you'll no longer be able to access the original via `checker`.

docs/types.rst
bsmithers: Interesting use case. I agree that it seems pretty core to JSON schema. Nonetheless I've reworded the docs slightly. It is slightly clunky but one would simply not provide a `TypeChecker` in such a case.

Julian:
Yeah unfortunately the perf stuff is still ad-hoc because I can't find a place that I can push them to yet.

jsonschema/validators.py
```diff
@@ -56,24 +61,91 @@ def _validates(cls):
...
+        warn("DEFAULT_TYPES is deprecated, use type_checker",
+             DeprecationWarning)
+        return self._DEFAULT_TYPES
+
+
+def create(meta_schema, validators=(), version=None, default_types=None,
+           type_checker=None):
+
+    use_default_types=False
+    if default_types is not None or type_checker is None:
```
Julian: This looks like it will mis-warn in the case that someone doesn't provide neither default types nor type checker -- that doesn't seem like it's a deprecated case, is it?

jsonschema/validators.py
```diff
@@ -56,24 +61,91 @@ def _validates(cls):
...
+        TYPE_CHECKER = type_checker
 
         def __init__(
-            self, schema, types=(), resolver=None, format_checker=None,
-        ):
-            self._types = dict(self.DEFAULT_TYPES)
-            self._types.update(types)
+            self, schema, types=(), resolver=None, format_checker=None):
+
+            if types:
```
Julian: This won't warn in the (unlikely but possible) case that someone provides a falsy value to types, even though that code will break when the argument is removed.

Julian:
Just accidentally merged this in the process of trying to get Travis to behave :D -- still have a few things that need resolving but this is super close -- going to leave the comments here as I find them during review but totally alright with knocking off these myself as well if need be.

Julian:
And thanks again! Now back to getting the rest of the stack done :)

docs/validate.rst
```diff
@@ -32,10 +32,14 @@ classes should adhere to.
         will validate with. It is assumed to be valid, and providing
         an invalid schema can lead to undefined behavior. See
         :meth:`IValidator.check_schema` to validate a schema first.
-    :argument types: Override or extend the list of known types when
+    :argument types:
+        .. deprecated:: 2.7.0
+           Instead, create a custom type checker and extend the validator. See
```
Julian: I wonder whether it's worth deprecating format_checker at the same time (and telling people to use extend).

docs/validate.rst
```diff
@@ -152,24 +183,24 @@ more general instance checks can introduce significant slowdown, especially
...
+:class:`IValidator` via :meth:`extend`.
 
 .. code-block:: python
 
     class MyInteger(object):
-        ...
+        pass
+
+    def is_my_int(checker, instance):
+        return (Draft3Validator.TYPE_CHECKER.is_type(instance, "number") or
```
Julian: Ech, so I didn't know whether this was intentional or not, but now I see that there's some infinite recursion nonsense here when you upcall...

jsonschema/validators.py
bsmithers: I believe so? If neither are provided, we must use the old default behaviour, which relies on default_types? Those that directly use `DraftXValidator` will not see this as these provide a type checker.

jsonschema/validators.py
bsmithers: Good catch. It should check for anything other than an empty tuple.

docs/validate.rst
bsmithers: Is there? In order to use the `is_my_int`, a user must generate a new `TypeChecker` instance (via, e.g. `redefine`), which is then given to a new validator (via `create` or `extend`). In other words:

```
Draft3Validator.TYPE_CHECKER.is_type(instance, "number")
```

Is not changed.

jsonschema/validators.py
Julian: We have to preserve the old *behavior*, but presumably in a new way -- the old behavior was "if you provide neither thing, you get a validator with type checking via some 'sane' defaults" -- the new way to do that would be to construct a suitably equivalent `TypeChecker`.

But I actually see that the code here is making the "new" default behavior be an empty type checker (with no types registered), but I can't see when that would actually be encountered, which is what's confusing me.

docs/validate.rst
Julian: (This was a half-baked comment that I just wrote to make sure I didn't forget to address it, sorry for the missing context -- it's the following:)

Not in the way you wrote it, but in the way I *thought* you mean to write it, which was:

`return checker.is_type(instance, "number") or isinstance(instance, MyInteger)`

That will do nasty infinite things because it calls itself, but I think the expectation there is clearly to upcall to the original definition of number.

And the above case was basically my motivation for suggesting we include the `checker` as one of the two arguments to the function (because it allows for forward extension).

The infinite recursion case though makes that a bit more complicated, it means there's a different thing you do if you're overriding the same type and want to upcall as if you're trying to access *another* type checker.

I think this needs a bit more thought -- my first reaction when I realized the above was "maybe we should pass in a third argument too, the original function from the checker that was redefined", but I think we just need to think through the 2 use cases again and make sure this API makes sense (the use case where you're accessing another checker, and the one where you're trying to upcall to the original one).

jsonschema/validators.py
bsmithers: > We have to preserve the old behavior, but presumably in a new way -- the old behavior was "if you provide neither thing, you get a validator with type checking via some 'sane' defaults" -- the new way to do that would be to construct a suitably equivalent TypeChecker.

Agreed. 

> But I actually see that the code here is making the "new" default behavior be an empty type checker (with no types registered), but I can't see when that would actually be encountered, which is what's confusing me.

I think maybe the confusion surrounds default _behaviour_ and _default_ values. The type checking behaviour always depends on the value of _both_ `default_types` and `type_checker`. 

The default value for `type_checker` is indeed an empty type checker. The default value for `default_types` is as it used to be, i.e. a sane type mapping. So to maintain the old behaviour - when neither are given, we redefine the empty type checker using the default type mappings. 

Does that make sense? We cannot make the default for `type_checker` anything other than  empty because we need to correctly process non-default-value `default_types`.

docs/validate.rst
bsmithers: > And the above case was basically my motivation for suggesting we include the checker as one of the two arguments to the function (because it allows for forward extension)

Apologies as I think we could have had this conversation earlier. I was concerned about attempted infinite recursion too but liked the ability to reference another type via `checker`.

The third argument doesn't immediately feel right to me - mostly because of the awkward basecase of a null function (in fact its not a null function, it's one that would raise an exception for an unknown type). 

Will have a think about it.

jsonschema/validators.py
Julian: Is a user who does the following doing something wrong:

```
⊙  ~[jsonschema:python] -c 'import jsonschema; print jsonschema.validators.create({})'                                                                                                                                                                                                                                                                      Julian@Macnetic
jsonschema/validators.py:173: DeprecationWarning: default_types is deprecated, use type_checker
  DeprecationWarning,
<class 'jsonschema.validators.Validator'>
```

they didn't provide `default_types` at all, why do they see a warning about it?

bsmithers: My thought process was that they are using it, just implicitly. However, I think you're right and this should not raise a warning. Any future removal of `default_types` should then preserve the same sane type checking defaults through the default value of `type_checker`.

docs/validate.rst
bsmithers: Sorry for going AWOL on this - turns out changing jobs and moving takes up a fair bit of time. 

As far as I remember, this point was the main remaining issue. I think I am in favour of the current API as a balance of simplicity and power. To me, the biggest benefit for passing in the type checker object (i.e. the current API) is for the ease of custom types, e.g. "string_or_int". In addition, it feels OK to be pretty explicit about _which_ type checking function you are upcalling to in the case of a type redefinition. 

Passing in _both_ the object and function that is being overwritten feels like we are providing a somewhat bizarre OO inheritance mechanism, which I dislike.

What do you think? Perhaps we simply add a warning to the documentation on the potential for infinite recursion? If you're happy with that I'll sort a PR for this and the other minor comments.

Julian: Definitely no need to apologise, trust me staying at the *same* job and *not* moving is also pretty killer :(

I think the warning might be the thing, but I wanna reread my own comment again and see if a months time has given me any subconscious enlightenment (I doubt it...)

PR #378, opened 2017-11-30
perf: Make is_type faster and cleaner

alex-hofsteede:
Instead of flattening to find out `is_number`, we can just check `issubclass(int, pytypes)` which doesn't require flattening the types list. Only if that passes do we need to look further.


alex-hofsteede:
This is similar to #232 which looks to be stuck in review for quite a while. I don't really have the time to rigorously benchmark it, but it is around 10% faster on the workloads I'm using it for. Perhaps it can be accepted as _prima facie_ better than the current master?

Julian:
@alex-hofsteede I think that thanks to a recent contributor #232 is actually benchmarkable regardless -- will have a look both at this patch and the one there.

PR #388, opened 2018-01-25
Move requests import to runtime

hmehta:
Importing requests takes well over 100ms and is used only for resolving refs from remote URIs - move the import to runtime.

--

Hopefully the test case modification is ok too - mocking a runtime import was not something I've done before :) If there is more elegant or "industry standard" solution, please let me know, and I'll fix it.

### DIFF
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1013,6 +1013,26 @@ class TestValidate(TestCase):
         )
 
 
+class MockImport(object):
+
+    def __init__(self, module, _mock):
+        self._module = module
+        self._mock = _mock
+        self._orig_import = None
+
+    def __enter__(self):
+        self._orig_import = sys.modules.get(self._module, None)
+        sys.modules[self._module] = self._mock
+        return self._mock
+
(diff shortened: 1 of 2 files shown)


Julian:
Thanks! I'm on a flight today so might not get a chance to read this carefully until the weekend.

I also think we'd want to have a benchmark (in the `jsonschema.benchmarks` package) for the actual `import jsonschema` timing if that's behavior you want to rely on, otherwise it's possible it'll regress.

Really appreciated!

apiraino:
hi @hmehta I'm quite interested in this PR. If I understand, moving the `requests` import inside `resolve_remote()` would shave ~100ms off each time a schema validation is run, correct?

I'm asking because we are using this library pretty heavily in request/response cycles; any small improvement would dramatically impact our general backend performances.

Is some perf unit testing is needed, can this PR move forward? It would great to have it merged :pray:

Julian:
It's not per validation, this PR is for the one time cost that happens
during import (but still do want to move it along)

Julian:
Merged finally, apologies for the delay... really appreciated!

Issue #391, opened 2018-02-09
Add support for optional strict validation of types


adi69:
One should have the freedom to check for types inside strings.
For instance, given the following schema
```
    {
        "schema": {
            "id": "schema_id",
            "properties": {
                "age": {
                    "type": "integer",
                    "description": "",
                }
            }
        }
    }
```
above schema should validate true for following data when strict validation is False
```
{ "age": "23" }
```

Why? It'll be helpful in API validation of query in GET requests and headers.

Julian:
This sounds like the opposite of "strict" :)

It's possible already though to do now that we've got [fully extendable type checkers](http://python-jsonschema.readthedocs.io/en/latest/validate/#type-checking). Haven't tidied up enough to do a release just yet, but if you give HEAD a try you'd be able to do this.

Feedback welcome.

PR #392, opened 2018-02-09
Add support for optional strict type validation

adi69:
https://github.com/Julian/jsonschema/issues/391


Julian:
As I mentioned on the issue, I think this isn't needed anymore, you'd be able to do it with the new API. Please have a look there and let me know what you think.

PR #401, opened 2018-04-25
Fix: Convert properties to string before regex

gcallaghan:
It is not uncommon for developers to create schemas in YAML, for example
when developing an Open API Specification. When keys are numeric and
parsed with the yaml parser they are typed and throw an error when
attempting to validate patternProperties or similar pattern matching
validation. This changeset converts the property to a string before
attempting to match the regex.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -101,7 +101,10 @@ def find_additional_properties(instance, schema):
     patterns = "|".join(schema.get("patternProperties", {}))
     for property in instance:
         if property not in properties:
-            if patterns and re.search(patterns, property):
+            prop = property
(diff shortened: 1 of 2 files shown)


Julian:
Hey, thanks.

This is a dupe of #286 / #285.

Have a look and let me know what you think.

Julian:
(for which by the way my opinion at the minute is still that this should be up to the deserializer totake care of immediately after deserializing)

gcallaghan:
Sounds fair. I'll move down stream to open API validator that lead me here in the first place and probably round trip yaml to JSON.

The Type Error was cryptic and annoying, but again more of a down stream issue.

Julian:
Great!

Issue #403, opened 2018-05-08
Email format not properly validated


staab:
```
from jsonschema import Draft4Validator, FormatChecker

validator = Draft4Validator({'type': 'string', 'format': 'email'}, format_checker=FormatChecker())
validator.validate('test@example.com ')
```

The above code doesn't raise an error, but should, at least as far as I understand [RFC 5322](https://tools.ietf.org/html/rfc5322#section-3.4.1).

----------------

Upon further investigation, your [email checking function](https://github.com/Julian/jsonschema/blob/118e6a2412afcb64544c607425fec2ef346880bf/jsonschema/_format.py#L151) is _terrible_. I've been trusting this library to handle the hard parts of validation for me (particularly emails!), but even a _regex_ would have been a better way to validate emails. Just for instance, this passes validation:

```
validator.validate('#@%^%#$@#$@#')
```

I'd really like to see the email validator pass [this suite of tests](https://blogs.msdn.microsoft.com/testing123/2009/02/06/email-address-test-cases/). 

It's fair that you haven't implemented the validation, since the jsonschema spec doesn't require format validation, but partway implementing it, and not warning users (either through raising errors on schemas relying on incompletely-implemented formats, or through documentation), that your format checker doesn't actually validate against the RFC, is pretty bad. My trust in this library has been shaken, and I'll be auditing my code to add custom validation logic where necessary, or switching away entirely.

Julian:
The email validator is intentionally unimplemented, because validating email addresses isn't really a reasonable concept. It's the one and only way that `jsonschema` intentionally deviates from the specification, its done to prevent downstream users from footgunning themselves with a concept that I wish didn't exist in the world, and it was my intention to voice upstream that I think we should remove the email validator. I could have sworn that was noted in the docs, but it's certainly noted in previous issue tickets about this. Patches are certainly welcome.

That being said, your attitude and commentary isn't welcome. Use the library if it helps you, don't if it doesn't, I don't really care about your "trust" in the library.

staab:
After looking through the [validation](https://python-jsonschema.readthedocs.io/en/latest/validate/) section of the docs again and not seeing anything calling out email validation in particular, I went ahead and created #404.

It still seems worth it to me to remove or warn about the email validator, as well as validators that require dependencies or formats that aren't implemented, even though the library is following jsonschema's specification to let that stuff pass. You'd at least get fewer people creating duplicate issues like this one.

Julian:
[...]

PR #404, opened 2018-05-08
Clarifying limitation of email validator in docs

staab:
Related to issue #403

### DIFF
--- a/docs/validate.rst
+++ b/docs/validate.rst
@@ -278,6 +278,12 @@ validation can be enabled by hooking in a format-checking object into an
     Traceback (most recent call last):
         ...
     ValidationError: "-12" is not a "hostname"
+    
+.. note::
+    
+    For the sake of speed and simplicity, format checkers may implement only 
+    partial validation; in particular, full rfc5322_ validation for email 
+    addresses is not provided. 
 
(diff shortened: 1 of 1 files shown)


Julian:
Thanks! Glad to add something like this -- I have to elaborate a bit more in a bit (I believe you asked on the actual issue ticket), but the reasoning here doesn't have to do with simplicity or speed, and doesn't apply to anything else other than email, so I wouldn't want to imply that anything else is potentially compromised.

The reasoning here is essentially https://hackernoon.com/the-100-correct-way-to-validate-email-addresses-7c4818f24643 -- validating email addresses via regex just isn't a reasonable thing to do without sending emails to them, not on the server side -- on the frontend, to just show a warning and not fail, maybe, but not where `jsonschema` is likely to be used.

staab:
Thanks for the link, that helps make sense of this for me a bit more. I'll go ahead and re-word the PR (and fix it so it builds).

Also, I want to apologize for my abrasiveness on the issue; it really wasn't my intention to be rude or ruin your Tuesday. It's something I'm working on 😅

Julian:
This is great, thanks!

And no hard feelings! All good, communication is hard for us all :)

PR #415, opened 2018-05-29
Update draft6 meta-schema from json-schema.org

robherring:
Update the local copy of the draft6 meta-schema to match current
json-schema.org version.

### DIFF
--- a/jsonschema/schemas/draft6.json
+++ b/jsonschema/schemas/draft6.json
@@ -57,6 +57,10 @@
             "type": "string"
         },
         "default": {},
+        "examples": {
+            "type": "array",
+            "items": {}
+        },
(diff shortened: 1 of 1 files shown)


Julian:
Hm, interesting, I only pulled that recently, didn't know Draft 6 was getting changes pushed to it.

Thanks though!

handrews:
> Hm, interesting, I only pulled that recently, didn't know Draft 6 was getting changes pushed to it.

Only bugfixes, as in "as originally published, it would fail validation on some valid schemas, and/or misleadingly allow some invalid ones".  I forget what the bug was.  We've had to fix draft-04 a few times as well.

Julian:
Ah, of course, makes sense.

PR #416, opened 2018-05-29
Attach additional properties error to each bad key

robherring:
Instead of a single error for unexpected additional properties attached
to the container dict()'s path, create an error for each unexpected key
and include the key in the path. This is useful when parsing errors to
identify the exact failing object, which is helpful when identifying
the location of the failing key.

In the particular case of using the ruamel.yaml parser, line & column
numbers are preserved in the data structure, and having the full path to
the failing key means error messages can include the line & column
numbers.

Signed-off-by: Grant Likely <grant.likely@arm.com>

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -55,8 +55,9 @@ def additionalProperties(validator, aP, instance, schema):
             )
             yield ValidationError(error)
         else:
-            error = "Additional properties are not allowed (%s %s unexpected)"
-            yield ValidationError(error % _utils.extras_msg(extras))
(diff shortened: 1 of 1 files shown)


Julian:
Hi!

Thanks!

This needs tests :)

Also, I don't think path is the right place to put this for the reasons I mentioned in #119.

I suspect we need a new attribute, or to go with one of the other solutions proposed at the bottom of that issue, but if one specific direction to go in is unclear there (it's been awhile since I've looked at that ticket) I can review it and suggest one.

PR #417, opened 2018-05-29
Handle $id property for draft6 in validates()

robherring:
Fix another spot where 'id' vs. '$id' needs to be handled for draft 6+.

Signed-off-by: Grant Likely <grant.likely@arm.com>
[robh: re-word commit message]

### DIFF
--- a/docs/spelling-wordlist.txt
+++ b/docs/spelling-wordlist.txt
@@ -15,6 +15,7 @@ indices
 ipv
 iterable
 jsonschema
+pre
 programmatically
 recurses
 regex
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -76,6 +76,28 @@ class TestCreateAndExtend(TestCase):
             validators.create(meta_schema={u"id": "id"})
         self.assertFalse(validates.called)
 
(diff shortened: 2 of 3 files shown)


Julian:
Here though I think this (unfortunately) is the right solution, this too needs tests!

(And thanks again!)

Julian:
And likely a quick note in the docstring.

robherring:
Something like this?

    def test_custom_validator_draft6(self):
        Validator = validators.create(
            meta_schema={"$id": "meta schema id"},
            version="12",
        )
        schema = {"$schema": "meta schema id"}
        self.assertIs(
            validators.validator_for(schema),
            Validator,
        )

Julian:
Yep, exactly!

I'd add one more that uses `validates` directly since that too is public API and is being changed here (with a similar assertion to what you have there, or just one that asserts on `meta_schemas` since unfortunately that's public API)

And ideally a third that tests the behavior for when `id` and `$id` are both present -- the code you have here (intentionally or otherwise) prioritizes one over the other, but I'm inclined to think that we should raise an exception and refuse to guess there.

robherring:
@Julian Any issues with this change?

Julian:
Didn't notice you'd updated it, thanks for the ping! Will have to re-review.

PR #419, opened 2018-06-04
RefResolver export function for resolving references

kave:
I had a use-case to take a schema with references and create a ref-resovled version. 

I figured this maybe useful for other users to have the ability to export a reference resolved version of a schema.

## Additions
- RefResolver export function
- unittests to cover use case

## Testcase Example
Before
```
{
  "$id": "tree",
  "type": "object",
  "$schema": "http://json-schema.org/draft-06/schema#",
  "properties": {
    "name": {
      "type": "string"
    },
    "fruits": {
      "type": "object",
      "properties": {
        "apples": {
          "$ref": "apples"
        },
        "oranges": {
          "$ref": "oranges"
        }
      }
    }
  }
}
```
after
```
{
  "$id": "tree",
  "type": "object",
  "$schema": "http://json-schema.org/draft-06/schema#",
  "properties": {
    "name": {
      "type": "string"
    },
    "fruits": {
      "type": "object",
      "properties": {
        "apples": {
          "$id": "apples",
          "type": "object",
          "$schema": "http://json-schema.org/draft-06/schema#",
          "properties": {
            "name": {
              "type": "string"
            },
            "quantity": {
              "type": "integer"
            },
            "types": {
              "type": "array",
              "items": {
                "type": "string"
              }
            }
          }
        },
        "oranges": {
          "$id": "oranges",
          "type": "object",
          "$schema": "http://json-schema.org/draft-06/schema#",
          "properties": {
            "name": {
              "type": "string"
            },
            "quantity": {
              "type": "integer"
            },
            "types": {
              "type": "array",
              "items": {
                "type": "string"
              }
            }
          }
        }
      }
    }
  }
}
```


kave:
@Julian hey not entirely sure why some of these Draft6 tests are failing only added a new function and test. the test i wrote for this passes so was curious if you could help me debug?

Julian:
Hey!
Going to likely be due to the combination of the draft 6 tests hitting the
network at the minute (because of an unfinished part of its implementation)
+ the json-schema.org site being down at the minute.

Would ignore it, I can have a look!

(Haven't reviewed this yet)

maddanio:
How would this behave, or expected to behave, with recursive references? I guess it can only give up, right?

gazpachoking:
> How would this behave, or expected to behave, with recursive references? I guess it can only give up, right?

I'm not fully sure the consequences, but python is fine with recursive data structures, so I don't think that should be a problem.

maddanio:
Hmm, true, guess one would have to be quite careful though not to try and perform any kind of tree-like traversal though :)

Julian:
So, sorry for letting this sit around for so long, but I think this should really live externally to jsonschema as a small library on top.

(It even might want to have a CLI that takes a file with refs and resolves them).

But yeah as-is, other than that there's the known bug that jsonschema has with location-independent references, I don't think there's a good enough reason to require that this lives in the library itself. 

Happy to hear further thoughts, but for now going to close. Of course much appreciated, and apologies again that I couldn't make my mind up sooner.

PR #422, opened 2018-06-11
add parameter checking to checks()

martingkelly:
Currently, we don't check that the arguments to checks() are actually a
string and a function (in that order). It's easy to reverse the order of
arguments, which leads to silent failure because the function is not a
match when we do a key lookup in self.checkers.

Fix this by validating the paramaters to checks().

### DIFF
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -60,6 +60,12 @@ class FormatChecker(object):
         """
 
         def _checks(func):
+            if not isinstance(format, str_types):
+                raise FormatError('specified format %s is not a string'
+                                  % format)
+            if not callable(func):
+                raise FormatError('specified function %s is not callable'
+                                  % func)
             self.checkers[format] = (func, raises)
             return func
         return _checks
--- a/jsonschema/tests/test_format.py
+++ b/jsonschema/tests/test_format.py
@@ -62,3 +62,8 @@ class TestFormatChecker(TestCase):
             validator.validate("bar")
 
         self.assertIs(cm.exception.__cause__, cause)
+
+    def test_bad_args(self):
+        checker = FormatChecker()
+        func = checker.checks(self.fn, raises=ValueError)
+        self.assertRaises(FormatError, func, "foo")


Julian:
Not a real fan of type checking, this is Python after all :)

Have you considered using keyword arguments? They'd have prevented your bug here as well

martingkelly:
Although I agree with following Python philosophy in general, I would argue that stopping bugs should come first, as the aim of Python philosophy in the end is to write better code with fewer bugs. I would add that there are many `isinstance` calls in `_format.py` already, so if there is a stylistic convention here, it's already been violated. If you can suggest another way to prevent this bug that you prefer, I am all ears.

The issues I see with keywords are as follows:
- If we change the function signature, we break compatibility with existing use, which is bad.
- If instead we add an optional keyword argument, then it duplicates the existing `format` and `func` arguments. So now it's ambiguous what happens if someone calls `checks("a", format="b")`. We could check for this, but now we've added a way to "use the API wrong", which is bad.

martingkelly:
One other to add wrt "type checking is bad": The spec requires that the format be a string. If we don't actually check that, we're violating the spec.

http://json-schema.org/latest/json-schema-validation.html#rfc.section.7
`The value of this keyword is called a format attribute. It MUST be a string.`

handrews:
> One other to add wrt "type checking is bad": The spec requires that the format be a string. If someone tries to add a format that is not a string, they're doing something that can never work for a valid schema (it's just a bug).

But you can already check this with a meta-schema.  This is what drives me nuts about people wanting strong type checking in loosely typed languages when working with JSON Schema (not you specifically, there's a completely separate and much more irritating personal backstory here :-P ).  JSON Schema already handles it.

martingkelly:
> But you can already check this with a meta-schema. This is what drives me nuts about people wanting strong type checking in loosely typed languages when working with JSON Schema (not you specifically, there's a completely separate and much more irritating personal backstory here :-P ). JSON Schema already handles it.

In this case, the bug in the interaction between my calling code and the `jsonschema` implementation code; the schema cannot help us here, as it's entirely an implementation issue and occurs before the schema even gets looked at.

Julian:
@martingkelly not sure my comment was understood, so I'll elaborate a bit -- you should use `@checks(format="foo")`, not `@checks("foo")`, in which case it is very very hard to think you should pass a function in there rather than a string. This is already a thing, all arguments in Python can be passed in both ways.

More generally, IMHO, you should use keyword arguments (named arguments) everywhere in Python (and otherwise), unless you can justify why passing something in positionally is acceptable.

And FWIW:

> Although I agree with following Python philosophy in general, I would argue that stopping bugs should come first, as the aim of Python philosophy in the end is to write better code with fewer bugs. I would add that there are many isinstance calls in _format.py already, so if there is a stylistic convention here, it's already been violated.

No `isinstance` calls appear in this module or any other other than where required by the JSON Schema specification.

All the Python APIs that `jsonschema` provides do not type check, consistent with the above.

Let me know if you understood the above though and still disagree.

handrews:
> as format must be a string, so asking to register a format that is not a string doesn't make any sense and should be treated as an error.

That seems pretty reasonable.  The specific string is unconstrained by the spec, but the fact that the value needs to be a string is part of the spec.  I'm still not sure why meta-schemas are insufficient here, though.

martingkelly:
> I'm still not sure why meta-schemas are insufficient here, though.

The bug is in code library code and the code calling it, not in the schema data (the schema data itself is just fine). How can a meta-schema help you validate Python code?

Julian:
> I think this is essentially required by the spec, as format must be a string, so asking to register a format that is not a string doesn't make any sense and should be treated as an error.

You can make this argument for every function in this library :). The spec requires you to use your implementation correctly.

The distinction is between places where the spec says some behavior is required, and where the Python library has introduced a concrete Python API to do something. In the former, `isinstance` checks are "necessary" unfortunately.

In the latter, they are not.

> which would have broken. Though it does not stop me from writing:

> format_checker.checks(name=func)(func=name)

It does stop you from writing that, because passing keyword arguments that don't exist is a `TypeError` same as passing extra (or too few) positional ones.

> I can certainly see the argument for this, though I'm not super happy with it as a solution, as now we have a "safe" and "unsafe" way to call a function, which is confusing and contradicts the "only one way to do it" tenet of Python.

You'll find Python itself breaks that rule quite a few times :). Here you see one example manifested in the fact that there are two ways to pass any given argument in the first place.

handrews:
@Julian thanks, that makes sense.

martingkelly:
> You can make this argument for every function in this library :). The spec requires you to use your implementation correctly.

Yes, you can, and I actually would make this argument for every function. I think we have a philosophical difference, in that I believe that if you can stop an invalid use-case (one that can never result in correct code) from occurring, then you should, even if it violates idiomatic Python. It sounds like you would prefer to follow idiomatic Python, even if it allows the invalid use-case to silently succeed.

> It does stop you from writing that, because passing keyword arguments that don't exist is a TypeError same as passing extra (or too few) positional ones.

In this case, I've swapped the order of the parameters, but the keywords are still correct (so "name" is pointing to a function instead of a string, and "func" is pointing to a string instead of a function). This would not result in an exception, but would instead silently succeed.

> You'll find Python itself breaks that rule quite a few times :). Here you see one example manifested in the fact that there are two ways to pass any given argument in the first place.

I know Python breaks that rule, and many other of its own rules in various places. To me, that's more argument for writing code to be correct as possible, making exceptions to idioms when it would improve code by eliminating bugs and "ways to use the API wrong". I think we differ on that.

Julian:
We're going to agree to disagree here yes.

This is a pretty common thing in Python (and I personally consider libraries that do what you're suggesting to be buggy). There are good reasons *not* to do this in Python, and very few reasons to do it that aren't already solved by simply writing unit tests for your code (which would have caught this too).

Regardless, the contribution is truly appreciated even if I'm going to decline it.

PR #425, opened 2018-06-19
Return regex pattern error if regex failed to compile

manycoding:
If an invalid regex pattern were provided, validation fails.
So I added validate_regex to check it and also a yield "Pattern error in" .


manycoding:
@Julian The tests are ok after I rebased and merged the last master.
Would you please review it?

Julian:
Hi -- this wouldn't be the right place to do that check, it'd need to happen during schema checking, but that should already be the case, e.g.:

```
⊙  ~[jsonschema:python] -c 'import jsonschema; jsonschema.validate(u"foo", {u"pattern": object()})'                                                              Julian@Magbook ●
  File "<module>", line 1, in <module>
  File "jsonschema/validators.py", line 823, in validate
    cls.check_schema(schema)
  File "jsonschema/validators.py", line 244, in check_schema
    raise SchemaError.create_from(error)
SchemaError: <object object at 0x000000010b2a8020> is not of type u'string'

Failed validating u'type' in metaschema[u'properties'][u'pattern']:
    {u'format': u'regex', u'type': u'string'}

On schema[u'pattern']:
    <object object at 0x000000010b2a8020>
```

Are you experiencing some specific issue here that's prompting adding this?

manycoding:
@Julian 
Yes, I meant the case when the regexp is a valid string, but an invalid python regex.
```
>>> Draft6Validator.check_schema({"pattern": "\q"})
>>> jsonschema.validate("s", {"pattern": "\q"})
  File "<stdin>", line 1, in <module>
  File "/Users/valery/Documents/_code/jsonschema/jsonschema/validators.py", line 822, in validate
    cls(schema, *args, **kwargs).validate(instance)
  File "/Users/valery/Documents/_code/jsonschema/jsonschema/validators.py", line 304, in validate
    for error in self.iter_errors(*args, **kwargs):
  File "/Users/valery/Documents/_code/jsonschema/jsonschema/validators.py", line 280, in iter_errors
    for error in errors:
  File "/Users/valery/Documents/_code/jsonschema/jsonschema/_validators.py", line 248, in pattern
    not re.search(patrn, instance)
  File "/Users/valery/.local/share/virtualenvs/jsonschema-zk5Nmia0/lib/python3.6/re.py", line 182, in search
    return _compile(pattern, flags).search(string)
  File "/Users/valery/.local/share/virtualenvs/jsonschema-zk5Nmia0/lib/python3.6/re.py", line 301, in _compile
    p = sre_compile.compile(pattern, flags)
  File "/Users/valery/.local/share/virtualenvs/jsonschema-zk5Nmia0/lib/python3.6/sre_compile.py", line 562, in compile
    p = sre_parse.parse(p, flags)
  File "/Users/valery/.local/share/virtualenvs/jsonschema-zk5Nmia0/lib/python3.6/sre_parse.py", line 855, in parse
    p = _parse_sub(source, pattern, flags & SRE_FLAG_VERBOSE, 0)
  File "/Users/valery/.local/share/virtualenvs/jsonschema-zk5Nmia0/lib/python3.6/sre_parse.py", line 416, in _parse_sub
    not nested and not items))
  File "/Users/valery/.local/share/virtualenvs/jsonschema-zk5Nmia0/lib/python3.6/sre_parse.py", line 502, in _parse
    code = _escape(source, this, state)
  File "/Users/valery/.local/share/virtualenvs/jsonschema-zk5Nmia0/lib/python3.6/sre_parse.py", line 401, in _escape
    raise source.error("bad escape %s" % escape, len(escape))
sre_constants.error: bad escape \q at position 0
```

If it happens it causes an exception which is unwanted in a typical use case.

Julian:
The right place for that to have raised an exception is in the first line
there (the check_schema line). Looks like the reason it didn't is probably
just that we don't currently pass format checkers through when validating
against the meta schema, but that seems like possibly a reasonable change
to make (and even be the default).

manycoding:
@Julian Would you recommend the best way to handle `patternProperties`?

I thought about calling pattern from within `addittionalProperties()` but it doesn't work and I have no idea why

```
def additionalProperties(validator, aP, instance, schema):
    if not validator.is_type(instance, "object"):
        return

    extras = set(_utils.find_additional_properties(instance, schema))

    if validator.is_type(aP, "object"):
        print("additionalProperties")
        # print(f"extras {extras}")
        for extra in extras:
            print("here")
            pattern(validator, extra, instance[extra], aP)
            for error in validator.descend(instance[extra], aP, path=extra):
                yield error
```

Julian:
Hey, sorry, things are a bit hectic, have just returned from vacation and started a new job :)

To elaborate, it's the [`check_schema` method](https://github.com/Julian/jsonschema/blob/3eabefb0c538d44bec414d84a7ab576dac5de6ee/jsonschema/validators.py#L248-L250) that should be changing -- namely, it should grow a `format_checker` argument, which it should default to *creating* one (and then passing it into the validator instance it creates). I think other than that two line change (and the associated tests) nothing else should need to change.

Does that help?

manycoding:
@Julian Hi, I passed the format_checker but I cannot figure out how to create a default one. As far as I understand, each meta schema should have it's own format checker .

Also, it doesn't check patternProperties since the format_checker checks only the value, not the instance name.

Last but not least, I added tests for both (patternProperties obviously fails).

Julian:
Hey! Thanks for continuing to tackle this!

I have to re-review your patch, but on the patternProperties front, I believe that the recent drafts added `propertyNames`, which is a thing that can apply schemas to properties.

There's a possibility though that the meta schema isn't actually trying to apply that to `patternProperties` / `properties`, which I think would be a bug to file upstream if so.

But tl;dr -- I think if what you've got isn't catching `patternProperties` issues, that that needs addressing upstream (in the spec, or in the meta schemas).

Will try to leave some additional comments after reviewing, but yeah thanks again, this is definitely appreciated!

manycoding:
@Julian Hi, I would appreciate if you review this one once you have a free minute.

Julian:
Thanks! This is looking close to what I expected. Looks like CI is failing though. Possibly you just need to merge in master.

jsonschema/tests/test_validators.py
```diff
@@ -978,9 +989,9 @@ def test_it_returns_true_for_formats_it_does_not_know_about(self):
         )
         validator.validate("bugs")
 
-    def test_it_does_not_validate_formats_by_default(self):
+    def test_it_validates_formats_by_default(self):
```
manycoding: Also updated this one

manycoding:
I merged master and also updated this one https://github.com/Julian/jsonschema/pull/425/files#diff-523a4ebdb045b5a1613d6711add86c7fL981 and disabled https://github.com/Julian/jsonschema/pull/425/files#diff-523a4ebdb045b5a1613d6711add86c7fR889

jsonschema/tests/test_validators.py
```diff
@@ -978,9 +989,9 @@ def test_it_returns_true_for_formats_it_does_not_know_about(self):
...
         validator = self.Validator({})
-        self.assertIsNone(validator.format_checker)
+        self.assertIsInstance(validator.format_checker, FormatChecker)
```
Julian: This test shouldn't pass, so if it does, something seems wrong -- the goal here wasn't to change the default for all validators, it was just to do so for calling `.check_schema`

manycoding:
I made suggested updates, there are still some failures though:
1. I guess test_it_does_not_validate_formats_by_default fails because I pass format_checker in create method
2. I have no idea why all those ref tests for TestDraft43 TestDraft4 are failing

jsonschema/validators.py
```diff
@@ -220,7 +227,7 @@ def __init__(
             schema,
             types=(),
             resolver=None,
-            format_checker=None,
+            format_checker=format_checker,
```
Julian: It's failing because of this line, where you've changed the default here.

jsonschema/validators.py
```diff
@@ -257,7 +265,7 @@ def iter_errors(self, instance, _schema=None):
                 return
             elif _schema is False:
                 yield ValidationError(
-                    "False schema does not allow %r" % (instance,),
+                    "False schema does not allow '{}'".format(instance)
```
Julian: Can you undo this too?

jsonschema/validators.py
```diff
@@ -20,6 +20,12 @@
     ValidationError,
 )
 
+from jsonschema._format import (
+    draft3_format_checker,
+    draft4_format_checker,
+    draft6_format_checker,
```
Julian: Draft 7 is on master too, so would be great if you just add one more line here and below to do the same as what you've done for draft 7.

Julian:
Think we're close!

manycoding:
Done. Can it be that that passed format_checker messes with uri format_checker?

jsonschema/tests/test_validators.py
```diff
@@ -880,6 +880,17 @@ def test_invalid_properties(self):
         with self.assertRaises(SchemaError):
             self.Validator.check_schema({"properties": {"test": object()}})
 
+    def test_invalid_pattern(self):
+        with self.assertRaises(SchemaError):
+            self.Validator.check_schema({"pattern": "\q"})
+
+    @unittest.skip("This test fails but it shouldn't, see"
```
Julian: Right so here, this test doesn't fail on draft 7, which properly says that the keys need to be regexes for `patternProperties`. It does that though using a new validator (`propertyNames`) that didn't exist before draft 6.

Draft 6 though seems to have a bug in the metaschema (by not specifying propertyNames for patternProperties even though that existed for draft 6), which sounds like something that should be filed  upstream.

And drafts 3 and 4 don't have that, so that's why this fails -- so it seems like the right thing to do isn't to skip this here, it should pass on draft 7, and draft 6 when/if the metaschema is fixed.

Julian: Ah, right, I think the reason it's not there in draft6 is because draft6 (and 4) accidentally removed the `regex` format.

So, this seems to be a thing that really only should pass on draft 7, and "correctly" (FSVO correctly) blows up on earlier drafts.

manycoding: Understood, how can I skip the test on particular drafts? I don't understand the skip lambda https://github.com/Julian/jsonschema/blob/master/jsonschema/tests/test_jsonschema_test_suite.py#L81

manycoding:
I don't understand skip lambda, how can I skip tests with it?
I think some test cases are generated but I cannot find where, for example `valid tree` keywords.

Julian:
@manycoding hey! So I wouldn't bother with that stuff here, what we need here is just a test case we only run for draft 6 and 7 -- the most straightforward way to do that is just a `InvalidRegexMixin` class that you then mix in to 2 test cases, a `TestDraft6` one and a `TestDraft7` one. If that doesn't make eminent sense (I don't remember if those 2 test cases still exist or not) let me know, it's fairly easy so if I'm unclear I can give you a specific line to look at.

manycoding:
I moved pattern tests to their own draft class, but it seems that my code affects ref in Draft3 and 4.

manycoding:
@Julian I really would like to finish this one in this year :)

Julian:
Hey @manycoding so would I!

I've unfortunately got extremely limited time, which currently even for `jsonschema` needs to be prioritized on getting a production-ready draft 6/7 release out. Have tried to give you some tips to help you find out why your change is breaking things but really would need to be you figuring out how to get all of them to pass until I manage to have time to help.

Definitely appreciate the PR.

Julian:
@manycoding FWIW I am going to try to *finally* find some time to give you another breadcrumb (or ideally figure out what's up here) some time next week :/

Fingers crossed!

PR #394, opened 2018-03-01
Update setup.py syntax for pipenv (bug) compliancy

apiraino:
Hi,

I'm using your library (cool, thanks!) and I'm migrating my project to using [pipenv](https://docs.pipenv.org/) instead of pip.

Pipenv [has a bug](https://github.com/pypa/pipenv/issues/1279#issuecomment-364623908) that prevents to correctly parse the old `extra_requires` syntax. Apparently this syntax is old and superseded by `install_requires`.

In short, what happens is that in my py3 virtualenv, pipenv wrongly installs `functools32` and that breaks the whole package installation process. 

This patch allows me to use `jsonschema` in my project migrated to pipenv.

I'd be happy to have your opinion on this small patch, thanks!

PR #426, opened 2018-06-30
functools32 into extra requires

macisamuele:
The goal of this PR is to maintain support of older versions of pip/setuptools while maintaining the conditional `functools32` dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,11 @@ _static
 _templates
 
 TODO
+
+.eggs/
+.tox/
+_trial_temp/
+build/
+*.egg-info/
+__pycache__/
+*.pyc
--- a/setup.py
+++ b/setup.py
@@ -40,9 +40,11 @@ setup(
         "attrs>=17.4.0",
         "pyrsistent>=0.14.0",
         "six>=1.11.0",
-        "functools32;python_version<'3'",
     ],
     extras_require={
+        ":python_version<'3.2'": [
+            "functools32",
+        ],
         "format": [
             "jsonpointer>1.13",
             "rfc3987",


.gitignore
```diff
@@ -3,3 +3,11 @@ _static
 _templates
 
 TODO
+
```
Julian: Please keep PRs to doing one single thing.

In general here in particular has been "this belongs in your global gitignore, not the one for every repo, which is why it isn't here already", but if you disagree, we should discuss in a separate PR :)

setup.py
```diff
@@ -40,9 +40,11 @@
         "attrs>=17.4.0",
         "pyrsistent>=0.14.0",
         "six>=1.11.0",
-        "functools32;python_version<'3'",
     ],
     extras_require={
+        ":python_version<'3.2'": [
```
Julian: So, this [undos a previous PRs change](https://github.com/Julian/jsonschema/pull/394/files).

I should have said it then, which was my mistake, but this needs tests :). I'm inclined to prefer supporting newer tools than older setuptools in the event that supporting both isn't possible -- IIRC the versions of setuptools that this doesn't work in are super super old -- but regardless, if support for setuptools is useful to you, it'd need to have something in CI ensuring that it stays working :). Tests I guess are needed now to ensure `pipenv` compatibility too.

And tangentially -- 3.2 isn't supported (actually IIRC even 3.3 isn't supported), so the two ways of writing this conditional are essentially equivalent here.

.gitignore
macisamuele: My personal opinion is that each repo should ignore the relevant files.
But as you mentioned this goes a bit beyond the real goal of this PR. 
So I'll remove for now the gitignore changes

setup.py
macisamuele: I see the point of preferring support for newer versions rather than old versions.
The proposed change should allow support of older and newer version, that's the main reason I proposed it.

About the constraint on python 3.2 it was mostly to keep accurate dependencies, even if your library does not provide official support for some python versions nothing prevents users to use them.
Using the accurate constraint has no real cost and could reduce issues for users on unsupported python versions (this is mostly a theoretical thinking as I would not use a library not officially supported for the python version in use :))

Lastly, about testing. Do you have something specific in mind?

Julian: > I see the point of preferring support for newer versions rather than old versions.
The proposed change should allow support of older and newer version, that's the main reason I proposed it.

Please see the link I posted. There's apparently a bug in `pipenv` (a newer tool) that doesn't like the old `extras_require` syntax, which is how this used to be written. My comment was "if the choice is support pipenv or support older setuptools, I'd prefer the former".

> Using the accurate constraint has no real cost 

> Using the accurate constraint has no real cost and could reduce issues [...]

This is the literal definition of "support" -- if there are changes that are made strictly to support versions that are not supported, the cost is *making* those changes :) [and of course ensuring that they don't break anything else in the process]. In this case it may look "trivial", but as usual with software there are always surprises, so sometimes inclination is to take a hard line. I believe in this case that anyone on 3.2 or less likely deserves that hard line, those versions are essentially unusable.

> Lastly, about testing. Do you have something specific in mind?

A test that fails before this change and succeeds afterwards -- specifically, if version X.Y of setuptools does not work before this, there should be an automated test that installs that version of setuptools, then this package, then makes an assertion about the package functioning, which should fail before this change and succeed afterwards.

The same should be for `pipenv`, we'd need a test that asserts it functions properly given the way the `setup.py` is written at any given point in time.

macisamuele: I was not aware of the `pipenv` issue and for what I see it does not look that the new system will support anytime soon the old format that allows `setuptools<20.2` compatibility. 

So I'm fine in closing myself this PR as this will provide better support for new tools. After all the workaround for my specific case is to update setuptools to something newer

Julian: No worries, and thanks, the PR is appreciated even if it's not going to be merged!

PR #427, opened 2018-07-02
Fix unclosed resource in validators

ApamNapat:
The resource used by `urlopen` was never explicitly closed, running an app with pythons `-W all` flag raised loads of `ResourceWarning`s. Using a context manager is a simple fix.

### DIFF
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1199,8 +1199,8 @@ class TestRefResolver(TestCase):
                     self.assertEqual(resolved, 12)
 
         urlopen_mock.assert_called_once_with("http://bar")
-        resource_manager_mock.__enter__.assert_called_once()
-        resource_manager_mock.__exit__.assert_called_once()
+        self.assertEqual(resource_manager_mock.__enter__.call_count, 1)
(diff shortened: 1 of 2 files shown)


Julian:
Thanks! Really appreciated.

Could you write a test to ensure this doesn't regress?

ApamNapat:
@Julian I fixed and extended an already existing test to check for context manager's calls. If there is still something lacking please let me know.

jsonschema/tests/test_validators.py
```diff
@@ -1186,14 +1186,21 @@ def test_it_retrieves_unstored_refs_via_requests(self):
     def test_it_retrieves_unstored_refs_via_urlopen(self):
         ref = "http://bar#baz"
         schema = {"baz": 12}
+        resource_manager_mock = mock.MagicMock()
+        (resource_manager_mock.
+         __enter__.return_value.
+         read.return_value.
+         decode.return_value) = json.dumps(schema).encode("utf8")
```
Julian: This doesn't look correct, and is a good example of why mocking is sadness :/

`decode` returns unicode, not bytes.

Will fix on merge though, thanks.

Julian:
Merged, thanks!

PR #428, opened 2018-07-05
Use uritools instead of urlparse, and support proper file scheme.

agoose77:
Currently, the ref resolver opens files (or any other unsupported scheme) implicitly. Instead, we should be using a "file" scheme in the URI. In addition, the use of urllib for RFC3986 URIs is discouraged, as it does not entirely conform to the specification. 

This PR does the following:
1. Use `uritools` functions instead of the `urllib` equivalents
1. Use the existing scheme map to handle "http", "https", and "file" schemes with `requests` / `urlopen`.
1. Require an explicit scheme "file://" for files, and raise `ValueError` for any unsupported scheme. Requests does not handle "file" schemes by default, but there exists [an adaptor which implements this support](https://github.com/dashea/requests-file). If the library cannot be imported, we default to `urlopen` which supports the "file" scheme.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -15,7 +15,8 @@ class URIDict(MutableMapping):
 
     def normalize(self, uri):
         result = urisplit(uri)
-        return uriunsplit((result.scheme, result.authority, result.path, None, None))
+        return uriunsplit((result.scheme, result.authority,
+                           result.path, None, None))
 
     def __init__(self, *args, **kwargs):
         self.store = dict()
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -4,7 +4,7 @@ import pkgutil
 import re
 
 from jsonschema.compat import str_types, MutableMapping
-from uritools import urisplit
+from uritools import urisplit, uriunsplit
 
 
 class URIDict(MutableMapping):
@@ -14,7 +14,8 @@ class URIDict(MutableMapping):
     """
 
     def normalize(self, uri):
-        return urisplit(uri).geturi()
+        result = urisplit(uri)
+        return uriunsplit((result.scheme, result.authority, result.path, None, None))
 
     def __init__(self, *args, **kwargs):
         self.store = dict()
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -3,7 +3,8 @@ import json
 import pkgutil
 import re
 
-from jsonschema.compat import str_types, MutableMapping, urlsplit
+from jsonschema.compat import str_types, MutableMapping
+from uritools import urisplit
 
 
 class URIDict(MutableMapping):
@@ -13,7 +14,7 @@ class URIDict(MutableMapping):
     """
 
     def normalize(self, uri):
-        return urlsplit(uri).geturl()
+        return urisplit(uri).getpath()
 
     def __init__(self, *args, **kwargs):
         self.store = dict()
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -14,7 +14,7 @@ class URIDict(MutableMapping):
     """
 
     def normalize(self, uri):
-        return urisplit(uri).getpath()
+        return urisplit(uri).geturi()
 
     def __init__(self, *args, **kwargs):
         self.store = dict()
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1217,7 +1217,7 @@ class TestRefResolver(TestCase):
                     self.assertEqual(resolved, 12)
         requests.Session().get.assert_called_once_with("file://bar.json")
 
-    def test_it_retrieves_unstored_refs_via_urlopen_without_requests_file(self):
+    def test_it_retrieves_unstored_refs_via_urlopen_no_requests_file(self):
         ref = "file://bar.json#baz"
         schema = {"baz": 12}
 
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1195,6 +1195,41 @@ class TestRefResolver(TestCase):
                     self.assertEqual(resolved, 12)
         urlopen.assert_called_once_with("http://bar")
 
+    def test_it_retrieves_unstored_file_refs_via_urlopen(self):
+        ref = "file://bar.json#baz"
+        schema = {"baz": 12}
+
+        with MockImport("requests", None):
+            with mock.patch("jsonschema.validators.urlopen") as urlopen:
+                urlopen.return_value.read.return_value = (json.dumps(schema).encode("utf8"))
+                with self.resolver.resolving(ref) as resolved:
+                    self.assertEqual(resolved, 12)
+        urlopen.assert_called_once_with("file://bar.json")
+
+    def test_it_retrieves_unstored_file_refs_via_requests_and_requests_file(self):
+        ref = "file://bar.json#baz"
+        schema = {"baz": 12}
+
+        with MockImport("requests", mock.Mock()) as requests:
+            with MockImport("requests_file", mock.Mock()) as requests_file:
+                requests.Session.get.return_value.json.return_value = schema
+                with self.resolver.resolving(ref) as resolved:
+                    self.assertEqual(resolved, 12)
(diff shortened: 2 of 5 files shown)


Julian:
Hi thanks, really appreciated.

This all would need unit tests, would you be able to take a shot at them?

Thanks!

agoose77:
@Julian Of course, actually I noticed there were some bugs which were failing the tests, so I'm just fixing those first.

jsonschema/_utils.py
```diff
@@ -13,7 +14,9 @@ class URIDict(MutableMapping):
     """
 
     def normalize(self, uri):
-        return urlsplit(uri).geturl()
+        result = urisplit(uri)
```
agoose77: Hmm, if we're concerned about the performance of these methods, we should probably cache this. I suggest passing in the cache as a special keyword argument (py2 doesn't support kw-only args), or by using a factory which returns a new type of URIDict with a particular cache configuration. Thoughts?

jsonschema/validators.py
```diff
@@ -595,9 +598,38 @@ def from_schema(
...
         return cls(base_uri=id_of(schema), referrer=schema, *args, **kwargs)
 
+    def http_handler(self, uri):
+        try:
+            import requests
+        except ImportError:
+            pass
+        else:
+            if hasattr(requests.Response, "json"):
+                session = requests.Session()
```
Julian: This code essentially creates and discards a session each time it's invoked. Seems like it definitely should be holding onto it instead (and likely allowing a user to pass one in). The previous code would be using the global requests session.

jsonschema/validators.py
```diff
@@ -595,9 +598,38 @@ def from_schema(
...
+
+                requests_supports_scheme = True
+                if urisplit(uri).scheme == "file":
+                    try:
+                        import requests_file
```
Julian: This library doesn't seem too widely used. Is it actually?

jsonschema/validators.py
```diff
@@ -522,15 +523,15 @@ class RefResolver(object):
             A mapping from URI schemes to functions that should be used
             to retrieve them
 
-        urljoin_cache (functools.lru_cache):
+        urijoin_cache (functools.lru_cache):
```
Julian: This would need to stay as-is for backwards compatibility.

Not sure the URL/URI distinction is too important anymore anyhow.

jsonschema/validators.py
```diff
@@ -641,28 +673,28 @@ def resolving(self, ref):
...
+        self.push_scope(uri)
         try:
             yield resolved
         finally:
             self.pop_scope()
 
     def resolve(self, ref):
-        url = self._urljoin_cache(self.resolution_scope, ref)
-        return url, self._remote_cache(url)
+        uri = self._urijoin_cache(self.resolution_scope, ref)
```
Julian: See above, this seems like just diff churn, and can't anyhow be renamed for backwards compatibility reasons.

jsonschema/validators.py
```diff
@@ -641,28 +673,28 @@ def resolving(self, ref):
...
+            document = self.store[uri]
         except KeyError:
             try:
-                document = self.resolve_remote(url)
+                document = self.resolve_remote(uri)
             except Exception as exc:
                 raise RefResolutionError(exc)
 
-        return self.resolve_fragment(document, fragment)
+        return self.resolve_fragment(document, fragment or '')
```
Julian: I don't see a test covering this change.

setup.py
```diff
@@ -41,6 +41,7 @@
         "pyrsistent>=0.14.0",
         "six>=1.11.0",
         "functools32;python_version<'3'",
+        "uritools>=2.2.0",
```
Julian: I would at some point like to be using [hyperlink](https://pypi.org/project/hyperlink) here. Seems likely that we can make due without using `uritools` if we use it.

jsonschema/tests/test_validators.py
```diff
@@ -1195,6 +1195,42 @@ def test_it_retrieves_unstored_refs_via_urlopen(self):
                     self.assertEqual(resolved, 12)
         urlopen.assert_called_once_with("http://bar")
 
+    def test_it_retrieves_unstored_file_refs_via_urlopen(self):
```
Julian: I'd love to start moving towards real tests instead of the mocking nonsense that happens in the existing ones here (i.e. reading a real file that is shipped as part of the test suite).

It's very difficult to see whether these tests pass even before your change other than just needing to mock the right data into place.

jsonschema/validators.py
agoose77: Agreed, lazy on my part.

jsonschema/_utils.py
Julian: I wouldn't worry about performance until/unless it's observed to be slow (in which case something like that sounds reasonable).

jsonschema/validators.py
agoose77: Not particularly, but in all it's quite simple (but not simple enough to warrant copying the source imo). Off the top of my head, urllib2 has a builtin file schema handler. There are several options to this:

1. Remove requests file and re-write it (see https://stackoverflow.com/a/22989322/1215241)
1. Don't use requests for files, just use urllib2 / urllib.request
1. Don't use any http getter and manually open the file with open() (see https://github.com/python/cpython/blob/3.7/Lib/urllib/request.py#L1976)

I avoided the third option because my knowledge of the path handling quirks is limited. Feel free to make a suggestion here.

jsonschema/validators.py
agoose77: Sure, I made a judgement that it wasn't used internally so it could be removed, in hindsight not a good decision.

jsonschema/validators.py
agoose77: (y)

jsonschema/validators.py
agoose77: It's a consequence of the implementation differing (the original uri method does this internally)

setup.py
agoose77: Hmm, I had looked at this problem a while ago and concluded that uritools was the best tool. Seems that I missed hyperlink. I agree may be a better solution

jsonschema/tests/test_validators.py
agoose77: :thumbsup:
Some of the mocks are indeed quite fragile, I'll make that change as necessary. Not sure where to write to/from, perhaps a tmpfile?

jsonschema/validators.py
Julian: Got it, thanks! So an existing test fails without this change?

jsonschema/tests/test_validators.py
Julian: Yeah or just a real file that's in the `jsonschema.tests` directory which is likely even easier, though you still need to use like a relative file reference presumably.

agoose77: Do you have gitter or anything better for more general q's on the jsonschema codebase?

jsonschema/_utils.py
agoose77: I mention performance because after publishing the other PR (now closed), I noticed that the use of explicit push_ / pop_ scope methods was as a result of loop unrolling due to performance (three years ago).

jsonschema/validators.py
agoose77: The original method of `urldefrag` returns '' if the fragment is None, whereas `uridefrag` returns None :)

jsonschema/tests/test_validators.py
Julian: I'm on IRC pretty often (and when I'm not I'm idling) -- there's a #json-schema channel that I'm in.

There's also the JSON Schema Slack: https://join.slack.com/t/json-schema/shared_invite/enQtMjk1NDcyNDI2NTAwLTcyYmYwMjdmMmUxNzZjYzIxNGU2YjdkNzdlOGZiNjIwNDI2M2Y3NmRkYjA4YmMwODMwYjgyOTFlNWZjZjAyNjg

Which I'm on, and you can certainly ping me on!

(And thanks again)

jsonschema/validators.py
agoose77: Hmm, this is already done internally with `get` by default, so this is no slower than the current approach.

PR #429, opened 2018-07-06
Use context managers instead of unrolled scopes

agoose77:
Depends upon GH-428.

I've seen #158 which suggests that this is a deliberate feature. Perhaps then this isn't appropriate.

However, in https://github.com/agoose77/jsonschema/blob/master/jsonschema/_validators.py#L304 we have more complicated logic where we look for the contextmanager, and use it if available, otherwise defaulting to an unrolled loop. As far as I can tell, this is to facilitate the mock ref resolver which only provides the context manager. If performance is still a concern, I suggest we eliminate one branch.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -15,7 +15,8 @@ class URIDict(MutableMapping):
 
     def normalize(self, uri):
         result = urisplit(uri)
-        return uriunsplit((result.scheme, result.authority, result.path, None, None))
+        return uriunsplit((result.scheme, result.authority,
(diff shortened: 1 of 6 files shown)


Julian:
See my comments on the other PR which would likely apply here too, it's certainly harder to deal with this if it's dependent on the other one, but the one change I can see here does break backwards compatibility as well, so depending on what the goal is here it can't be merged as-is, but possibly there's a solution that covers whatever the goal is.

agoose77:
Leave this for now.

PR #430, opened 2018-07-06
Fix bug in current registration of validators and address #417

agoose77:
Alternative solution to #417 (fixes #417) 
Currently, the validator registration is hardcoded to use the deprecated "id" keyword, rather than using the new "id_of" function passed to `create`. This means that only draft 3 and 4 are currently registered. This is not visibly a bug at present, because there is only one other schema, and the fallback mechanism selects the most recent. However, for good practice we should still register the draft 6 schema.

This PR adds a `Validator.ID_OF` staticmethod, and call it from the registrar.

### DIFF
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1030,6 +1030,7 @@ class TestValidatorFor(TestCase):
         Validator = validators.create(
             meta_schema={"id": "meta schema id"},
             version="12",
+            id_of=lambda s: s.get("id", ""),
         )
         schema = {"$schema": "meta schema id"}
         self.assertIs(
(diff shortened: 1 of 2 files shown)


robherring:
Presumably @Julian will want test cases for this. You can extract the ones I wrote from #417 if you want.

agoose77:
Hmm, do you think we need to test this? I suppose we could check all the current drafts exist in the registry.
If you have those tests already, that would be very helpful!

Julian:
Yes, every change needs tests that cover every part of it :)

@robherring I think I prefer this way a bit, but yours has tests :), so will see if we can combine those together.

Julian:
OK, merged! Thanks.

PR #431, opened 2018-07-06
The `validates` decorator is not needed in the current form.

agoose77:
Instead, create a new function `register_validator` which is called from `validates`.
Test `register_validator` instead of `validates`, and deprecate `validates` use internally.

### DIFF
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -62,17 +62,16 @@ class TestCreateAndExtend(TestCase):
         )
 
     def test_if_a_version_is_provided_it_is_registered(self):
-        with mock.patch("jsonschema.validators.validates") as validates:
-            validates.side_effect = lambda version: lambda cls: cls
+        with mock.patch("jsonschema.validators.register_validator") as validates:
             Validator = validators.create(
                 meta_schema={u"id": ""},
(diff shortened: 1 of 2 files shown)


Julian:
Not sure I understand the motivation here, what's the advantage to doing this? It just looks like a different way to accomplish the same goal.

agoose77:
It's a small change - I suggest you remove the `validates` decorator (but I didn't just remove it because it *could* be used externally (not sure why though)

Julian:
Yes I'm asking *why* :)?

Is `register_validator("foo", Foo)` meaningfully better than `validates("foo")(Foo)`, enough so to invalidate all the existing code that does the latter already?

agoose77:
Personally, I think so - I realise that this is "small change" in terms of code (barely worth a code review), but using a decorator API in a non-decorator context can feel clunky and obfuscated (albeit very mildly) the purpose of the decorator. If you think it's a negligible change, feel free to close the PR!

Julian:
Think I'll pass here even though it's appreciated!

It's true if this was a new API it likely would have looked like the latter, but I don't see it as inconvenient enough to justify churning its API -- someday perhaps it'll churn to get rid of it being global state, at which point maybe this will happen too.

PR #439, opened 2018-07-20
jsonschema: Fix non http/https url RefResolution exception

knsathya:
When using file:///schema-dir-name URLs we are hitting AttributeError
exception during refresolution. Reason for it is, there isn't any
context manager implemented for urlopen (i.e. it doesn't have
__enter__ and __exit__ methods defined).

This patch fixes this issue.

Signed-off-by: Kuppuswamy Sathyanarayanan <sathyaosid@gmail.com>

### DIFF
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -755,8 +755,8 @@ class RefResolver(object):
                 result = requests.get(uri).json
         else:
             # Otherwise, pass off to urllib and assume utf-8
-            with urlopen(uri) as url:
-                result = json.loads(url.read().decode("utf-8"))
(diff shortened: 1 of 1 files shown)


Julian:
This is a bug introduced by #427, which was just merged.

I was unhappy with the test submitted, but didn't have enough time to fully fix it or explain to the submitter why :/

That'll teach me I suppose.

The test here needs to be a "normal" test, no mocking (#335).

The correct implementation of this likely will use `contextlib.closing`, in order to fix #427's underlying issue as well as this one.

Would you be able to try writing one?

Issue #420, opened 2018-06-04
Handling YAML $refs


martingkelly:
I'm using currently handling schemas written in YAML (for human-readability) but validated by jsonschema. I parse the schema in YAML (and make sure I subset my YAML to stay JSON-compatible) and then run the parsed dictionary through jsonschema. It is working nicely with one issue: When I want to use file:// includes, the files are implicitly assumed to be json.

@Julian I would like to add support for these files being YAML and am wondering if this is something you would support. If so, what is your recommendation for the cleanest way to add it? If it's not something you support, what would you recommend as the cleanest way to add the support for my own code without hacking/forking jsonschema? I thought about adding custom resolver as is done in https://github.com/Julian/jsonschema/issues/225, but it sounds like `RefResolver` is not intended to be API-stable, so subclassing is liable to break in the future. Is there another way?

Thanks,
Martin

martingkelly:
Upon further code reading, I found the `handlers` argument to `RefResolver`. If I attach a `yaml://` handler (or similar), I can make this work. That said, it feels hacky, in that `yaml://` is not really a location (like `file://` or `http://`) and it would be better to decouple the file format (YAML/JSON) from the location (including the logic for handling base URIs and $id).

robherring:
@martingkelly I'm doing the same thing with YAML and have $ref's working. See here: https://github.com/robherring/yaml-bindings/blob/master/dtschema.py#L212

martingkelly:
Thanks @robherring, I ended up doing the same thing. I think it's not ideal in that by implementing your own handler, you lose the more general caching logic, and you lose the ability to do JSON http... ideally the document format and transport would be better decoupled.

Btw, I really like your device-tree YAML project; I think it's an excellent idea :).

handrews:
You can see one approach to this problem in JavaScript in the [AutoExtensionDereferencer](https://github.com/cloudflare/json-schema-tools/tree/master/workspaces/json-schema-transform) function of [JSON Schema Tools](https://github.com/cloudflare/json-schema-tools/tree/master/workspaces/json-schema-transform).  Although the "dereferencer" part is not necessarily relevant here, just the automatic file extension search approach, which also understands overriding the URI scheme to load from a file as if it were a `file://` URI.

martingkelly:
Yes, something like that would work. @Julian any thoughts?

Julian:
[...]

PR #445, opened 2018-08-06
Add referrer to handler's call

mes3yd:
For handlers to resolve relative references, they need a referrer to be
passed in.

Looking at the code and tests especially I'm starting to doubt if this was the intent of the referrer argument. It can be used in this way, but what resolver.resolve() returns as first member of the tuple is a bit magical to me. However, we do need a way to be able to resolve relative references.

The code also introduces a default scheme, with a default conforming to what is normal in the URI resolver world.

### DIFF
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1260,7 +1260,7 @@ class TestRefResolver(TestCase):
         )
         with resolver.resolving(ref) as resolved:
             self.assertEqual(resolved, schema)
-        foo_handler.assert_called_once_with(ref)
+        foo_handler.assert_called_once_with(ref, referrer={})
 
     def test_cache_remote_on(self):
         ref = "foo://bar"
@@ -1272,7 +1272,7 @@ class TestRefResolver(TestCase):
             pass
         with resolver.resolving(ref):
             pass
-        foo_handler.assert_called_once_with(ref)
+        foo_handler.assert_called_once_with(ref, referrer={})
 
     def test_cache_remote_off(self):
         ref = "foo://bar"
@@ -1301,6 +1301,37 @@ class TestRefResolver(TestCase):
         with self.assertRaises(validators.RefResolutionError) as exc:
             resolver.pop_scope()
         self.assertIn("Failed to pop the scope", str(exc.exception))
+
+    def test_scheme_default_fallback(self):
+        ref = './paths.json'
+        file_handler = mock.Mock()
+        resolver = validators.RefResolver(
+            "", "", handlers={"file": file_handler}
+        )
+        with resolver.resolving(ref):
+            pass
+        file_handler.assert_called_with(ref, referrer='')
+
+    def test_relative_resolver(self):
+        def file_handler(uri, referrer=None):
+            # Crude, do not use as a reference implementation
+            import os.path
+            no_scheme = referrer[7:]  # file:/// 0-based
+            return 'file://' + os.path.abspath(os.path.join(no_scheme, uri))
+
+        ref = '../../schemas/common.json'
+        base = 'file:///home/julian/spec/'
+        referrer = base + 'components/common.json'
+        expected = base + 'schemas/common.json'
+        resolver = validators.RefResolver(
+            "", referrer, handlers={"file": file_handler}
+        )
+        with resolver.resolving(ref):
+            key, resolved = resolver.resolve(ref)
+
+        # Why is it this, really?
+        self.assertEqual('schemas/common.json', key)
+        self.assertEqual(expected, resolved)
 
 
 class UniqueTupleItemsMixin(object):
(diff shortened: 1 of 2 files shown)


mes3yd:
If this is part of the jsonschema spec (and I admit, I haven't read it), then you're right it's the wrong place to fix it. From a practical point of view, you'd need to pre-process documents that have relative references and it really wouldn't make sense anymore to use jsonschema's RefResolver class or the handler interface.
So perhaps it makes more sense to create a `RelativeRefResolver` class that has the API contract I'm proposing? That way there's no backwards compatibility issue. If you agree, I'll rework using that approach.

Julian:
> From a practical point of view, you'd need to pre-process documents that have relative references

I don't think you need to preprocess do you?

The `RefResolver` should know what the base URI is, so it's just that each handler shouldn't have to deal with resolving relative references, the `RefResolver` before it calls the handler resolves the relative reference to its base URI.

(By the way I know close to 0 about URIs so apologies in advance if anything I have said, say, or will say makes negative amounts of sense :)

Julian:
> P.S. I'm guessing the PyPy failure has nothing to do with my PR.

It is actually, it's [pointing out](https://travis-ci.org/Julian/jsonschema/jobs/412749661#L2447) that [`urlsplit`](https://github.com/Julian/jsonschema/pull/445/files#diff-ed2f83de8c2ff5b1c1138f9f28ba4ba6R545) isn't a word, though sphinx-spelling's output is not exactly easy to read.

The fix there is to use a Sphinx reference (`urlsplit`), but let's make sure we agree first on whether this makes sense before too many specific comments on the implementation.

There's also #306 and to a lesser extent #434 open which might be relevant to have a look at.

mes3yd:
> The RefResolver should know what the base URI is, so it's just that each handler shouldn't have to deal with resolving relative references, the RefResolver before it calls the handler resolves the relative reference to its base URI.

Relative references are relative to the document they are specified in (the referrer). Which is why even with a base URI, you cannot translate correctly if you don't know the location of the referrer:
```python
base = '/home/julian'
uri = '../common/schemas.yaml'
referrer = '/home/julian/endpoint/paths.yaml'
```
With just `base` and `uri`, you can only come up with `/home/common/schemas.yaml` as the absolute uri. Adding referrer to the mix as the document where the reference is made, you will see that the intended relative reference is `/home/julian/common/schemas.yaml' (since a double dot from a document refers to the parent of it's container, not it's container).

Now if we go to section [5.1.3 of RFC 3986](https://tools.ietf.org/html/rfc3986#section-5.1.3), then one could argue that each `RefResolver` should be handed a different base uri, corresponding to the containing document. So there would be no global base, but `referrer` would be passed in as the base. Again, this can't be done as the handler context is unable to keep such state information without the help of jsonschema. At least, I don't see a way to do that, let alone do that thread safe.

Julian:
> Relative references are relative to the document they are specified in (the referrer). Which is why even with a base URI, you cannot translate correctly if you don't know the location of the referrer:

The `RefResolver` has both the [base URI and the referring document](https://github.com/Julian/jsonschema/blob/master/jsonschema/validators.py#L506-L512), and the latter should contain its own URI.

It's still possible I'm missing something basic here certainly, but I'm 90% sure that every handler shouldn't be doing its own URL resolution.

A handler is the equivalent of "given a scheme, go fetch the resource that lives at a given fully, correctly resolved URL with that scheme".

All the actual resolution is (or was intended) to live external to the handler, since my understanding is that that shouldn't be scheme dependent. Is that understanding incorrect?

In theory, if *more* flexibility is required on how to dispatch particular URLs to handlers than just per-scheme, that's where something like #420 comes in too.

mes3yd:
So from your description I defer that with every reference, you should be creating a new RefResolver, using a somewhat static set of handlers and static base url, but variable referrers. This new RefResolver instance should translate the relative reference to an absolute URL (including scheme) and pass that on to the handler.

If the above is correct, then [this code](https://github.com/p1c2u/openapi-spec-validator/blob/master/openapi_spec_validator/validators.py#L67) seems to adhere to that, except I'm still ending up with a RefResolver exception in very simple cases. So I need to investigate deeper where things go wrong.

Julian:
@mes3yd lemme know what you find!

mes3yd:
So finally got around to this: This turned out to be a documentation issue in the upstream library. When you pass a dict to the create_spec function (to obtain the result from a yaml format parsing), the `base_url` is passed in as optional. Technically it is, if you only have one file or abolute URLs. But when using relative URLs the base_url is required.

I am still not sure how the resolver handles this internally, since with only one base URL you need to somehow keep some state somewhere, but that's for another time.

Thanks Julian for the fast responses!

Julian:
Glad you figured it out and yeah definitely let me know if you find
anything further to address.

PR #456, opened 2018-09-03
auto deply with py37

graingert:


.travis.yml
```diff
@@ -16,3 +17,11 @@ addons:
     - libenchant-dev
 git:
   depth: false
+deploy:
+  provider: pypi
+  user: Julian
+  password: ...
```
graingert: You'll need to add your PyPI password here: https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml

.travis.yml
```diff
@@ -16,3 +17,11 @@ addons:
     - libenchant-dev
 git:
   depth: false
+deploy:
+  provider: pypi
+  user: Julian
```
graingert: I'd recommend adding a deploy specific username/password here with maintainer-ship access only to `jsonschema`

eg `julian-jsonschema-deploy`

Julian:
Huh, I somehow completely missed that you pushed to this again after that first discussion, so apologies for letting this sit here...

I'm going to push this (hopefully in a minute) via GitHub actions -- probably first non-automatically (i.e. not on every commit) but will likely move to that after a release or two.

Really appreciated, and yeah sorry for not seeing the last commit.

PR #462, opened 2018-09-22
Return the instance from `validate`

bcb:
The validate function currently returns None. Returning the instance
allows the user to use validate in a function chain, for example:

```python
try:
    return process(validate(deserialize(string)))
except jsonschema.ValidationError:
    ...
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ for Python (supporting 2.7+ including Python 3).
 
     >>> # If no exception is raised by validate(), the instance is valid.
     >>> validate({"name" : "Eggs", "price" : 34.99}, schema)
-    {"name" : "Eggs", "price" : 34.99}
+    {'name' : 'Eggs', 'price' : 34.99}
 
     >>> validate(
     ...     {"name" : "Eggs", "price" : "Invalid"}, schema
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ for Python (supporting 2.7+ including Python 3).
 
     >>> # If no exception is raised by validate(), the instance is valid.
     >>> validate({"name" : "Eggs", "price" : 34.99}, schema)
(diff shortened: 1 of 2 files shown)


Julian:
Hey -- thanks for the PR, it's appreciated.

It's a bit late in jsonschema's life to make this kind of change. If it were being rewritten from scratch it's very likely this is how this would have looked, and someday if/when [Seep](https://github.com/Julian/Seep) gets some love, it has an API like this, but at this point I don't think I want to take this unfortunately.

There's a few reasons other than just inertia -- they're minor but I'll mention them -- one is that in exchange for being able to write the code you put up there (which I'd call only a millimeter better than just using separate lines), someone now will overextend their except block and try to cover exceptions from all 3 functions at the same time -- often incorrectly so, and then be hit by bugs when exceptions they don't expect come from functions they don't expect. `ValueError` is a great example, since that is the awful exception the `json` module blesses us with and is a good possibility for your `deserialize`. There's also another minor one that typically stuff that returns things gets named in past tense, so `validated` here instead of `validate`. And lastly, I'm sure for better or worse someone has (incorrectly) written a line of code that looks like `if jsonschema.validate(...) is None` in attempting to test whether things passed validation, which turns this into a very slight issue of backwards compatibility, since `None` is a perfectly valid instance. These all (even together) were probably ignorable, but pointing them out as well since they do at least come to mind.

Overall I think this is really appreciated, and definitely a reasonable change to propose, but I'd push you towards contributing within Seep if you're trying to make something that looks like this happen!

bcb:
Thanks for the thought-out reply.

Worth mentioning that since Python 3.5 the json module raises a more specific `JSONDecodeError`.

For now I can write my own wrapper `jsonschema.validate`.

Julian:
Ah, good to know!

PR #472, opened 2018-10-02
Fix py2 urlopen

gaetano-guerriero:
urlopen() was used as a context manager on python 2 too.

This happened when resolving remote refs and not using `requests`, for instance when resolving `file:///` refs.


Julian:
Hi! Thanks a lot for this, it'll close #439.

Left a comment on how to shorten up the test -- it also looks like the test might be failing on Windows.

.gitignore
```diff
@@ -1,5 +1,6 @@
 _cache
 _static
 _templates
+*.pyc
```
Julian: Can you back out these unrelated changes so we can focus on just the urlopen fix?

jsonschema/tests/test_validators.py
```diff
@@ -1,23 +1,19 @@
+import json
```
Julian: (And here.)

jsonschema/tests/test_validators.py
```diff
@@ -1,23 +1,19 @@
...
 from decimal import Decimal
 from io import BytesIO
 from unittest import TestCase
-import json
-import sys
-import unittest
 
-from twisted.trial.unittest import SynchronousTestCase
 import attr
+from twisted.trial.unittest import SynchronousTestCase
```
Julian: (And here)

jsonschema/tests/test_validators.py
```diff
@@ -1341,6 +1337,16 @@ def fake_urlopen(url):
             pass
         self.assertEqual(resolved, 12)
 
+    def test_it_retrieves_local_refs_via_urlopen(self):
+        fd, tempf = tempfile.mkstemp()
```
Julian: `tempfile.NamedTemporaryFile(delete=False)` might knock a line or two off of this.

.gitignore
gaetano-guerriero: removed

jsonschema/tests/test_validators.py
gaetano-guerriero: right, done

gaetano-guerriero:
I've applied the comments. 
In order to fix the added test on windows I need to find a windows host so it will take some days.

PR #492, opened 2018-11-15
Created JSONWithValidatorDecoder

tomerghelber:
JSONWithValidatorDecoder is a JSONDecoder that validates our json after decoding it. It gives us encapsulation of Validator and Decoder together.
I used it a couple of times and I thought more people would like to use it.

### DIFF
--- a/json_decoder
+++ b/json_decoder
@@ -1,0 +1,13 @@
+import json
+
+import jsonschema
+
+class JSONWithValidatorDecoder(json.JSONDecoder):
+    def __init__(self, validator, **kwargs):
+        super(JSONWithSchemaDecoder, self).__init__(**kwargs)
+        self.__validator = validator
+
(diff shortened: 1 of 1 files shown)


Julian:
Hi! Thanks for this. This is more the sort of thing that'd go in a higher level library on top of this, of which I've proof-of-concepted [Seep](https://github.com/Julian/Seep). Would definitely not mind a discussion of this there (with tests!) if it's not already in there (I don't remember whether it is, it might be).

Appreciated regardless though!

tomerghelber:
I see that Seep is dead for 5 years, and not even in PIP.
The add of JsonDecoder, which is a built-in, can sit whenever. It's just validating a json "while" creating.
I can add tests (which I am not sure what to test rally, just a case?) and would like to add it here. If I put it in Seep, I want to see Seep in PIP.
Thank.

PR #494, opened 2018-11-19
Docstring + gitignore edits

omarryhan:
- Added some testing/build files to gitignore
- Added docstring for jsonschema.validators.create id_of kwarg

### DIFF
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -179,6 +179,14 @@ def create(
             will be converted to functions and redefined in this object's
             `jsonschema.TypeChecker`.
 
+        id_of:
+        
+            Function that gets $id from a schema.
+            $id is a reserved keyword. It serves as way to:
+
+                1. Declare an identifier for the schema or subschema
+                2. Declaring a base URL against which $ref URLs are resolved
+            
     Returns:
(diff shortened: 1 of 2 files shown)


Julian:
Hi! Thanks. Can you split these changes from each other?

The docstring fix seems good more or less. The gitignore stuff I've generally said belongs in your global gitignore rather than in every project (including this one), which is where it is [for me](https://github.com/Julian/dotfiles/blob/master/.config/git/ignore).

jsonschema/validators.py
```diff
@@ -179,6 +179,14 @@ def create(
             will be converted to functions and redefined in this object's
             `jsonschema.TypeChecker`.
 
+        id_of:
+        
+            Function that gets $id from a schema.
```
Julian: Will leave a more specific comment but the point of this function is to essentially abstract over `$id` (newer drafts) vs `id` (older drafts) vs whatever crazy new way we come up with for future drafts, so I'd rather not mention a specific name of a validator, would be a bit better to just say what the id *concept* is, which is return the ID of the schema (so, this is a function that, given a schema, returns its ID).

Make sense?

PR #501, opened 2018-12-03
Document Python 3.7 support in setup configuration

johnthagen:
This is [tested on Travis](https://travis-ci.org/Julian/jsonschema), so would be good to get into the official package details.


Julian:
Great, thanks!

Issue #477, opened 2018-10-17
Make ValidationError Hashable


adksujan:
cannot log ValidationError as an exception. Making the exceptions hashable would be great to use in standard logging libraries. Any plans to make ValidationError exception hashable before the v3.0 release??

Julian:
Hey yeah this is one of the remaining things I'm thinking about before the release.

The issue with making them hashable is that doing so requires an API change, because they've got mutable attributes on them at the minute, so those would need to become immutable ones.

stuarteberg:
Of course, `ValidationError` and `SchemaError` would be trivially hashable if they didn't define `__eq__`.  Is it necessary for those classes to be comparable?  I tried grepping the `jsonschema` code base for `==` and `!=`, and as far as I can tell, exceptions are never compared.  (But I could definitely be missing something.)

If comparability is not necessary, I vote for simply removing `_Error.__eq__()` and `_Error.__ne__()`.

---

>Making the exceptions hashable would be great to use in standard logging libraries.

Here's another use-case: `pytest` can't handle unhashable exceptions, so any users whose tests might raise a `ValidationError` will currently see an `INTERNALERROR` traceback from `pytest`, crashing the whole test suite.

For example:

```python
# test.py
from jsonschema import validate
def test():
    validate([], {"type": "object"})
```

```
$ pytest test.py

...

INTERNALERROR>   File "/prefix/lib/python3.6/site-packages/_pytest/_code/code.py", line 398, in exconly
INTERNALERROR>     lines = format_exception_only(self.type, self.value)
INTERNALERROR>   File "/prefix/lib/python3.6/traceback.py", line 136, in format_exception_only
INTERNALERROR>     return list(TracebackException(etype, value, None).format_exception_only())
INTERNALERROR>   File "/prefix/lib/python3.6/traceback.py", line 462, in __init__
INTERNALERROR>     _seen.add(exc_value)
INTERNALERROR> TypeError: unhashable type: 'ValidationError'
```

Julian:
It's used in tests, yes, via `assertEqual`.

> Here's another use-case: pytest can't handle unhashable exceptions, so any users whose tests might raise a ValidationError will currently see an INTERNALERROR traceback from pytest, crashing the whole test suite.

Sounds like a `py.test` bug worth filing regardless, though presumably it's just the same underlying one (that py~34-36 broke this, and then fixed it again in 3.7). pytest seems like the sort of thing though that should be trying to fix it even on those though.

[...]

PR #502, opened 2018-12-06
Make exceptions hashable (but not comparable)

stuarteberg:
This PR addresses #477.

Currently, `ValidationError` and `SchemaError` implement `__eq__`, but don't implement `__hash__`, which makes then unhashable.  That prevents them from being used in certain scenarios, notably in a test suite that uses `pytest` [Edit: It fails under python 3.6, but not 3.7].

The easiest way to make them hashable is to simply remove the `__eq__` implementation.  I doubt many users of `jsonschema` rely on the fact that those errors are comparable.

The only problem is jsonschema's own test suite, which *does* rely on `ValidationError.__eq__()`.  But that's easy to change, as I've done in this PR.  Thoughts?

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -55,16 +55,6 @@ class _Error(Exception):
         for error in context:
             error.parent = self
 
-    def __eq__(self, other):
-        if not isinstance(other, self.__class__):
-            return NotImplemented
-        return self._contents() == other._contents()
-
-    def __ne__(self, other):
-        if not isinstance(other, self.__class__):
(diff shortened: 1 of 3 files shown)


Julian:
So, I'd actually like to move in the reverse direction longer term (where users *can* rely on comparisons, which is what they *should* do, a large reason why I added this originally was that there were many internal tests that were having to make many assertions instead of just one single comparison).

But! In the short term, I think this does make sense until we do the deprecations and can therefore roll this out more sanely.

So overall, thanks :), looks reasonable.

PR #511, opened 2019-01-02
Depend on setuptools

### LINKED ISSUE #510: `pkg_resources` used without setuptools dependency
`pkg_resources` is imported inside `__init__.py` yet `setuptools` is not called out as an explicit dependency in `setup.cfg`.  Without `setuptools` this package cannot be imported.

Please reference https://github.com/pypa/pip/issues/6100#issuecomment-450997535 for an error case.

uSpike:
`pkg_resources` requires `setuptools`.  Fixes #510

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ install_requires =
     attrs>=17.4.0
     pyrsistent>=0.14.0
     six>=1.11.0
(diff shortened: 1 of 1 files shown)


Julian:
Great, thanks!

Issue #470, opened 2018-10-02
Release schedule to V3?


adamdavis40208:
https://pypi.org/project/jsonschema/#history

What is the roadmap for moving the stable release to 3? Do you all need any help?

Thanks for all the hard work you put into this!

Julian:
Hi, thanks so much!

I'd say it's probably about a month before a stable release, there's some non-critical-functionality I'd still love to get out before a real release, and want to give some more time for any feedback on the alpha.

Could definitely use a hand on a few things! Definitely if any of the open issues look like the sort of thing you might be interested in I'm more than happy to chat about how to implement them. And there's quite a few open PRs unfortunately, some of which are waiting for a review, which is another area that definitely could use a hand on!

opethe1st:
I would also love to help. Your library is awesome! I was gushing as I was using it :) 👍

Julian:
:) thanks to you both!

If the offer to help is open, #191 and #460 are 2 fairly self contained issues (the former being quite old at this point).

#191 is pretty straightforward on how to approach I suspect -- a test is needed, and then presumably something is calling like `extendleft` when it should be calling `extendright` or something.

And #460 would also be great to start on before a 3.0 release so that the deprecation could be part of the release. The first step on that one would be the always fun task of deprecation, which is still quite hard to do in Python given that we still don't have a great library for it. So ideally someone can propose something for that issue that's fairly small.

Happy to give further pointers of course (though yeah my time comes in bursts at this point).

tmontes:
@Julian,

First and foremost, thanks for sharing jsonschema: I've been exploring it recently, evaluating its possible application in a near future project and I quite like it, in particular the detailed validation failures it produces.

Having said that, I'd like to target draft-06 of the spec which, I understand, is included in the current version 3 alpha, not published on PyPI yet.

Wanting to contribute in moving things forward, I found this issue, which I understand is closed, but seemed like the best place to chime in on... Maybe it could be kept open until v3 is released? Or, alternatively, all pending v3 issues/PRs could be assigned to the [v3.0.0 milestone](https://github.com/Julian/jsonschema/milestone/2), like #133 is? What do you say?

Question: are the above mentioned issues the ones that need work before v3 is deemed releasable? Maybe others? Can you please pinpoint a kind of "roadmap" to v3?

Again thanks.

Julian:
Hi! Great, thanks so much for the offer!

[...]

PR #516, opened 2019-01-11
Docs now build successfully out of the box on macOS CPython.

tmontes:
Exploring possibilities of collaboration, a following up on [this conversation at #470](https://github.com/Julian/jsonschema/issues/470#issuecomment-453289512), I found out that I could not build the documentation on my laptop, running macOS with vanilla installations of CPython 3.6 and 3.7 from www.python.org.

The failure is in the `docs/jsonschema_role.py` Sphix extension, in the `fetch_or_load` function, which uses `urllib.request.urlopen` with an HTTPS URL. Unfortunately, such CPython installs fail to validate server-side TLS certificates.

The motive stems from the fact that, as stated in the installer's README, a private OpenSSL copy is included in the distribution which does not use the system certificate store for validation; instead, the suggested approach is doing a system-wide installation of the `certifi` third-party package and then creating a symlink from a file in that package to a file in the standard library (argh!).

Since I explicitly avoid installing any system-wide Python packages -- and, thus, completing the above mentioned install -- I often need to find ways to make things work. I believe many others experience similar failures due to a) actively understanding the issue and not installing system-wide things (like me), or b) not being aware of the issue and not reading the installer README where a possible solution is documented and suggested (many/most people, I'd venture).

However `urllib.request.urlopen` can be explicitly given a `cafile` argument which, again explicitly in the context of `jsonschema`, can use `certifi`'s CA list. This PR adds `certifi` as a requirement to build docs and adjusts the `urlopen` call such that is uses `certifi`'s CA list.

With this in place, the docs can be build on any platform, regardless of the (unfortunately somewhat sad) state of the underlying Python's TLS certificate validation mechanism / dependencies.

Feedback is welcome, of course.
Thanks.

### DIFF
--- a/docs/jsonschema_role.py
+++ b/docs/jsonschema_role.py
@@ -8,6 +8,7 @@ try:
 except ImportError:
     import urllib.request as urllib
 
+import certifi
 from lxml import html
 
 
@@ -62,7 +63,7 @@ def fetch_or_load(spec_path):
             raise
 
     request = urllib.Request(VALIDATION_SPEC, headers=headers)
-    response = urllib.urlopen(request)
+    response = urllib.urlopen(request, cafile=certifi.where())
 
     if response.code == 200:
         with open(spec_path, "w+b") as spec:
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,4 @@
+certifi
 lxml
 sphinx==1.6.7
 sphinxcontrib-spelling


Julian:
Cool! Thanks, so two comments --

Firstly, purely on the matter of "supporting" CPython as a development platform, I don't use it -- I'm totally happy to "support" it (i.e. guarantee that this and related stuff "works" when developing `jsonschema`), but would love for that to be tested, to ensure it doesn't break, especially given that it's not an environment I'm likely to test under locally. Long story short, I think for this, it'd be nice to add building the docs under CPython to the `tox.ini` if that's a thing that we want to make sure always works.

I *suspect* that making the tox.ini build the docs on multiple interpreters will clutter up this PR a bit, so to me probably perfectly reasonable to do it "right after" this one, but yeah lemme know if that makes sense to you. I *don't* for example necessarily think we need to support building the docs (or development in general) on multiple *versions* of Python (beyond PyPy2.latest, PyPy3.latest, CPython2.latest, Cpython 3.latest).

Overall... Obviously I get queazy around security things :)

Everything you said makes total sense, but I wonder whether it's even safer to just decide to use `requests` there instead? We're anyways adding a doc-dep.

tmontes:
> Firstly, purely on the matter of "supporting" CPython as a development platform, I don't use it -- I'm totally happy to "support" it (i.e. guarantee that this and related stuff "works" when developing `jsonschema`), but would love for that to be tested, to ensure it doesn't break, especially given that it's not an environment I'm likely to test under locally.

Out of curiosity, if you don't develop on CPython, which interpreter(s) do you use, pypy/pypy3? Or do you mean you don't develop on CPython on macOS?

> Long story short, I think for this, it'd be nice to add building the docs under CPython to the `tox.ini` if that's a thing that we want to make sure always works.

That makes sense, yes. For now, like I commented yesterday, the environment I prepared is not using `tox` so I just built the docs manually with `cd docs && pip install -r requirements.txt && make html`.

> I _suspect_ that making the tox.ini build the docs on multiple interpreters will clutter up this PR a bit, so to me probably perfectly reasonable to do it "right after" this one, but yeah lemme know if that makes sense to you. I _don't_ for example necessarily think we need to support building the docs (or development in general) on multiple _versions_ of Python (beyond PyPy2.latest, PyPy3.latest, CPython2.latest, Cpython 3.latest).

Those platforms make sense, yes. Even a 3-only subset would make sense, I'd say, regarding "developing documentation", but maybe that's a stretch...

> Overall... Obviously I get queazy around security things :)
> 
> Everything you said makes total sense, but I wonder whether it's even safer to just decide to use `requests` there instead? We're anyways adding a doc-dep.

I gave that a quick thought and said to myself "`urlopen` is good enough, no need for more dependencies, even if doc-building only". Less is more, I said. :)

tmontes:
All AppVeyor and Travis CI jobs failed... :/

* Most due to the fact that test `jsonschema.tests.test_jsonschema_test_suite.TestDraft7.test_time_validation_of_time_strings_a_valid_time_string` is failing...
* The **pypy** one on Travis [here](https://travis-ci.org/Julian/jsonschema/jobs/478426085)  failed for a different motive.

Does this make any sense to you?

Julian:
I just reverted the time test, it's related to that Format Checker issue I
mentioned in that comment.

On Jan 11, 2019 12:11, "Tiago Montes" <notifications@github.com> wrote:

All AppVeyor and Travis CI jobs failed... :/

   - Most due to the fact that test
   jsonschema.tests.test_jsonschema_test_suite.TestDraft7.test_time_validation_of_time_strings_a_valid_time_string
   is failing...
   - The *pypy* one on Travis here
   <https://travis-ci.org/Julian/jsonschema/jobs/478426085> failed for a
   different motive.

Does this make any sense to you?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/Julian/jsonschema/pull/516#issuecomment-453588716>, or mute
the thread

Julian:
Uh sorry and as for

> The pypy one on Travis here failed for a different motive.

That is https://github.com/pyga/ebb-lint/pull/13 which I now have admin rights to so have merged the PR, but in the process of getting access to PyPI to do a release to fix it. I *hope* that'll happen in the next few days, but if not we can pin `intervaltree<3` to fix it temporarily.

Julian:
> Out of curiosity, if you don't develop on CPython, which interpreter(s) do you use, pypy/pypy3? Or do you mean you don't develop on CPython on macOS?

`pypy` yeah. I haven't particularly used CPython for anything other than "check what CPython does on this code" in quite a number of years now, so while of course it's supported for end-users (who haven't learned yet to use PyPy :) I don't generally deal with it on a day to day basis.

And this is green now with the two fixes, so think that's a good to merge then, though yeah lemme know if you can add the tox.ini things.

PR #532, opened 2019-02-28
Add missing id_of to validators.extend

raphendyr:
Hi,

`id_of` was added to `create` in 3.0.0, but it was forgotten from `extend`.


Julian:
Hi! Thanks very much.

This looks likely to be a bad bug, but it needs tests.

Would you be able to write one?

raphendyr:
I added a minimal test, which can be used to identify the bug and to validate, that the fix at least seems to work.

Julian:
Thanks!

PR #539, opened 2019-03-07
Fix example in FAQ subschema default

dales:


Julian:
Hi! Thanks so much for the patch.

It's actually hard to see though what this fixes.

Would you possibly be able to add a doctest for whatever it is that this does fix?

Julian:
Closing this for cleanliness but happy to re-review after understanding the above!

PR #555, opened 2019-04-25
Fix parsing relative references to other files

drd-ableton:
This sets the $id attribute to the absolute file URI if it is not already set,
so that a base URI is available and the existing code for resolving references
can be used.

### DIFF
--- a/jsonschema/cli.py
+++ b/jsonschema/cli.py
@@ -3,6 +3,8 @@ import argparse
 import json
 import sys
 
+from six.moves import urllib, urllib_parse
+
 from jsonschema._reflect import namedAny
 from jsonschema.validators import validator_for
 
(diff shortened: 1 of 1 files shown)


drd-ableton:
I took the simplest possible approach I could find here, instead of changing a lot of things to be able to pass the original URI along with the documents through all the layers of code.

I'm not sure how to fit this into the existing test suite, though.  The cli tests mock everything, and the standard json-schema test suite (unfortunately, and hopefully just for now) does not contain any tests for this.  Suggestions on how this should be tested welcome.

Julian:
Hi, thanks!

Maybe a good start would be if you could show a reproducing example?

Definitely appreciated!

drd-ableton:
It turns out that this approach doesn't work because it can break "strict" schemas with `additionalProperties: false`, but I ended up writing my own command line script to get around other issues, that uses an appropriately configured `RefResolver` as mentioned in some other issues.

I still think it would be nice if the stock `jsonschema` script supported this correctly, though, so I'll try to come up with a simple example and/or a test and tinker some more with implementing it.

Julian:
Thanks! Yeah I'm happy to merge improvements, the rub so far has been making sure to have simple test cases to merge alongside the improvement.

Julian:
Closing this, though definitely happy to discuss if you can show some concrete examples we can work on!

Issue #499, opened 2018-11-27
Running async procedure in custom validator


rmedaer:
Hi,

I'm using __jsonschema__ for a while... Going deeper in schema validation I figured out that I can define my own validation procedures.[(Thanks Radek Lat for his nice article)](https://lat.sk/2017/03/custom-json-schema-type-validator-format-python/)

So one of the feature I'd like to use is the custom validators. However (in my use case), it would be useful only if I can run asynchronous task.  Furthermore I'm running validation from a coroutine.

So here is my question: How can I execute async calls in a custom validator ? Does the generator can help me ?

Julian:
Think I mentioned via PM that this is a long term goal (to better support async use cases, more so for ref resolution than validation but sure there too). Closing this for now but of course happy to review any proposed changes.

rmedaer:
I think I finally found an elegant way to do mixed async/sync validation. Indeed it needs some changes but few ! Here is my proposal...

The idea would be to declare a new type of `ValidationError` to yield during the validation (for instance `AsyncValidation`) . This new "pseudo validation error" may go through the generator validation tree until a `Validator#async_iter_errors` method where we may check if it `isinstance(error, AsyncValidation)`. If yes, we shall call asynchronously `error.async_validate` which will populate a buffer of errors later yielded by the original validator method. Pay attention that it will **NOT work** out of the box with "conditional checks" (like anyOf, ..) but I'm pretty sure that there is not a lot of changes to do to handle these cases.

If `AsyncValidation` is not supported (for instance if we "only" call `iter_errors`) it will generate a validation error.

Quiet confusing ? Ok, here is an example:

```python
import asyncio
import functools
from jsonschema import ValidationError, Draft7Validator
from jsonschema.validators import extend


class AsyncValidation(ValidationError):

    def __init__(self, callback, value, *args, message='async validation not supported', **kwargs):
        super(AsyncValidation, self).__init__(message, *args, **kwargs)
        self.callback = callback
        self.value = value
        self.errors = []

    async def async_validate(self):
        async for error in self.callback(self.validator, self.value, self.instance, self.schema):
            self.errors.append(error)

class AsyncValidator():

    def __init__(self, draft_src, validators, schema):
        validator = extend(draft_src, validators)(schema)
        self.__class__ = type(validator.__class__.__name__, (self.__class__, validator.__class__), {})
        self.__dict__ = validator.__dict__

    async def async_iter_errors(self, body):
        for error in self.iter_errors(body):
[...]

PR #565, opened 2019-05-23
Feature: Asynchronous validators

rmedaer:
This (WIP) pull-request contains a new feature ! It allows user to write (custom) **asynchronous validators**.
It has been already discussed in #499 . It is here an attempt to integrate the mechanism without having to rewrite the whole library.


Julian:
Hi! Thanks so much for trying to tackle this. I need to spend some time reading through exactly what you're trying to make work, but I can say from a first look that this as is looks like a backwards incompatible change -- so at least, there's that to work through.

E.g., existing validators, that previously only dealt with exceptions, now might be being passed through your new sentinel objects.

Will try to have a closer look though.

rmedaer:
> Thanks so much for trying to tackle this.

You're welcome ! ;-)

> (...) is looks like a backwards incompatible change

I'm not sure there is incompatible change. Here is an explanation of the way it works. I hope it will help you (or anyone) to understand the mechanism and how to use it...

My idea is to use a kind of _generator-based coroutine_. It's basically a way to pause/break code execution using the `yield` keyword. Some frameworks ([like Tornado](https://www.tornadoweb.org/en/stable/gen.html)) had to use this mechanism when `async`/`await` keywords were not available (<= 3.5). This kind of  code interruption is exactly what I'm looking for in the validation process to use with asynchronous validator.

Hopefully in @Julian's jsonschema implementation, the validation is a big generator which iterates recursively on each schema keyword.
Thanks to this implementation, I "only" have to implement [my own _Future/coroutine_ object](https://github.com/Julian/jsonschema/pull/565/files#diff-21bac42ae60c15af7c029a0441204a4fR141) which will be `yield`ed [when we detect an asynchronous validator](https://github.com/Julian/jsonschema/pull/565/files#diff-ed2f83de8c2ff5b1c1138f9f28ba4ba6R334). This object contains the asynchronous validator to call with its arguments.
Obviously we don't want to forward this object to the user as `ValidationError`. [Instead we intercept it in `async_iter_errors`](https://github.com/Julian/jsonschema/pull/565/files#diff-ed2f83de8c2ff5b1c1138f9f28ba4ba6R294) which is running asynchronously the validator in current async context.
Although if user is calling `iter_errors`, this object is **not** intercepted. It has to be handled as a `SchemaError`.
The limitation I detected for now is the "error buffering" (aka not transversal). Meaning somewhere in the  code where the error are not directly forwarded to the user code. I detected and fixed [2 validators](https://github.com/Julian/jsonschema/pull/565/files#diff-66289c46b2d9052f921a55730be5c8bdR311) where I have to forward error if and only if it's a AsyncValidationError.

Here are illustrations of code flow with asynchronous validator compared to synchronous one:

**Synchronous** validator with async_iter_errors in between:
async_validation (1)

**Asynchronous** validator flow:
async_validation (2)

Julian:
Thanks! That diagram is helpful. Apologies for still not getting a chance to look through this more carefully then, but trust me I haven't forgotten it.

rmedaer:
Hi Julian,

Did you have some times to review this merge request ? Is there anything I can do to help you ?

Kind regards,

Julian:
Hi! Yes I did have a chance to review, and so sorry again about the delay
-- will follow up with some comments on it soon -- I think in terms of
priority I wanted to get support for the new draft out first (I know I said
that last time :/) but I definitely haven't forgotten about this.

Julian:
So I'm finally getting at least a moment to even read through this -- first most basic question -- I see you implemented it for anyOf -- is your initial goal specifically for "compound" validators? i.e. anyOf, allOf, oneOf?

And you're trying to get the event loop a chance to run in between each of the branches?

Or is there some other reason anyOf is special here?

rmedaer:
> So I'm finally getting at least a moment to even read through this -- first most basic question -- I see you implemented it for anyOf -- is your initial goal specifically for "compound" validators? i.e. anyOf, allOf, oneOf?

I had to do changes for "compound" validators because of following kind of code branching:

```python
errs = list(validator.descend(instance, subschema, schema_path=index))
if not errs:
    break
```
Indeed, if an `AsyncValidationBreakpoint` is yielded by children (aka descendant) validators, I have first to forward it before testing there is actually an error...

> And you're trying to get the event loop a chance to run in between each of the branches?

It's exactly what I'm doing with the `AsyncValidationBreakpoint` mechanism. When a validator is asynchronous, it raises an exception `AsyncValidationBreakpoint` which [ascends up to async_iter_errors function](https://github.com/Julian/jsonschema/pull/565/files#diff-ed2f83de8c2ff5b1c1138f9f28ba4ba6R298).

I also [implemented it for `oneOf`](https://github.com/Julian/jsonschema/pull/565/files#diff-66289c46b2d9052f921a55730be5c8bdL322).

For `allOf`: not any change is required [since all the errors are just yielded](https://github.com/Julian/jsonschema/blob/fe1c4dd1ed3285f39b3f971834bb04cffde9a3dd/jsonschema/_validators.py#L316)

rmedaer:
Hi @Julian,

Do you need more details for this pull request ? Can I help you in any way ?

Julian:
Hi @rmedaer -- really sorry this keeps dragging, but still have too many things on my plate -- the next short term goal for `jsonschema` is still to get support for the new draft out [2019-09] (which is by now no longer new :/)

I hope after that I can finally think about whether I agree with the approach here or not.

Sorry again... too many things, too little time :(

Julian:
The one other thing that occurs to me immediately though is -- `async_iter_errors` should be fairly easy to implement externally as a normal function to start, no?

If the only long operation you're interrupting here is `anyOf` and `oneOf`, you'll spend longer in those two methods than you might want to, but that may be ok depending on your async app.

Which is part of what makes it not immediately obvious to me that this is the correct approach -- it doesn't actually get us very much async -- e.g., there still wouldn't be an easy way to inject in asyncronous ref resolution.

The only real internal async thing this would do is let event loops run in between iterations of any/oneOf, which doesn't seem like a huge win if you don't have the other stuff over just running the event loop between iterations of `iter_errors`.

(All of which I meant to elaborate on at some point but yeah see above on not enough time :( :( )

Julian:
@rmedaer any thoughts?

rmedaer:
Hi @Julian 

> The one other thing that occurs to me immediately though is -- `async_iter_errors` should be fairly easy to implement externally as a normal function to start, no?

Not sure I fully understand. What do you mean by _"fairly easy to implement externally as a normal function to start"_ ?

Would you mean that we should have a equivalent to `validate` function for `async_iter_errors` ?
Or would you mean that we should "externalize" `async_iter_errors` in another file/project/repo ?

> If the only long operation you're interrupting here is anyOf and oneOf, you'll spend longer in those two methods than you might want to, but that may be ok depending on your async app.

Actually I'm not interrupting `anyOf` and `oneOf` more than other validator. `anyOf` and `oneOf` are just _compound validators_ which needed some refactoring to allow `AsyncValidationBreakpoint` interruptions.

> (...) it doesn't actually get us very much async -- e.g., there still wouldn't be an easy way to inject in asyncronous ref resolution.

I can implement asynchronous ref resolution (and others) if needed. This PR is a first step in async validation. Btw, if we decide to go to async (and lazy-loaded) ref resolution, we will need this PR anyway.

> The only real internal async thing this would do is let event loops run in between iterations of any/oneOf

Apologize if I misspoke. With this PR you can use async "anywhere" with "any validator". **Not only** between iterations of any/oneOf.

For instance:

```python
async def async_validator(validator, value, instance, schema):
    await asyncio.sleep(0)
    if not value:
        yield exceptions.ValidationError(u"Async whoops!")

all_validators = dict(validators.Draft7Validator.VALIDATORS)
all_validators['async_valid'] = async_validator
validator = validators.create(
    meta_schema=validators.Draft7Validator.META_SCHEMA,
    validators=all_validators,
)({})

instance = {}
schema = {
    {
        "type": "object",
        "properties": {
            "example": {
                u"async_valid": True
            }
        }
    },
}
errors = [e async for e in validator.async_iter_errors(instance, schema)]
```

rmedaer:
Hi @Julian ,

I try to fix the tests, however in list of interpreters there is still `py2`. Is it a mistake ? Otherwise how would you like to organize tests to run some of them only from minimum python version (3.6 in my case) ?

Julian:
@rmedaer will try to make some time in the next week or two to give this another pass -- thanks truly again for being on top of it.

On the py2 thing specifically, `jsonschema` still supports Py2 (at least versions of Py2 that are still supported, i.e. PyPy) -- this will change at some point but no timeline yet. From what I saw I think what you did looked reasonable, but will have a look.

rmedaer:
Hi @Julian ,

> will try to make some time in the next week or two to give this another pass

Did you had some time to a new pass ?

Julian:
Hey @rmedaer thanks for being so on top of this... I did, but honestly it hasn't yet sat fully well with me as-is given that it handles only the validator side and not the ref resolution side. I still don't really have a full idea of what to do about that yet. Would love to see comments from any other users certainly who want this or otherwise have an opinion.

PR #580, opened 2019-07-11
Add --version argument

### LINKED ISSUE #578: Make jsonschema --version work

vladmunteanu:
Closes #578 

[PY2 argparse action](https://docs.python.org/2/library/argparse.html#action)
[PY3 argparse action](https://docs.python.org/3/library/argparse.html#action)

### DIFF
--- a/jsonschema/cli.py
+++ b/jsonschema/cli.py
@@ -3,6 +3,7 @@ import argparse
 import json
 import sys
 
+from jsonschema import __version__
 from jsonschema._reflect import namedAny
 from jsonschema.validators import validator_for
 
@@ -48,6 +49,11 @@ parser.add_argument(
         "validators that are registered with jsonschema, simply the name "
         "of the class."
     ),
+)
+parser.add_argument(
+    "--version",
+    action='version',
+    version=__version__,
 )
 parser.add_argument(
     "schema",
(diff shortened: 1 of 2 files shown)


Julian:
Hi! Thanks so much.

Do you think you could give a shot at adding a simple test to the test suite?

Something simple like just making sure that `__version__ in check_output(...)` is probably good enough?

vladmunteanu:
Hello @Julian,

I have tried adding the test, verifying the full version string after stripping newlines.

I can't run the test suite locally because I can't get pyenv and tox to play nice with each other.
Besides that, pypy seems to fail and I can't figure out why at the moment.

Julian:
Thanks! Should just be that we need to make sure to run the right `jsonschema` CLI depending on the running Python, so just tacked a commit on top to do that.

Curious to hear more though about

> I can't run the test suite locally because I can't get pyenv and tox to play nice with each other.

if you think the issue might not just be yours.

Otherwise thanks again!

vladmunteanu:
Thanks @Julian! And sorry I didn't figure out the stderr thing myself.

The problem I had with pyenv and tox has been fixed by running `pyenv global <versions>` before running tests, not sure if this is actually the recommended way.

PR #591, opened 2019-07-30
Adding falsy value evaluation to Unset

taliamax:
Causing Unset to have falsy behaviour can simplify tests and other checks.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -47,7 +47,6 @@ class Unset(object):
     def __repr__(self):
         return "<unset>"
 
-    
     def __nonzero__(self):
         return False
 
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -47,6 +47,10 @@ class Unset(object):
     def __repr__(self):
         return "<unset>"
 
+    
+    def __nonzero__(self):
+        return False
+
 
 def load_schema(name):
     """


Julian:
Hi! Thanks for the patch.

Happy to consider this, but do you think you could demonstrate some of the refactoring you're mentioning there?

Appreciated!

taliamax:
@Julian 

One of the biggest things is that now you can test for the existence of anything that could be a default value of unset or something else. For example, if I'm not sure if my schema value will return something that isn't a default, I can do this:

```python
if test_value.schema and some_condition(test_value.schema):
    pass
```

Whereas before to check for Unset you'd have to do something along the lines of
```python
if not isinstance(test_value.schema, jsonschema._util.Unset) and some_condition(test_value.schema):
    pass
```

Julian:
Unset is a private object, so no end-user code should really be interacting
with it.

I thought you meant refactoring within jsonschema itself.

Did I misunderstand?

taliamax:
Unless I misunderstand, you can interact with it when exceptions occur, since it's the default value for (at the very least) ValidationError: https://python-jsonschema.readthedocs.io/en/stable/errors/

Julian:
Do you have an example of that happening?

It's an internal default that as far as I know should never leak, it's
always replaced by the time someone who's an end user sees a ValidationError

Julian:
Going to close this, because yeah I hope that this shouldn't be happening -- but if you do see it, definitely happy to talk further.

Appreciated though!

PR #594, opened 2019-08-09
Handle DistributionNotFound

trollfred:


Julian:
Hi. Thanks. What situation are you encountering that you're trying to fix here?

trollfred:
hi @Julian! I'm packing it as a dependency with [py2app](https://pypi.org/project/py2app/) and pkg_resources doesn't work there, so I thought some handling won't hurt, right?

Julian:
So, understandable, but py2app isn't currently supported because it doesn't run in CI.

I'm perfectly happy to support it if it doesn't cause major issues, but it'd need to run in CI -- i.e., if you say added a tox environment that tries to build jsonschema, and that failed, and passed with this change, that's fine (even though py2app is doing something wrong there), but yeah as-is without coverage it's unfortunately not a supported configuration because I have no way of knowing whether some other future otherwise-totally correct change will break py2app.

trollfred:
I still see two possibilities:
1. I could try adding a test case for py2app, gonna be complicated but ok
2. I could detect py2app and reraise DistributionNotFound if wasn't detected - this way this change will be 100% transparent for the rest of the world
what do you think?

Julian:
It's not about backwards compatibility -- all code changes have to have
tests

The test should essentially do whatever exact procedure someone would go
through to run py2app

Julian:
I don't know that it fixes the issue here (and again, much as I sympathize, I think "officially", this is still a not-really-supported configuration unless it's under CI), but you might find that the next release (3.1) *happens* to fix your issue here (because we've now moved to importlib-metadata instead of pkg_resources, which is the newer way to do what was done here before).

I don't know for sure that it fixes this, but it may.

PR #531, opened 2019-02-28
Added support for 'file://' URL schema on the RefResolver

USSX-Hares:
Fixes Julian/jsonschema#398

### Difference
Before this fix, RefResolver created with `base_uri` starting with `file://` could work not properly.

#### Example code to test:
```python
import os.path
import json

from jsonschema import validate, RefResolver

contact = \
{
    "name": "William Johns",
    "age": 25,
    "birthDate": { "month": "apr", "day": 15 }
}

def main():
    schema_path = "schemas/main_schema.json"
    schema = json.load(open(schema_path))
    validate(contact, schema, resolver=RefResolver(base_uri='file://' + os.path.dirname(os.path.abspath(schema_path)), referrer=schema))

if (__name__ == '__main__'):
    main()
```

#### Work tree:
```
    /
    +-- code.py
    +-- schemas /
    |   +-- dependencies /
    |   |   +-- date.json
    |   |   +-- month.json
    |   |   +-- positive_integer.json
    |   +-- main_schema.json
```

#### $ref Dependencies:
```
main_schema => date, positive_integer
date => month, positive_integer
```

### Tests
Tests passed on:
 - [x] Windows 7, Python 2.7.15
 - [x] Windows 7, Python 3.5.2
 - [x] Windows 7, Python 3.7.2

Would also test on:
 - [ ] Linux (Ubuntu 16.04), Python 2.7.
 - [ ] Linux (Ubuntu 16.04), Python 3.5.2
 - [ ] Linux (Ubuntu 16.04), Python 3.6.7
 - [ ] Linux (Ubuntu 16.04), Python 3.7.2

PR #595, opened 2019-08-16
Eterna2/fix relative ref

### LINKED ISSUE #570: "RefResolutionError" exception raised when resolving "$ref"
Taking the example from the json-schema documentation on structuring complex schema: [https://json-schema.org/understanding-json-schema/structuring.html](url).

I create a file **address_pass.json** where address is defined internally:

```
{

  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "street_address": { "type": "string" },
        "city":           { "type": "string" },
        "state":          { "type": "string" }
      },
      "required": ["street_address", "city", "state"]
    }
  },

  "type": "object",

  "properties": {
    "billing_address": { "$ref": "#/definitions/address" },
    "shipping_address": { "$ref": "#/definitions/address" }
  }
}
```
The code below illustrates this validates the example instance:
```python
import json

instance = {
  "shipping_address": {
    "street_address": "1600 Pennsylvania Avenue NW",
    "city": "Washington",
    "state": "DC"
  },
  "billing_address": {
    "street_address": "1st Street SE",
    "city": "Washington",
    "state": "DC"
  }
}

with open('address_pass.json') as json_data:
    schema = json.load(json_data)
    json_data.close()
    
validate(instance=instance, schema=schema)
```
All good. Now, the example does not define the content of **definitions.json** but I believe it should look like this:
```
{
  "address": {
    "type": "object",
    "properties": {
      "street_address": { "type": "string" },
      "city":           { "type": "string" },
      "state":          { "type": "string" }
    },
    "required": ["street_address", "city", "state"]
  }
}
```
With corresponding reference from within **address_fail.json** as below:
```
{

  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "street_address": { "type": "string" },
        "city":           { "type": "string" },
        "state":          { "type": "string" }
      },
      "required": ["street_address", "city", "state"]
    }
  },

  "type": "object",

  "properties": {
    "billing_address": { "$ref": "definitions.json#/address" },
    "shipping_address": { "$ref": "definitions.json#/address" }
  }
}
```
However, the same code as above now fails with the RefResolutionError reported above:
```python
import json

instance = {
  "shipping_address": {
    "street_address": "1600 Pennsylvania Avenue NW",
    "city": "Washington",
    "state": "DC"
  },
  "billing_address": {
    "street_address": "1st Street SE",
    "city": "Washington",
    "state": "DC"
  }
}

with open('address_fail.json') as json_data:
    schema = json.load(json_data)
    json_data.close()
    
validate(instance=instance, schema=schema)
```
Furthermore, replacing ```$ref``` with ```file``` and a full path to **definitions.json**  illustrates the external reference does work. So the issue seems to be simply in locating the file.

eterna2:
Fix #570 
- Support relative local fs ref
- Support relative remote ref
- Added corresponding unittests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,9 @@ _cache
 _static
 _templates
 
+.eggs
+.tox
+
+*.py[cod]
+
 TODO
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1669,7 +1669,7 @@ class TestRefResolver(SynchronousTestCase):
             self.addCleanup(os.remove, tempf.name)
             json.dump({"foo": "bar"}, tempf)
 
-        ref = "{}#foo".format(pathname2url(tempf.name))
+        ref = "{}#foo".format(os.path.relpath(tempf.name))
         with self.resolver.resolving(ref) as resolved:
             self.assertEqual(resolved, "bar")
 
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1664,7 +1664,7 @@ class TestRefResolver(SynchronousTestCase):
         with self.resolver.resolving(ref) as resolved:
             self.assertEqual(resolved, "bar")
 
-    def test_it_retrieves_local_relative_refs_via_urlopen(self):
+    def test_it_retrieves_local_relative_refs_via_open(self):
         with tempfile.NamedTemporaryFile(delete=False, mode="wt") as tempf:
             self.addCleanup(os.remove, tempf.name)
             json.dump({"foo": "bar"}, tempf)
@@ -1672,6 +1672,30 @@ class TestRefResolver(SynchronousTestCase):
         ref = "{}#foo".format(pathname2url(tempf.name))
         with self.resolver.resolving(ref) as resolved:
             self.assertEqual(resolved, "bar")
+
+    def test_it_retrieves_relative_refs_via_urlopen(self):
+        ref = "fuzz.json#baz"
+        schema = {"baz": 12}
+        resolver = validators.RefResolver(
+            "http://foo.com/bar.json", self.referrer, self.store,
+        )
+        if "requests" in sys.modules:
+            self.addCleanup(
+                sys.modules.__setitem__, "requests", sys.modules["requests"],
+            )
+        sys.modules["requests"] = None
+
+        @contextmanager
+        def fake_urlopen(url):
+            self.assertEqual(url, "http://foo.com/fuzz.json")
+            yield BytesIO(json.dumps(schema).encode("utf8"))
+
+        self.addCleanup(setattr, validators, "urlopen", validators.urlopen)
+        validators.urlopen = fake_urlopen
+
+        with resolver.resolving(ref) as resolved:
+            pass
+        self.assertEqual(resolved, 12)           
 
     def test_it_can_construct_a_base_uri_from_a_schema(self):
         schema = {"id": "foo"}
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1695,7 +1695,7 @@ class TestRefResolver(SynchronousTestCase):
 
         with resolver.resolving(ref) as resolved:
             pass
-        self.assertEqual(resolved, 12)           
+        self.assertEqual(resolved, 12)
 
     def test_it_can_construct_a_base_uri_from_a_schema(self):
         schema = {"id": "foo"}
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1664,6 +1664,15 @@ class TestRefResolver(SynchronousTestCase):
         with self.resolver.resolving(ref) as resolved:
             self.assertEqual(resolved, "bar")
 
+    def test_it_retrieves_local_relative_refs_via_urlopen(self):
+        with tempfile.NamedTemporaryFile(delete=False, mode="wt") as tempf:
+            self.addCleanup(os.remove, tempf.name)
+            json.dump({"foo": "bar"}, tempf)
+
+        ref = "{}#foo".format(pathname2url(tempf.name))
+        with self.resolver.resolving(ref) as resolved:
+            self.assertEqual(resolved, "bar")
+
     def test_it_can_construct_a_base_uri_from_a_schema(self):
         schema = {"id": "foo"}
         resolver = validators.RefResolver.from_schema(
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -4,6 +4,7 @@ from warnings import warn
 import contextlib
 import json
 import numbers
+import os
 
 from six import add_metaclass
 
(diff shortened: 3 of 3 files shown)


eterna2:
can't figure out why coverage for diff is not 100%. My 2nd unit test shld have covered it.

Julian:
Thanks a lot for trying to tackle this!

Does this work properly with requests?

If so, I'm really more inclined at this point to just drop support for urllib entirely and make requests fully required -- the conditional dependency is... like 7 years old at this point, back when it wasn't necessarily the case that requests was a good idea to depend on.

eterna2:
@Julian 

Actually, my implementation does not use request (but there is no issue using request either). 

It is more of a fallback logic (cuz I do not want to change too much of ur code), it will only attempt to try:
- resolve as a local relative path (concat `base_uri` with `ref_uri` and try `json.load`).
- otherwise resolve as remote relative path (strip spec filename to get base url and resolve path with relative `ref_uri` with respect to the base url (I use urllib/urlopen here in this case) - but i still need urlsplit and urlunsplit to interpret the url so that I can find the base_url. But frankly it is not difficult to just implement that part (instead of using urllib).

```py
 # need to reimplement this if u r deprecating urllib
scheme, network, path, qs, frag = urlsplit(self.base_uri) 

path = "/".join(path.split("/")[0:-1])  # base path

 # need to reimplement this if u r deprecating urllib
base_url = urlunsplit((scheme, network, path, qs, frag)) 

# this part can be easily replaced with request instead
with urlopen(urljoin(base_url, uri)) as url:
  result = json.loads(url.read().decode("utf-8"))
```

I made this PR with minimal change to ur codebase. 


Ideally, if u are going to deprecate `urllib` we simplify the RefResolver alot by only using ur existing custom_handlers hook - i.e. we can write a few default handlers (http/https, file, relative_file, relative_remote) and has a fallback instead of a dict (i.e. a list instead of dict only)

This is not working code, just my 2cents, if u ever going to do any major refractoring.

```py
handlers_fallback = [std_handler, relative_file_handler, relative_remote_handler]

def get_std_handlers(uri, custom_handlers={}):
  if uri[0] == "#":
    return local_ref_handler
  if uri[0:7] == "http://" or uri[0:8] == "https://":
    return request_handler

   return custom_handlers.get(schema, file_handler)

def request_handler(uri, base_uri=""):
  return request.get(uri).json()

def std_handler(uri, base_uri=""):
  handler = get_std_handlers(uri)
  try:
    if handler:
      return handler(uri, base_uri=base_uri)
  except Exception:
    return file_handler(uri, base_uri=base_uri)

def local_ref_handler(uri, base_uri=""):
  return definitions[_get_definition_key(uri)]

def file_handler(uri, base_uri=""):
  file_path = strip_protocol_if_any(uri)
  with open(file_path) as fin:  
    return json.load(fin)
 
def relative_file_handler(uri, base_uri=""):
  ref_path = os.path.join(strip_protocol_if_any(base_uri), strip_protocol_if_any(uri))
  return file_handler(ref_path)

def relative_remote_handler(uri, base_uri=""):
  base_url = get_base_url(base_uri)
  url = "%s/%s" % (base_url, strip_protocol_if_any(uri))
  return request_handler(url)
  
def handle_ref(handlers, handler_kwargs, index=0):
  if index >= len(handlers):
    raise Exception("Unable to resolve ref")
  try:
    return handlers[index](**handler_kwargs)
  except Exception:
    return handle_ref(handlers, handler_kwargs, index=index+1)

```

Julian:
> Actually, my implementation does not use request (but there is no issue using request either).

Sorry so are you saying then that the issue doesn't exist when requests is installed?

I'm not sure I follow the rest of your comment, since:

> Just my 2 cent, if u are going to deprecate urllib and do some major refractoring, can make use of the existing custom_handler hook - i.e. create a http/https handler, file handler.

That wouldn't be any major refactoring, it'd literally just delete the entire `else` block you modified here, but otherwise everything stays exactly the same -- so basically what I'm asking is "in that case, is whatever bug you're trying to fix still present"?

Julian:
I should say I'm separately not very keen to do any kind of fallback logic / chaining -- it's easy to tell what's a remote ref and what's a local one, so we shouldn't really need any try/excepting or if/elsing, so will need to try to understand what the motivation is there.

Julian:
This also looks pretty similar to #531, where I seem to have asked the exact same question :)

eterna2:
@Julian 

This is different from #531. #531 is asking to support `file://`.  #570 is asking to support relative reference.

The use case is mostly when u do not want to publish ur schema (i.e. http/https) and only wish to retrieve it from local fs. `urllib` supports `file://` but this is not ideal, as it requires absolute path which is not desirable as different machine will have different paths.

So the feature request is:

if `$ref = "some_schema.json#some_definition"`, `RefResolver` should load from local fs relative to where the original schema (i.e. `$schema`) resides in the fs. If `$schema` is in some remote endpoint, it should similar load relative from that remote endpoint.

The current implementation does not support this:
- request only target http/https (no fs or file:// at all)

And  there are multiple ways to go about doing this:
- u can check `$ref` to see if http/https is specified and use request.
- unless u r dropping `file://` support (it is currently supported because of `urlopen`), you will need to do a check for this too, and implement an equivalent without using `urlopen`.
- then u need to do a check if the original json schema is http/https or a local fs, then implement a local relative `json.load`, or a `request.get(base_url+url).json()` respectively. 

or u can just do a fallack, which i did in this PR. The relative ref will never hit ur request `if` statement because it is relative ref (e.g. `some_spec.json#some_def` instead of `http://foo.bar/some_spec.json#some_def`).

Julian:
Got it thanks -- that makes sense, but is this really that hard to do currently? If it's really just for development, just invent your own scheme, and include a handler for it?

I.e.:

```
import json
from jsonschema import RefResolver, validate


def relative(url):
    _, _, path = url.partition("://")
    with open(path) as file:
        return json.load(file)


with open("schema.json") as file:
    schema = json.load(file)
    resolver = RefResolver.from_schema(schema, handlers={"relative-file": relative})
    validate("", schema, resolver=resolver)
```

eterna2:
This is a PR. U can see my fix here. It isn't hard, and it is officially in the spec (relative ref I mean).

Your approach is doable but not elegant, as in it feels hackish. Cuz the main use case here is validation as part of our unit test/ci process. Would be hackish if we write a custom handler with a custom protocol for all our projects just to support relative ref.

Regardless of urllib support of otherwise, the fix is relatively simple, it is just falling back to local relative fs -> remote relative fs. All about path manipulation, not much on io.

Julian:
The behavior in the PR is definitely not in the spec, so I'm trying to narrow down exactly what/which issue *is* in the spec that you're trying to fix here.

The spec definitely does not say "every time you see a ref, try cascading through first resolving it absolutely, then assume it's a local path, ..."

It has rules about what you assume the base URI is for schemas that forgot to declare it, so fully believe there's something to fix there, but yeah that's what I was referring to above about not having fallback behavior.

Julian:
The other thing I meant to mention here is whether you're really asking about loading schemas that don't specify their base URI and then not telling `jsonschema` what that base URI is that you loaded them from. E.g.:

```
import os
import json
from jsonschema import RefResolver, validate


with open("schema.json") as file:
    schema = json.load(file)
    resolver = RefResolver(
        referrer=schema,
        base_uri="file://" + os.getcwd() + "/",
    )
    validate("", schema, resolver=resolver)
```

if you want the base URI to resolve relative to the current working directory, or some other one.
```

eterna2:
Yeah this is what I meant and needed. Thanks!

Will this work recursively? As in this schema loads another schema which loads another schema? All the schemas are in the same folder but does not have a $id defined.

Because my usecase is a pretty complex set of config files that I need to validate for multiple projects.

Julian:
Yes it will, that's a global base URI to apply.

PR #609, opened 2019-10-04
Use JS regex syntax

### LINKED ISSUE #607: Support for ECMA 262 regex patterns
The value of [the `pattern`] keyword [SHOULD be a valid regular expression](https://json-schema.org/latest/json-schema-validation.html#pattern), according to the ECMA 262 regular expression dialect.  However [sticking to a common subset of the syntax is recommended](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html).

I spent the weekend doing some research on the differences between Python and JS regex, and seeing how difficult it might be to patch over them.  In short: https://pypi.org/project/js-regex/ now exists. (🎉)

It currently supports Python 3.6+ only, but solely because that's a much nicer way to develop a new project - if you were willing to use it in `jsonschema` I'd be happy to make a back-compatible release to support Pythons 2.7 and 3.5.

Zac-HD:
Note that `js_regex.compile` has a cache, just like `re.compile`, so calling it in loops does a lot of dict lookups but very little work.

Closes #607.


jsonschema/_utils.py
```diff
@@ -92,10 +93,10 @@ def find_additional_properties(instance, schema):
     """
 
     properties = schema.get("properties", {})
-    patterns = "|".join(schema.get("patternProperties", {}))
+    patterns = "|".join(sorted(schema.get("patternProperties", {})))
```
Zac-HD: Sorting ensures that this is always a single entry in the regex cache, which is quite important for performance (of this tiny function, anyway).

Julian:
Hooray, thanks!

Can you also enable the tests for this in the test suite? Should basically just be adding a line like this: https://github.com/Julian/jsonschema/blob/master/jsonschema/tests/test_jsonschema_test_suite.py#L71 to each of the drafts.

Zac-HD:
Done!  I've left out Draft3 because it doesn't have the relevant test cases upstream, which I think is json-schema-org/JSON-Schema-Test-Suite#274.  

But drafts 4, 6, and 7 should all be validated... and there's a couple of failures.  Welp.   ...I think I'll need to take a little longer to debug that, and ensure that I got the everything right upstream :disappointed:

Zac-HD:
@Julian - I think this just need to be run on top of the *fixed* version of my upstream tests, and we'll be good to go :smile:

Julian:
Think we're getting there!

I think there's still one that looks like it's failing: https://travis-ci.org/Julian/jsonschema/jobs/594558002#L3575

Zac-HD:
Aaaah, I see what's happening there: this *isn't* one of the "does it match" tests I added - it's actually for `{type: string, format: regex}`.  Compiling with `js_regex` instead of `re` should do the trick :smile:

Zac-HD:
And done! :sparkles::tada::rocket:

Julian:
Awesome, well done man, really appreciated.

Zac-HD:
@Julian - any advice on when this will make it to PyPI?  I've been asked to support it in `hypothesis-jsonschema`, but doing so before `jsonschema` itself will cause lots of test failures... 

If it's going to be a while I can write a shim, but I'd feel pretty silly if I did that and you updated the day after!

Julian:
Probably today :)

Zac-HD:
Awesome!  I'll aim to do my integration tomorrow then :grin:

Julian:
(Done) :)

Issue #247, opened 2015-09-08
multipleOf fails validation (decimals)


harmv:
I noticed that json schemas with `multipleOf` fail to validate when using decimals.

1.1 is accepted as multiple of 0.1, but 1.9 is not.

### example program that demonstrates the problem
```python
    from jsonschema import validate, __version__
    import json

    schema = json.loads("""
    {
        "$schema": "http://json-schema.org/draft-04/schema#",
        "type": "object",
        "properties": {
            "foo": {
                "type": "number",
                "multipleOf" : 0.1
            }
        },
        "additionalProperties": false,
        "required": [
            "foo"
        ]
    }
    """)

    example1 = json.loads("""
    { "foo": 1.1}
    """)

    example2 = json.loads("""
    { "foo": 1.9}
    """)


    def main():
        print "jsonschema.__version__ = ", __version__
        validate(example1, schema)  # this one works
        validate(example2, schema)  # this should work, but doesn't

    if __name__ == "__main__":
        main()
```

### result

    jsonschema.exceptions.ValidationError: 1.9 is not a multiple of 0.1

### expected result

validation passes

### version
    jsonschema.__version__ =  2.5.1

Julian:
Hi! Thanks.

I can read this more carefully tomorrow, but from quickly glancing at that, you don't have any decimals there, you have floats. Specifically you have `1.9`, which most likely is not in fact divisible by `0.1` because I suspect it's not representable exactly in binary. If you're unfamiliar and my guess is correct, you'd want to familiarize yourself with floating point (im)precision, and possibly to tell `json.loads` to deserialize your JSON `number`s as `decimal.Decimal`s or the like instead, if you actually want fixed or arbitrary precision.

Apologies if my guess is wrong, just let me know if you're aware of the above.

harmv:
I understand In json there is no decimal, just integers and floats. the json schema above uses 'multipleOf' as a way to restrict the 'number'  to 1 digit decimals.

Although it might be possible to solve this with `parse_float=decimal.Decimal` in `json.loads` (its not, see below)  one could argue that jsonschema should handle this somehow. (I did not think this through, so I'm not sure this is what jsonschema should do)

What would you expect jsonschema to do here ? Handle this transparantly or not ? In my case the python code doesn't mind that the actual number is a float, thats perfectly fine. I just want to be sure that the json validates against the schema.  

[...]

PR #616, opened 2019-10-14
fix 10001.12 is not a multiple of 0.01

weaming:


Julian:
Hi -- have a look at the failing CI, this change isn't behaviorally correct unfortunately.

There are always going to be float precision errors there -- at the minute we essentially pick a tolerance IIRC, but yeah this change would break lots of other unintended things.

weaming:
@Julian you can test `10001.12` with `0.01`, and will found it does't work. 
Maybe jsonschema could compatible with `Decimal` in #247 , and we can parse json with `parse_float=Decimal`

Julian:
jsonschema already supports validating Decimals if you parse your JSON into
them.

PR #619, opened 2019-10-23
non GPL format option

naimetti:
The  **format** extra includes two libs ([rfc3987](https://github.com/dgerber/rfc3987) and [strict_rfc3339](https://github.com/danielrichman/strict-rfc3339)) that are distributed under **GPL** licence, which renders jsonschema package to GPL, althought it is licensed under an MIT-like license.

To get rid of these two GPL libs dependencies I have implemented the *uri*, *uri-reference*, and *date-time* validators built-in in the lib, and added the **format_nongpl** extra to add format validators dependencies without *rfc3987* and *strict-rfc3339*.

An older pull request (#261) suggest using [rfc3986](https://pypi.org/project/rfc3986/) to replace *rfc3987*, but I have found the RFC3986 validation poorly implemented in that lib.

I have taken the idea to add the new extra in order to maintain backward compatibility from this [comment](https://github.com/Julian/jsonschema/pull/261#discussion_r45548385)

Docs are still pending upon discussion about this pull request.

Note: the  *iri* and *iri-reference* validators provided by *rfc3987* are not implemented in this pull request.


Julian:
Hi, thanks a lot for trying to tackle this!

I'm definitely still open to the idea of having non-GPL implementations available.

On the other hand though I don't want to maintain those here in this library -- (I actually don't want to maintain any format related code :) GPL or otherwise).

Meaning -- if you want to create libraries that essentially contain each of these pieces of functionality, *and* promise to maintain them (or obviously find existing such libraries :) then I'm happy to depend on those, but yeah I can't be maintaining huge monstrous regexes here within this library.

Hopefully that makes sense enough, but happy to hear your thoughts.

naimetti:
Hi, Julian! Thank you for your prompt response.

It makes sense to me. I will create two separate libraries (*validate-rfc3339* and *validate-rfc3986*) and refactor the PR code to use both libraries. In fact, when I was developing the validators I created them as two different projects with their UT and all, so I think it will be the best approach.

naimetti:
If you have any other suggestions I am eager to hear it.

naimetti:
Hi again, @Julian ! 
I have created two separate modules ([rfc3339-validator](https://github.com/naimetti/rfc3339-validator) and [rfc3986-validator](https://github.com/naimetti/rfc3986-validator)) to implement the validations and refactored the code to use them if these libs are present instead of *strict_rfc3339* and *rfc3987*.
In the unit test of both repos I've have include code to compare the behavior of the new code with the GPL versions libs used by jsconschema.

jsonschema/_format.py
```diff
@@ -248,7 +248,28 @@ def is_idn_host_name(instance):
...
     import rfc3987
 except ImportError:
-    pass
+
+    try:
+        from rfc3986_validator import validate_rfc3986
+    except ImportError:
+        pass
+    else:
+        @_checks_drafts(name="uri",)
```
Julian: Can you remove the stray comma at the end here

jsonschema/_format.py
```diff
@@ -248,7 +248,28 @@ def is_idn_host_name(instance):
 try:
     import rfc3987
 except ImportError:
-    pass
+
```
Julian: Can you remove the extra newline here

naimetti:
> I assume you're willing to also maintain those going forward hopefully right (i.e. triage issues)? (With no hard commitment, it's open source hooray...)

Yes, I am willing to also maintain these libs. No problem with that at all. 

> Also -- did you happen to check all the transitive dependencies (of everything) to ensure the non-GPL one really doesn't pull in any non-GPL code beyond just the direct format dependencies?

I've checked and the only two dependencies with MIT incompatible licenses are *rfc3987* and *strict_rfc3339*.

I have also solved the style issues and added some words about these changes in the docs.

naimetti:
About the coverage reports, I don't know exactly what could we do to test both *format* and *format-nongpl* at the same time. Maybe using some environment variables to set the lib preference for the checkers could do the trick to run the unit tests for the gpl and the non-gpl code together.

Julian:
Thanks! Awesome, think this probably is ready to merge, gonna take one last look first.

The coverage thing -- yeah that's something that's "broken" even before this, e.g. we should really be combining coverage across Py2+Py3 too to get the right unioned number, so I don't think that needs to hold this change specifically back, you're right, really we should union the format+formatNonGPL extras probably too.

Julian:
Merged! Thanks again.

naimetti:
Glad to help!
When I have some time I'll try to add support for rfc3987 too.
Thank you!

nicola-lunghi:
@naimetti thanks this is really useful

to have this is enough to include the patches:
- 10f8a3e Add format validators as separate modules
- af37707 non GPL format option
???
There's still something missing?
@Julian are you going to do a new release soon to include those changes ? -> I am preaparing to update a recipe in yocto to add this so I can wait for a release if its not too far away)

Julian:
Probably in the next 4-5 days yeah.

naimetti:
> to have this is enough to include the patches:
> 
> * [10f8a3e](https://github.com/Julian/jsonschema/commit/10f8a3e525f63f10555aeccd06e5d978e8a6958f) Add format validators as separate modules
> * [af37707](https://github.com/Julian/jsonschema/commit/af37707780faf8793ba0734929d7d96f0d31ebee) non GPL format option
>   ???

I think these two commits are enought.

PR #634, opened 2019-12-05
Avoid setup_requires offline installation issues

GermanoGuerrini:
Currently, it's not possible to install jsonschema offline, as pip cannot found setup_requires dependencies.
See https://www.scivision.dev/setuptools-pyproject-toml-offline/ and https://stackoverflow.com/questions/27307082/install-package-which-has-setup-requires-from-local-source-distributions.
Simply removing the option from setup.cfg do the trick.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,6 @@ classifiers =
 
 [options]
 packages = find:
-setup_requires = setuptools_scm
 install_requires =
     attrs>=17.4.0
     functools32;python_version<'3'


Julian:
Thanks.

Looks like CI is failing (on the one tox environment that does rely on setuptools). Moving it to `twine check` might be the easiest fix.

Julian:
@GermanoGuerrini do you think you'd be interested in giving a shot to the README check fixing?

PR #641, opened 2019-12-07
Update meta schema files in project

christian-weiss:
Official meta schema files has been changed since they got committed to the project.
This PR is to bring them in sync with official meta schema files. This PR enables us to do easy diffs in the future.

Official:
http://json-schema.org/draft-03/schema#
http://json-schema.org/draft-04/schema#
http://json-schema.org/draft-06/schema#
http://json-schema.org/draft-07/schema#

Good news is: draft-7 and draft-6 are pretty close to official files.

Files in ./jsonschema/schema/ has been reviewed:
a) all elements are re-sorted to be in sync with element ordering of official schema files (to enable easy PR review)
b) formatting is now in sync with official schema
c) git commit history is inspected to find intended changed
d) patches of the jsonschema project are kept intact
e) rest is updated to become in sync with official meta schema files

For easyer PR review i created separate commits on each step - please review this PR commit-by-commit.

Please execute your test before merge to master to ensure that everything works as expected. You may want to check if your "fail rate" (https://github.com/ebdrup/json-schema-benchmark/blob/master/reports/jsonschema.md) has changed.
If this PR undo an intended patch (now stated in the commit history), let me know, i will re-apply it / update the PR.

BTW: i have to deal with draft-03 and draft-04 JSON instances in a project - that is the reason why i put efforts in these old meta schema files.

### DIFF
--- a/jsonschema/schemas/draft3.json
+++ b/jsonschema/schemas/draft3.json
@@ -15,7 +15,7 @@
 
         "properties": {
             "type": "object",
-            "additionalProperties": {"type": "object", "$ref": "#"},
+			"additionalProperties" : {"$ref" : "#"},
             "default": {}
         },
 
@@ -47,7 +47,7 @@
         },
 
         "dependencies": {
-            "type": ["string", "array", "object"],
+			"type" : "object",
             "additionalProperties": {
                 "type": ["string", "array", {"$ref": "#"}],
                 "items": {
@@ -157,11 +157,6 @@
             "format": "uri"
         },
 
-        "maxDecimal": {
-            "minimum": 0,
-            "type": "number"
-        },
-
         "$schema": {
             "type": "string",
             "format": "uri"
@@ -171,6 +166,7 @@
     "dependencies": {
         "exclusiveMinimum": "minimum",
         "exclusiveMaximum": "maximum"
-    }
+	},
 
+	"default" : {}
 }
--- a/jsonschema/schemas/draft3.json
+++ b/jsonschema/schemas/draft3.json
@@ -1,199 +1,176 @@
 {
     "$schema": "http://json-schema.org/draft-03/schema#",
-    "dependencies": {
-        "exclusiveMaximum": "maximum",
-        "exclusiveMinimum": "minimum"
-    },
     "id": "http://json-schema.org/draft-03/schema#",
+    "type": "object",
+
     "properties": {
-        "$ref": {
-            "format": "uri",
-            "type": "string"
+        "type": {
+            "type": ["string", "array"],
+            "items": {
+                "type": ["string", {"$ref": "#"}]
+            },
+            "uniqueItems": true,
+            "default": "any"
         },
-        "$schema": {
-            "format": "uri",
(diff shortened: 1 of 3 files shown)


Julian:
Thanks.

Some of this is reasonable I suppose to make merging any future changes to the metaschemas easier to merge in, but is there any other reason you had for this change?

Also there's at least one error here, not sure why you dropped the propertyNames change but that was not dropped upstream, on the contrary, we probably just haven't *added* it yet upstream

christian-weiss:
On the first look, i thought your files where just outdated and wanted to push the most recent version.
(I think using an outdated meta schema do not make sense)
On a second look i saw your patch. So now the PR is the official meta schem, but patched.

christian-weiss:
On draft-6:
There is properties.propertyNames, but no properties.patternProperties.propertyNames. That's the reason why properties.patternProperties.propertyNames was dropped from project file.

Julian:
Again though, that change is incorrect. The upstream schemas are *missing* it, it should not be removed, it should be upstreamed.

christian-weiss:
I got your comment wrong and thought that you accidentally reversed properties.propertyNames & properties.patternProperties.propertyNames. Now your patch No.2 is re-applied to draft-6.

Draft-3, draft-4 and draft-7 do have the properties.patternProperties element, but your patch was missing in the old project files - i guess by intention. If not, ping me, and i will re-apply the patch No.2 to these files too.

I can not find your PR to upstream. If it is missing: now would be a good point in time to create that PR - i think (you can explain, better then me, why patch No.1 and No.2, is necessary)

Julian:
Great, thanks, so what changes remain now? Is it just essentially re-ordering all the keys, and then some style differences in how to denote objects (and some trailing whitespace), or are there other functional differences you noticed?

christian-weiss:
see https://github.com/Julian/jsonschema/pull/641/commits/a1fa9cf0bb517b4730ba66d3b70c6f03a692b314
+ patch No1 + patch No2

Julian:
OK, thanks -- I'm still a bit uncomfortable merging this given that we have not yet hit json-schema-org/JSON-Schema-Test-Suite#244 -- I'm going to have to either review it carefully, which may take a bit longer.

daniela-waranie:
I my opinion you could run:
1. test suite against current version
2. test suite against PR version
3. merge if both tests results are equal

That should be good enough and you should see unintended side effects.
As this represents "merged at the best current knowledge".

In addition you could leave a comment in json-schema-org/JSON-Schema-Test-Suite#244(or in a repo-specific issue) to run the test on latest release and in case of test result errors to re-scan unmerged and merge version again, to check if it was introduced by the merge (revert changed file if required).

Waiting for a "perfect future" is a way to let your backlog grow more fast then it should.

Julian:
The test suite does not cover any of this behavior.

christian-weiss:
I did NOT spot a problem in the last 3 month. LGTM. Ready to merge. Could be released as 3.2.1-RC1 to let others send feedback, if you still worry about your test coverage.

Issue #621, opened 2019-11-08
CLI: Allow piping of data


cweiske:
It would be nice if I could pipe data into the `jsonschema` for validation:

    $ curl data.json | jsonschema schema.json

Julian:
Makes sense, happy to see a PR adding this.

Julian:
Closed via #643.

PR #643, opened 2019-12-31
Get input from commandline when instance(s) not provided

talham7391:
Added for this: https://github.com/Julian/jsonschema/issues/621

Can now do something like this:
`cat test.json | jsonschema test.schema`

### DIFF
--- a/jsonschema/tests/test_cli.py
+++ b/jsonschema/tests/test_cli.py
@@ -1,4 +1,10 @@
-from unittest import TestCase, mock
+from unittest import TestCase
+
+try:
+    from unittest import mock
+except ImportError:
+    from mock import mock
+
 import json
 import subprocess
 import sys
--- a/jsonschema/tests/test_cli.py
+++ b/jsonschema/tests/test_cli.py
@@ -3,11 +3,6 @@ import json
 import subprocess
 import sys
 
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import StringIO
-
 from jsonschema import Draft4Validator, ValidationError, cli, __version__
 from jsonschema.compat import NativeIO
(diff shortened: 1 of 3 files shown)


Julian:
Thanks, definitely reasonable!

I'd very much prefer not using mock for testing though -- there should (already I think) be a way to just pass the input you want on stdin -- can you give that a shot? If there isn't, it should be fairly easy to add, but if you need some direction happy to help there.

talham7391:
@Julian, I made the requested change, what do you think? Not sure why the checks are failing as I didn't touch those files.

jsonschema/tests/test_cli.py
```diff
@@ -149,3 +154,20 @@ def test_version(self):
         )
         version = version.decode("utf-8").strip()
         self.assertEqual(version, __version__)
+
+    def test_piping(self):
+        sys.stdin = StringIO("{}")
```
Julian: Thanks, this definitely now doesn't use mock, but it does now leak global state across test runs -- from here on out, every test will see `sys.stdin` as being the StringIO here.

What I had in mind was more that you should pass the input into the `cli.run` call below.

Specifically, if we add a `stdin` argument to `run` (with a default of `sys.stdin`), then you can pass your `NativeIO` for `stdin` to `run`, same as how `stdout` and `stderr` are being passed to it.

talham7391:
Looks like the Code Cov is missing repository upload token.

Julian:
The coverage thing is https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/td-p/30678/highlight/true/page/2 which I haven't fully made my mind up on whether to disable it for PRs or just make the token public.

Otherwise think everything looks good to go -- thanks! Going to have one last look and then merge.

PR #645, opened 2020-01-07
Remove the dependency on pyrsistent

jgarver:
Mimic the behavior of pyrsistent by creating shallow copies of dict instead.  Verify all tests still pass for 3.7.

Sharing this patch in case you are interested.  Because of our packaging constraints, we wanted to remove this dependency.

I will not be offended if you reject the PR.  ;-)

### DIFF
--- a/jsonschema/_types.py
+++ b/jsonschema/_types.py
@@ -1,6 +1,5 @@
 import numbers
 
-from pyrsistent import pmap
 import attr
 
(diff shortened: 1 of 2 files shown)


Julian:
:)

I sympathize, so thanks for sending this at least.

But yeah, if anything, `jsonschema` will likely move in the reverse direction and more of its internals will move to `pyrsistent`.

PR #647, opened 2020-01-13
Schema exception handling

abhilash2901:
**Changes**

- Schema exception handling.
- Formatted errors.


Julian:
Hi. Thanks.

Can you elaborate on what this patch is trying to accomplish?

As is, it's backwards incompatible (as well as missing tests, though most of the suite is going to fail with this regardless).

It doesn't seem likely to be something I'd merge, but perhaps if we discussed what you were trying to do with it there's another solution.

Julian:
Any comment here?

Julian:
Closing but happy to discuss.

PR #652, opened 2020-01-23
Add license file (COPYING) metadata to packaged distributions

johnthagen:
Currently [pip-licenses](https://github.com/raimon49/pip-licenses) can't include the LICENSE file for this project because the correct metadata is not setup.

This PR sets up the metadata correctly so the LICENSE file is properly included/linked to in source and wheel distributions.


Julian:
Thanks! Do you think you might be able to help add a test for whatever is
relied on by pip-licenses? Possibly in the form of a simple toxenv that
invokes pip-licenses and makes sure the result is MIT if that's the goal?

johnthagen:
@Julian Hmm, not sure how to go about that one.

`pip-licenses` already does get the overall license fine (it's just looking at standard `setuptools`/`wheel` metadata). This PR just adds another missing piece which tells `setuptools`/`wheel` where the LICENSE file is.

Not sure if you really need a test for this (it's just another piece of metadata like the other things in `setup.cfg`).

Julian:
Right understood but I already see the license file being included in the dists --

```
⊙  ls                                                                                                                                  
jsonschema-3.2.0-py2.py3-none-any.whl  jsonschema-3.2.0.tar.gz

~/Desktop/jsonschema
⊙  unzip -l jsonschema-3.2.0-py2.py3-none-any.whl | rg copying                                                                                                                                                                                                                                         
     1057  11-18-2019 12:57   jsonschema-3.2.0.dist-info/COPYING

~/Desktop/jsonschema
⊙  tar tf jsonschema-3.2.0.tar.gz | rg copying                                                                                                                                                                                                                                      
jsonschema-3.2.0/COPYING
```

I don't doubt there's some issue here and that this fixes it, but by a test I mean something that demonstrates what thing doesn't work before and works after adding this metadata -- running `pip-licenses` seemed like an obvious thing if that's the end-user thing that doesn't currently work, but if all `pip-licenses` relies on is some `setuptools` or `pip` API that I'm not aware of or thinking of at the minute then calling that too should cover it.

Otherwise there's not really a way to be sure this won't just break again.

johnthagen:
I believe the reason that the file is included is because it is in `MANIFEST.in`:

https://github.com/Julian/jsonschema/blob/9131e5cdc919899894255a8b5c94d88e4289f51c/MANIFEST.in#L3

So the file is present, but the metadata so that the distribution knows that this file is specifically a license file is not.

I'll see if I can come up with a test for `pip-licenses`.

johnthagen:
Okay, as a first step, here is a way to manually check that the LICENSE file is missing:

```bash
$ python3 -m venv jtest
$ source jtest/bin/activate
(jtest) $ pip install -U pip pip-licenses jsonschema
(jtest) $ pip list
Package            Version
------------------ -------
attrs              19.3.0 
importlib-metadata 1.4.0  
jsonschema         3.2.0  
more-itertools     8.1.0  
pip                20.0.1 
pip-licenses       2.0.1  
PTable             0.9.2  
pyrsistent         0.15.7 
setuptools         40.8.0 
six                1.14.0 
zipp               2.0.1  
(jtest) $ pip-licenses --from=mixed --with-license-file --output-file=out.txt
```

The `out.txt`:

```
...
 Name                Version  License                  LicenseFile                                                                                        LicenseText                                                                      
 jsonschema          3.2.0    MIT License              UNKNOWN                                                                                            UNKNOWN
```

johnthagen:
@Julian Your intuition to add a test was merited. In the course of digging into this, I learned that because `jsonschema` uses "COPYING" as it's license file name rather than "LICENSE", `pip-licenses` still couldn't find it even after the fix in this PR.

I opened https://github.com/raimon49/pip-licenses/issues/59

johnthagen:
@Julian CI test added and passing.

Julian:
> @Julian Your intuition to add a test was merited. In the course of digging into this, I learned that because jsonschema uses "COPYING" as it's license file name rather than "LICENSE", pip-licenses still couldn't find it even after the fix in this PR.

Thanks. "Glad" to hear. I didn't say it, but part of the intuition is `pip-licenses` sounds like a great idea but I'd never heard of it, and so also I think it'd be great to make sure what's here actually tests "known" behavior (from `pip` or from `setuptools`) and not just what `pip-licenses` has implemented -- I don't know how correct its implementation is, whether it relies on underlying "actual" APIs that are stable or some heuristics, etc. etc. -- something like not looking for a `COPYING` file sounded suspicious, since yeah, both files are common.

Your test looks definitely in the right realm (though I think it can be simplified yet a bit more -- surely `pip-licenses` can output things rather than requiring a file on disk?) -- I can give some more specific notes, but I want to yeah also understand what behavior `pip-licenses` is implementing and then will probably merge after trying to apply the simplification.

Much appreciated!

Julian:
I played around a bit, found that `pip show` is similarly confused about the license without including explicitly in `setuptools` a parameter for `license` (which is quite odd, but whatever).

I simplified the test to just call that directly, and reduced the change to just fix the test and not make further changes -- it looks like that also fixes `pip-licenses` for you -- so thanks for raising this!

Issue #600, opened 2019-09-03
Deprecate jsonschema.cli / the CLI


Julian:
`check-jsonschema` will be the recommended CLI going forward.

We can deprecate the existing one (and perhaps add a CLI setuptools extra).

PR #661, opened 2020-02-19
[CLI] Improve the overall console output (fix #623)

### LINKED ISSUE #623: CLI: No error location for error "None: None is not of type 'string'"
When jsonschema validates that a given object property is of type "string", and it is `null` in the data, it outputs:

    None: None is not of type 'string'

No more information is shown, especially not the property that failed to validate.

This makes finding the error with large schemas nearly impossible.

# Example
Schema:
```
{
    "type": "object",
    "properties": {
        "foo": {
            "type": "string"
        }
    }
}
```

Data:
```
{"foo": null}
```

Output:
```
$ jsonschema -i test-data.json test-schema.json 
None: None is not of type 'string'
```

RomainTT:
Fixes #623

This is just a suggestion, as there are plenty of ways to improve the visual. The current result is described below.

schema:

```
{
    "type": "object",
    "required" : ["bar"],
    "properties": {
        "foo": {
            "type": "string"
        },
        "bar": {
            "type": "string"
        }

    }
}
```

data:

```
{
    "foo": null
}
```

Command line:

```
$ jsonschema -i data.json schema.json
———|ERROR|—————————————————
'bar' is a required property

Failed validating 'required' in schema:
    {'$schema': 'http://json-schema.org/draft-07/schema#',
     'properties': {'bar': {'type': 'string'}, 'foo': {'type': 'string'}},
     'required': ['bar'],
     'type': 'object'}

On instance:
    {'foo': None}
———|ERROR|—————————————————
None is not of type 'string'

Failed validating 'type' in schema['properties']['foo']:
    {'type': 'string'}

On instance['foo']:
    None
```


Julian:
Thanks -- I'm definitely not opposed to a PR like this, but the "advantage" of the current ugly format is it's machine parsable in some sense.

I.e. you can grep it, and it's one line per error.

Possibly that's still not great, and there should be a `json` output (especially now that draft2019-09 has a standard JSON output representation).

I'm not opposed to having a human-readable one either as you're adding, but we should probably preserve the existing one, and leave room to add the new one (JSON output).

RomainTT:
I agree with your comments. I applied a different strategy (leading to a lot of new code, I’m sorry) which should suits you more.

- The main thing is a human-readable `-H` option. If not turned on, the behaviour is the same than today. If turned on, every printed message is made more readable.
- I had to change the way json files are parsed, it is no longer made by `argparse`. It allowed to introduced several `try: … except:` to improve the readability of exceptions that can occure. These exception messages are also changed in function of `-H`.
- I added comments to this file (`cli.py`) because it became more complicated than it was.

Here is an overview of the result:

**Default behaviour**
```sh
$ jsonschema -i data.json -i data.good.json -i data.corrupt.json schema.json   
{'foo': None}: 'bar' is a required property
None: None is not of type 'string'
Failed to parse data.corrupt.json. Got the following error: Expecting property name enclosed in double quotes: line 2 column 2 (char 3)
```

**Human readable**
```sh
$ jsonschema -H -i data.json -i data.good.json -i data.corrupt.json schema.json
===[ERROR]===(data.json)===
'bar' is a required property

Failed validating 'required' in schema:
    {'$schema': 'http://json-schema.org/draft-07/schema#',
     'properties': {'bar': {'type': 'string'}, 'foo': {'type': 'string'}},
     'required': ['bar'],
     'type': 'object'}

On instance:
    {'foo': None}
===[ERROR]===(data.json)===
None is not of type 'string'

Failed validating 'type' in schema['properties']['foo']:
    {'type': 'string'}

On instance['foo']:
    None
===[SUCCESS]===(data.good.json)===
===[ERROR]===(data.corrupt.json)===
Failed to parse data.corrupt.json. Got the following error: Expecting property name enclosed in double quotes: line 2 column 2 (char 3)
```

**And what if the schema is corrupted ?**
```sh
$ jsonschema -i data.json schema.corrupt.json 
Failed to parse schema schema.corrupt.json. Got the following error: Expecting property name enclosed in double quotes: line 4 column 3 (char 84)
```
That one is not changed with `-H`

I think the path is clear to implement the `JSON` report, either with a new variable in the `--error-format` option, or with a brand new command-line option.

Julian:
Thanks. It's definitely important we agree on approach.

Because of the third output format I mentioned (JSON), I'd rather see an option-based argument rather than a flag -- i.e. rather than `--human`, something like `--output=plain|pretty|json`. Which you don't need to add the third one in this PR, but then at least there's a logical place it can be added.

Also stylistically, I'd prefer we use helper functions, rather than comments -- if the function does too much usually to me that's a sign it should be split up, rather than have comments divide up sections of it. I try to keep functions to be 10 lines or less. The code though will likely change, so we can think about ways to do that perhaps after the above.

This is now also definitely enough code to bring up we should have tests for the new behavior. You'll notice also some of the existing ones are failing after the change. There's quite a bit here, so perhaps you can see how to unit test the changes to the output?

RomainTT:
I am really sorry that I had to make some force-pushes, but I had an identity issue and cannot let that as it was.

I finally managed to update the existing tests to make them pass. Now I am going to create some new ones to complete the coverage.

RomainTT:
Hi @Julian ! I finally reached a version I am happy with.

It implements what you suggested, it is fully retrocompatible, and lets room for further improvements. The console output is still the same than in my [previous comment](https://github.com/Julian/jsonschema/pull/661#issuecomment-588934138).

I fixed the existing tests and added some new ones.

The coverage does not reach 100% because I don’t think we can test the `sys.stdin` input in a unittest test suite, and for the rest of the coverage I don’t really understand why it is not reached. What do you think ?

If everything is ok for you, I am done with it.

jsonschema/cli.py
```diff
@@ -1,6 +1,10 @@
 """
 The ``jsonschema`` command line.
 """
+
```
Julian: Could you possibly remove these heading-style comments? I don't find they add much.

jsonschema/cli.py
```diff
@@ -9,39 +13,134 @@
 from jsonschema import __version__
 from jsonschema._reflect import namedAny
 from jsonschema.validators import validator_for
+from jsonschema.exceptions import SchemaError, ValidationError
+
+
+# ------------------------------------------------------------------------------
+# CLASSES
+# ------------------------------------------------------------------------------
+class CliOutputWriter():
```
Julian: For better or **worse** this module is still public (which #600 will fix) -- so we should be careful not to add any *more* public things to it. Can you rename this to be private?

But also -- I think you have the right idea here of an output writer, but if you use object polymorphism, things get even nicer.

I.e., rather than have every method have `if whatever == "pretty": do_something_pretty()`, have `_PrettyOutputWriter` and `_PlainOutputWriter` with the same interface, and then just select which writer is used based on the command line flag.

Lemme know if that makes sense to you, otherwise happy to elaborate.

jsonschema/tests/test_cli.py
```diff
@@ -66,36 +66,89 @@ def test_find_validator_in_jsonschema(self):
...
+        )
+        self.assertIs(arguments["validator"], LatestValidator)
+
+    def test_none_instance(self):
+        arguments = cli.parse_args(
+            [
+                self.schema_file,
+            ]
+        )
+        self.assertTrue(
```
Julian: Any reason not to just make this `self.assertEqual(arguments["instances"], [])`?

jsonschema/tests/test_cli.py
```diff
@@ -173,3 +229,111 @@ def test_piping(self):
...
+                "validator": fake_validator(),
+                "schema": "schema.json",
+                "instances": ["foo2.json"],
+                "error_format": "",
+                "output": "pretty",
+            },
+            stdout=stdout,
+            stderr=stderr,
+        )
+        self.assertTrue(self.pretty_success_tag in stdout.getvalue())
```
Julian: `self.assertIn` will produce slightly nicer error messages.

jsonschema/tests/test_cli.py
```diff
@@ -131,15 +185,16 @@ def test_unsuccessful_validation_multiple_instances(self):
...
+                "instances": ["foo1.json", "foo2.json"],
+                "error_format": "{error.instance} - {error.message}",
+                "output": "plain",
             },
             stdout=stdout,
             stderr=stderr,
         )
         self.assertFalse(stdout.getvalue())
-        self.assertEqual(stderr.getvalue(), "1 - 9\t1 - 8\t2 - 7\t")
+        self.assertEqual(stderr.getvalue(), "1 - 9\n1 - 8\n2 - 7\n")
```
Julian: I have to read this more carefully, but this seems like a change in behavior.

Specifically, originally it looks like there's no implicit newline in `error_format`, but now it looks like there is.

It's good I think to minimize changes to unrelated tests. Here I assume this happened because you changed the type of `arguments["instances"]` to be filenames instead of the read contents right?

jsonschema/cli.py
```diff
@@ -9,39 +13,134 @@
...
+                exception=exception,
+            )
+        else:  # pragma: no cover
+            raise ValueError(
+                "Output mode '{}' is unknown by this function"
+                .format(self.output_format)
+            )
+        self.stderr.write(msg)
 
+    def write_valid_error(self, object_name, error_obj):
```
Julian: Can we avoid abbreviations? Here `write_valid_error` is particularly confusing, I assume this means `write_validation_error`?

jsonschema/cli.py
```diff
@@ -9,39 +13,134 @@
...
     dest="instances",
-    type=_json_file,
+    type=str,
     help=(
-        "a path to a JSON instance (i.e. filename.json) "
-        "to validate (may be specified multiple times)"
+        "A path to a JSON instance (i.e. filename.json) "
+        "to validate (may be specified multiple times). "
+        "If this option is not used, an instance can be given through stdin. "
+        "If none is given, only the schema is checked."
```
Julian: I don't think I follow the last sentence here -- is `none` different from "not given"? And if it means `None`, how does someone get that through from a command line?

jsonschema/cli.py
RomainTT: Of course that is much better ! I just didn’t think about it. But it is definitely the kind of things I like to write in Python.

jsonschema/cli.py
RomainTT: Ok everyone has its own conventions :) I’ll remove that.

jsonschema/cli.py
RomainTT: What I meant was “ If neither -i nor stdin is given, only the schema is checked."
I am changing that.

jsonschema/cli.py
RomainTT: Right. This is a confusing abbreviation, I did not notice it myself sorry.

jsonschema/tests/test_cli.py
RomainTT: I just discovered `unittest` while writing this, I’m more used to `pytest` where those nice assertions do not exist. Thanks for the tip!

jsonschema/tests/test_cli.py
RomainTT: Oops, indeed that is a change, not necessarily justified.
It comes from the following lines in the `CliOutputWriter`:

```python
if not oneline_format.endswith("\n"):
    self.oneline_format += "\n"
```

I’ll just remove that so user can use the format he wants. As the original behaviour, the default format if not given by user includes this `\n`

RomainTT: And I'll rename `oneline_format` to `plain_format`, more accurate.

jsonschema/tests/test_cli.py
RomainTT: At first I thought it would be more precise to check the type explicitly, but it is true that `equal to []` also checks the type. I was afraid of Python’s weird side effects. But you’re right.

jsonschema/cli.py
```diff
@@ -1,14 +1,107 @@
...
 import json
 import sys
 
 from jsonschema import __version__
 from jsonschema._reflect import namedAny
 from jsonschema.validators import validator_for
+from jsonschema.exceptions import SchemaError, ValidationError
+
+
+class _CliOutputWriter():
```
Julian: Thanks! This was close to what I meant, but there isn't really any benefit to the abstract base class (or the inheritance in general) -- the subclasses can just live on their own without any relationship in the inheritance hierarchy.

jsonschema/tests/test_cli.py
```diff
@@ -173,3 +226,111 @@ def test_piping(self):
         self.assertFalse(stdout.getvalue())
         self.assertFalse(stderr.getvalue())
         self.assertEqual(exit_code, 0)
+
+    def test_schema_parsing_error(self):
```
Julian: Perhaps a test for providing an unknown output format would be nice. The way things are implemented (via `choices`) you probably just can call `parse_args`, since the actual function won't run.

jsonschema/cli.py
```diff
@@ -17,31 +110,39 @@ def _namedAnyWithDefault(name):
...
+        "If this option is not used, an instance can be given through stdin. "
+        "If neither -i nor stdin is given, only the schema is checked."
     ),
 )
 parser.add_argument(
     "-F", "--error-format",
     default="{error.instance}: {error.message}\n",
     help=(
-        "the format to use for each error output message, specified in "
+        "The format to use for each error output message, specified in "
```
Julian: Maybe it should be an error if this is used with pretty?

jsonschema/cli.py
```diff
@@ -17,31 +110,39 @@ def _namedAnyWithDefault(name):
...
     dest="instances",
-    type=_json_file,
+    type=str,
     help=(
-        "a path to a JSON instance (i.e. filename.json) "
-        "to validate (may be specified multiple times)"
+        "A path to a JSON instance (i.e. filename.json) "
+        "to validate (may be specified multiple times). "
+        "If this option is not used, an instance can be given through stdin. "
+        "If neither -i nor stdin is given, only the schema is checked."
```
Julian: This behavior seems true (that if you don't provide an instance the schema is still validated), but I'm not sure it was fully intentional.

It's sensical certainly, so I'm happy definitely documenting it, but I think adding a test to make sure it stays the case would probably be smart.

jsonschema/cli.py
```diff
@@ -60,32 +161,122 @@ def _json_file(path):
...
 def parse_args(args):
     arguments = vars(parser.parse_args(args=args or ["--help"]))
     if arguments["validator"] is None:
         arguments["validator"] = validator_for(arguments["schema"])
+    if arguments["instances"] is None:
+        arguments["instances"] = []
     return arguments
 
 
+def make_validator(schema_path, validator_class, output_writer):
```
Julian: Can you make all these private as well?

jsonschema/cli.py
RomainTT: Indeed there is not in a functional point of view. But I think it helps to enforce a common interface to all of these classes, and it avoids some duplicate code (like CONSTANTS or common initialization). As in the future a new class will be added to print a `json` report, I think this architecture will be even more useful.

Maybe it is a personal preference. But if you think it harms more than it helps, I’ll remove it. You tell me.

jsonschema/tests/test_cli.py
RomainTT: I added a test in `TestParser`, is that what you were thinking about ?

jsonschema/cli.py
RomainTT: In that case the only way is to remove the default from argparse and manually process the arg. I added a new test for that.

jsonschema/cli.py
RomainTT: Added in `test_schema_validation`

jsonschema/cli.py
RomainTT: :+1:

jsonschema/cli.py
Julian: Great, thanks!

jsonschema/tests/test_cli.py
```diff
@@ -66,36 +67,118 @@ def test_find_validator_in_jsonschema(self):
...
+                arguments = cli.parse_args(
+                    [
+                        "--output", "foo",
+                        self.schema_file
+                    ]
+                )
+
+    def test_useless_error_format(self):
+        # Avoid the help message on stdout
+        argparse_mock = MagicMock()
```
Julian: This looks suspicious, definitely would rather well, not use mock at all, but if we're forced into it (which sometimes happens with argparse :/) then not be patching private things.

I don't see you asserting about stdout/error here, what happened if you remove this that wasn't desirable? In theory I'd just expect this test to assert that we exit-nonzero for this case, even when the instance and schema are valid.

jsonschema/tests/test_cli.py
Julian: Not sure I see it, maybe I'm missing it?

I'm talking about what happens if someone passes `--output=gibberish`.

jsonschema/tests/test_cli.py
RomainTT: I’m sorry I don’t get it.

`test_unknown_output` and `test_useless_error_format` are checking cases of invalid options given to `argparse`. In command line, argparse stops the program and returns a non-zero code. But here I cannot test the command line, I therefore directly call `cli.parse_args` which itself calls `argparse.parse_args` which **raises an exception** and **writes on stderr** when incorrect options are given. The consistency check of `--output` and `--error-format` is manually done by us, but an exception is raised in the same way thanks to `parser.error()`.

Hence the mock and this test. I don’t know how to capture *stderr* otherwise, and the exception seems to be the only indicator of a failure. If I was able to capture *stderr*, I agree we could also check that it is not empty.

Am I missing something ?

jsonschema/tests/test_cli.py
RomainTT: It is done by `test_unknown_output`, I’m writing about it [in this comment](https://github.com/Julian/jsonschema/pull/661#discussion_r390155906)

jsonschema/cli.py
RomainTT: Your code, your decision :blush: :+1:

jsonschema/tests/test_cli.py
RomainTT: About the unittest mock, I found another way to capture *stderr*, by creating a custom `@contextmanager` (in `_helpers.py`).

RomainTT: And finally, since I can capture these output, I added a test on both of them. Nothing must be in `stdout`, and a specific string must be in `stderr`. Maybe that is closer to what you expected.

jsonschema/tests/test_cli.py
Julian: Ah I see now, thanks!

RomainTT:
Do you want me to resolve the opened conversations or you do it ?

Julian:
We are close! I want to do one lookover this weekend when I have some
mindspace, there are a few small things I'd still like to fix and then we
should be good to go.

What do you prefer? You've been super engaged so thanks! If you're still up
for it I'll likely do a few things that are quicker to tweak than explain,
but there may be one or two other small things that I can leave a comment
for if you're still interested in one more round of tweaks, otherwise I can
knock them out as well.

Either way much appreciated!

RomainTT:
I’m ready to make any change ! If you think it will be faster to share the work, just ask :)


Otherwise you can do it all, I don’t mind. Thank you for your support in this PR !


De : Julian Berman <notifications@github.com>
À : Julian/jsonschema <jsonschema@noreply.github.com>
Sujet : Re: [Julian/jsonschema] [CLI] Improve the overall console output (fix #623) (#661)
Date : 13/03/2020 20:33:10 Europe/Paris
Copie à : Romain Taprest <romain@taprest.fr>;
   Author <author@noreply.github.com>


We are close! I want to do one lookover this weekend when I have some
mindspace, there are a few small things I'd still like to fix and then we
should be good to go.

What do you prefer? You've been super engaged so thanks! If you're still up
for it I'll likely do a few things that are quicker to tweak than explain,
but there may be one or two other small things that I can leave a comment
for if you're still interested in one more round of tweaks, otherwise I can
knock them out as well.

Either way much appreciated!

Julian:
So! After quite a bit of noodling, some of which not directly related to the PR, but this file needed some love, I'm going to merge.

There's still a bit more work I want to do to clean some of these files up but your contribution is very much appreciated!

Thank you for jumping in! Getting these things off the ground is 90% of the trouble typically, so if it weren't for you it wouldn't have happened :).

RomainTT:
@Julian nice !!!

Thank you very much for your support !

I have another PR coming for you :kissing_closed_eyes:

Issue #575, opened 2019-07-01
`0`  should not be considered equal to `False`


Zac-HD:
I'm working on some schema canonicalisation and intersection logic for [`hypothesis-jsonschema`](https://github.com/Zac-HD/hypothesis-jsonschema), and encountered the surprising-to-me property that `jsonschema` *sometimes* considers `0 == False` (and similarly `1 == True`):

```python
# All of these pass:
validate(0, {"just": False})            # 0 == False
validate(False, {"enum": [0.0]})        # 0.0 == False
# Fails, "On instance: [0, 0]", despite `unbool` logic
validate([0, False], {"uniqueItems": True})
```

[According to the core spec](https://json-schema.org/latest/json-schema-core.html#rfc.section.4.2.3), instances are only equal if they are of the same type.

(and there is only one number type in JSON, so `1 == 1.0`)

~~The validation spec doesn't actually say anything about how value equality (ctrl-f "equal") should be defined, but I certainly found this surprising and the `uniqueItems` message convinced me to report it.~~

Julian:
The latter likely related to your still-open #529.

I actually... don't have an opinion yet :D, have to think about what I expect.

Obviously the inconsistency is wrong, but not sure which to expect. I guess slightly leaning towards them always being equal, but have to think.

Thoughts welcome of course :), and yeah thanks for filing obviously!

Zac-HD:
It's not obvious, thats for sure!  My preference would be to treat all numeric zeros (integer or float) as equal, but booleans as *unequal* to numbers. Either way, it's a question that should be addressed by the specification since it does change the semantics.

Treating booleans as unequal to numbers unlocks several nice optimisations in the logic I usr for canonicalising schemas and calculating intersections - with I can treat them as enums; without I have to keep and track the type information too. Not sure if that will convince anyone but me though 😅

Final fun fact: IEEE754 allows you to represent negative zero, but it can't be serialised as JSON. So hypothesis-jsonschema already has special handling for zeros, even before this issue!

gregsdennis:
Having `0` synonymous to `false` is certainly a language-specific thing.  It is in many C-based languages, but notably not in C# (or any .Net language).  I imagine there are other such cases in other strongly-typed languages where integers and booleans are distinct concepts.

[...]

Issue #164, opened 2014-05-05
Is _unset check in ValidationError fails if custom-type has overriden __eq__() method (ie: pandas.DataFrame)


ankostis:
I use a custom schema that accepts as sub-trees `pandas.DataFrame` and `pandas.Series` instances, classes that have overridden the **eq**() method.

In case a `ValidationError` involves such an object, the generated exception is thereafter unusable, since the following check within it **unicode**() method fails:

```
    if _unset in (
        self.validator, self.validator_value, self.instance, self.schema,
    ):
```

The reason for this is that `_unset == self.instance` (what the compiler generates for `XX in YY` expressions) returns multiple boolean values when  the `self.instance` is a `pandas.DataFrame`.

Since `_unset` is a sentinel instance, i believe that a more proper approach would be to use the `is` operator instead, which it cannot be overridden.

The following code snippet would suffice:

```
if any(m is _unset for m in (
    self.validator, self.validator_value, self.instance, self.schema
)):
```

I would be glad to provide such a patch.

PR #675, opened 2020-04-03
Compare with `is` (vs ==) in enum rule

ankostis:
Similar to #164, this patch avoids calling overridden `__eq__()` in the modified `enum` rule after v3.0.2 (due to #575),
so that e.g. pandas (series & dataframes) don't choke on enum rules.

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -241,7 +241,7 @@ def dependencies(validator, dependencies, instance, schema):
 
 
 def enum(validator, enums, instance, schema):
-    if instance == 0 or instance == 1:
+    if instance is 0 or instance is 1:
         unbooled = unbool(instance)
         if all(unbooled != unbool(each) for each in enums):
             yield ValidationError("%r is not one of %r" % (instance, enums))


Julian:
Hi. Thanks.

This one is trickier -- it's not necessarily correct here. Unlike #164, those aren't singletons -- an implementation is free to have `0 == 0` but the two 0's not being the same object.

Julian:
(So might need to think about what a solution that'd work here would be, but a failing test would be great as well!)

ankostis:
> but a failing test would be great as well!

I had long thought of building one, just to weed out some more hurdles when dealing with pandas as instances (it is possible to do that in the end, needs 5-6 interventions in the rules).

Let's hope, now that i managed to teach my quarantined daughter not to interrupt me too much :-), i to serve such a test case.  
Would it be ok if it imported pandas?

Julian:
> Let's hope, now that i managed to teach my quarantined daughter not to interrupt me too much :-), i manage to serve such a test case.

Ha! Good luck (hope you, her and the rest of your family are all well...)

>  Would it be ok if it imported pandas?

If need be that's fine, though I suspect it'll be simpler to define a small class with an `__eq__` that has the peculiarity that's relevant here? But if it's tons simpler either is fine.

PR #678, opened 2020-04-10
add is_valid function as shortcut for validate + except ValidationError

web-devil:


Julian:
Hi. Thanks for the contribution!

This exists already [as an attribute on `Validator` instances](https://python-jsonschema.readthedocs.io/en/stable/validate/#jsonschema.IValidator.is_valid).

As a global function I'd rather not add it, since the intention is to just have one simple interface for new users, which `jsonschema.validate` serves -- anything more is meant to be served by instantiating a validator.

Issue #669, opened 2020-03-20
Support specifying a base URI explicitly to the CLI


koteez:
Hi,

I have a json schema that depends on other files that are located in the same directory. Running the CLI I got an issue because the resolver is trying to load the file from the same URI as my main script. As I am working locally I didn't have a server serving the schemas and thus the resolution fails. 

```
{
 "ArrayA":{
    "type": "array",
     "items": {
	  "$ref": "common.json#/definitions/ObjectB"
      }
   }
} 
```
Using __file:__ require to put the absolute path which is a non-sense inside the schema definition.
What could be intesting to do is having a parameter to the CLI so that we can override the base_uri with $PWD value which will work.

Thanks & regards

K

sirosen:
I believe -- I'm not totally certain -- that the local file resolution behavior used in `check-jsonschema` would fit this.
When a local schemafile is used, it sets the resolver to use a base-ref pointed at the filename.

I'm aware this issue is pretty old, but I'd love to hear from the OP if this works.

_I am the author of `check-jsonschema` and we are currently discussing making it the official CLI for `jsonschema`. For details, you can follow #910. If `check-jsonschema` has a behavior which you believe should be changed to handle your use-case, please [open an issue](https://github.com/sirosen/check-jsonschema/issues/new)!_

sirosen:
We're going to move forward and begin calling `check-jsonschema` the official CLI. I'm therefore closing this as a wontfix. I _believe_ that the behavior in `check-jsonschema` is sufficient. If not, I'm happy to take a bug report on that repo.

PR #680, opened 2020-04-16
WIP: [CLI] Add new option -r to resolve local file references

RomainTT:
When the CLI is used, the local path to the schema is known. Therefore, if some other files are referenced with the `ref` key, the program knows where to search for them.

In the [documentation](https://json-schema.org/understanding-json-schema/structuring.html?highlight=ref#structuring-a-complex-schema) of _json-schema_, it is stated:

> $ref can also be a relative or absolute URI, so if you prefer to include your definitions in separate files, you can also do that. For example:
> 
> { "$ref": "definitions.json#/address" }
> would load the address schema from another file residing alongside this one.
> 

And:

> The $id property is a URI that serves two purposes:
> 
> It declares a unique identifier for the schema.
> It declares a base URI against which $ref URIs are resolved.
> […]
> 
> But be aware of the second purpose of the $id property: that it declares a base URL for relative $ref URLs elsewhere in the file. For example, if you had:
> 
> { "$ref": "person.json" }
> in the same file, a JSON schema validation library that supported network fetching would fetch person.json from http://foo.bar/schemas/person.json, even if address.json was loaded from somewhere else, such as the local filesystem.
> 

I think that feature can be supported by the command line. I therefore added the `-r` option.

### DIFF
--- a/jsonschema/cli.py
+++ b/jsonschema/cli.py
@@ -8,6 +8,7 @@ import argparse
 import errno
 import json
 import sys
+import os
 import traceback
 
 import attr
@@ -16,7 +17,7 @@ from jsonschema import __version__
 from jsonschema._reflect import namedAny
 from jsonschema.compat import JSONDecodeError
 from jsonschema.exceptions import SchemaError
-from jsonschema.validators import validator_for
+from jsonschema.validators import validator_for, RefResolver
 
 
 class _CannotLoadFile(Exception):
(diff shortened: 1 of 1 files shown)


RomainTT:
The main issue I have is that I don’t know how to make tests about that. My understanding of the current CLI tests is that there is no actual file at all, only fake openings. In that case, how to test this feature ? Maybe by redefining some of the `os` methods ?

Julian:
Hi! Sorry, haven't gotten a chance to look at this yet but thanks again for the PR! Really appreciated.

Will try to give some guidance on how to test this (I *suspect* before digging to hard that an easy way to test it may be to rely on the metaschema files that are already shipped in the repo, and using those as the one you reference via relative references)

Also, on the option itself, I think my first reaction is I'd rather call this something like `--base-uri`? (Similar to what's in #669 which is about this essentially). What's being specified is a base URL that's used to resolve relative references, yeah?

RomainTT:
wow I totally forgot that PR, sorry. I’ll try to continue it soon.

Issue #686, opened 2020-05-18
`[[False], [0]]` should be considered unique


Zac-HD:
```python
jsonschema.validate([[False], [0]], {'type': 'array', 'uniqueItems': True})
```

fails, though the inner lists should be considered distinct due to the different types of their values.  (though all numbers are considered the same type, so `[[0], [0.0]]` should and does fail to validate)

This should probably go in the upstream test suite as well.  Checked on version 3.2.0.

(sorry for the short report, found this just as I was heading to bed)

Julian:
Closed (by @nezhar, thanks!) in 7b9b5c3.

PR #688, opened 2020-05-21
Fix bug with [False] and [0] equal

nomnoms12:
https://github.com/Julian/jsonschema/issues/686
Hey. I tried to fix the error. Hope this helps.

An additional isinstance check is needed because Python considers [False] and [0] equal.
I also had to remove the sort to avoid comparison errors.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -174,13 +174,17 @@ def equal(one, two):
 
 def unbool(element, true=object(), false=object()):
     """
-    A hack to make True and 1 and False and 0 unique for ``uniq``.
+    A hack to make True and 1, False and 0, [True] and [1], [False] and [0] unique for ``uniq``.
     """
 
     if element is True:
         return true
     elif element is False:
         return false
+    elif element == [True] and isinstance(*element, bool):
+        return [true]
+    elif element == [False] and isinstance(*element, bool):
+        return [false]
     return element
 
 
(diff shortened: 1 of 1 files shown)


Julian:
Hi! Thanks! Very happy to see you take a shot at this.

We need to add a test for this upstream to the test suite (here: https://github.com/json-schema-org/JSON-Schema-Test-Suite) so the first step here is to add one to that. You can test out whether that test exercises the bug by adding it directly to the `json` directory here in this repository -- that directory is just a literal copy of that test suite git repository.

Also, as for the fix itself, I suspect what we have in your diff will fix the specific instance @Zac-HD found (thanks to him as usual :), but intuitively seems unlikely to be correct I think -- it will likely fail again once a single additional layer of nesting is introduced. In order to fix it once and for all, the code in this area essentially tries to do lots of lazy unwrapping -- but give it a shot with both adding a test and then adding one that's doubly nested, I think you may see the trick here.

Zac-HD:
@Julian - you might want to steal [my `encode_canonical_json()` function](https://github.com/Zac-HD/hypothesis-jsonschema/blob/539f28776d383d1cac35250c09e25fcef357ea52/src/hypothesis_jsonschema/_canonicalise.py#L66-L98); it serialises JSON objects to strings which are equal if-and-only-if JSONschema considers the objects equal.  Not suitable for runtime use where performance is an issue, but it's great for testing!

nomnoms12:
@Julian Hello again :)

I tried to fix this error by implementing [breadth first search](https://en.wikipedia.org/wiki/Breadth-first_search) algorithm to apply the `unbool` function to every element in the container. In code, this looks pretty complicated because I had to process dictionaries and lists at the same time, but the implementation is non-recursive and I think it should be better than dumping to a string.

I'd love to hear what you think about it.

Julian:
Thanks! Will have a look -- one immediate thing that pops out though is that `jsonschema` supports arbitrary container types (via [`TypeChecker`](https://python-jsonschema.readthedocs.io/en/stable/validate/#type-checking)) so we can't type check on `list` and `dict` concretely -- we'd in theory have to use the `collections.abc` types..

If the existing implementation doesn't fail on some test (which it seems it doesn't), we probably should add a test (now just in this library, not upstream) to make sure we don't forget that we can have any type there, not just lists and dicts.

But will look more closely at the strategy you used, and definitely thanks again for trying to tackle this.

PR #697, opened 2020-06-13
Teach unbool about deeply nested objects/arrays

### LINKED ISSUE #686: `[[False], [0]]` should be considered unique

rjmill:
Fixes #686

I'm aware that TypeChecker throws a wrench into this, but I want to see if this approach seems reasonable to you before I go any farther.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -1,4 +1,3 @@
-import itertools
 import json
 import pkgutil
 import re
@@ -198,26 +197,6 @@ def unbool(element, true=object(), false=object()):
 def uniq(container):
     """
     Check if all of a container's elements are unique.
-
-    Successively tries first to rely that the elements are hashable, then
-    falls back on them being sortable, and finally falls back on brute
-    force.
     """
 
-    try:
-        return len(set(unbool(i) for i in container)) == len(container)
-    except TypeError:
-        try:
-            sort = sorted(unbool(i) for i in container)
-            sliced = itertools.islice(sort, 1, None)
-            for i, j in zip(sort, sliced):
-                if i == j:
-                    return False
-        except (NotImplementedError, TypeError):
-            seen = []
-            for e in container:
-                e = unbool(e)
-                if e in seen:
-                    return False
-                seen.append(e)
-    return True
+    return len(set(unbool(i) for i in container)) == len(container)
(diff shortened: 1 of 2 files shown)


willson-chen:
This PR looks good to me. I have tested the following four scenarios, and all got the expected result.

### Code
```
from jsonschema.validators import Draft7Validator, exceptions
validator = Draft7Validator({'type': 'array', 'uniqueItems': True})

# scene 1: nested dictionary
print(exceptions.best_match(validator.iter_errors([{'1': {'1': '1'}}, {'1': {'1': '1'}}])))
print("-----------------------------------------------------------------------")

# scene 2: nested list
print(exceptions.best_match(validator.iter_errors([[[False], [0]], [False, 0], [[False], [0]]])))
print("-----------------------------------------------------------------------")

# scene 3: nested tuple
print(exceptions.best_match(validator.iter_errors([('1', (1, 0), '1'), ('1', (1, 0), '1')])))
print("-----------------------------------------------------------------------")

# scene 4: nested set
print(exceptions.best_match(validator.iter_errors([{1, 2, 3}, {1, 2, 3}, {1, 1, 1}])))
```

### Result
```
[{'1': {'1': '1'}}, {'1': {'1': '1'}}] has non-unique elements

Failed validating 'uniqueItems' in schema:
    {'type': 'array', 'uniqueItems': True}

On instance:
    [{'1': {'1': '1'}}, {'1': {'1': '1'}}]
-----------------------------------------------------------------------
[[[False], [0]], [False, 0], [[False], [0]]] has non-unique elements

Failed validating 'uniqueItems' in schema:
    {'type': 'array', 'uniqueItems': True}

On instance:
    [[[False], [0]], [False, 0], [[False], [0]]]
-----------------------------------------------------------------------
[('1', (1, 0), '1'), ('1', (1, 0), '1')] has non-unique elements

Failed validating 'uniqueItems' in schema:
    {'type': 'array', 'uniqueItems': True}

On instance:
    [('1', (1, 0), '1'), ('1', (1, 0), '1')]
-----------------------------------------------------------------------
[{1, 2, 3}, {1, 2, 3}, {1}] has non-unique elements

Failed validating 'uniqueItems' in schema:
    {'type': 'array', 'uniqueItems': True}

On instance:
    [{1, 2, 3}, {1, 2, 3}, {1}]
```

PR #703, opened 2020-07-15
fix issue 601

willson-chen:
The original logic is not changed based on the relative path of the current file. For other files, delete the relative path and retain the value of base_uri. 
I tried, and it solved the questioner's problem, and then it didn't affect the previous test case. And I'm adding test cases for later submission.

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -257,7 +257,10 @@ def ref(validator, ref, instance, schema):
                 yield error
     else:
         scope, resolved = validator.resolver.resolve(ref)
-        validator.resolver.push_scope(scope)
+        if "#" in ref:
+            validator.resolver.push_scope(scope)
+        else:
+            validator.resolver.push_scope(str(scope).replace(ref, ""))
 
         try:
             for error in validator.descend(instance, resolved):


Julian:
@willson-chen hey! Thanks for trying to tackle these -- probably good to focus on just one (or the other), but yeah let me know whenever it's ready for review?

willson-chen:
The file protocol has compatibility issues in Linux and Windows. Therefore, compatibility processing is performed in the code. 
image
However, the coverage rate report contains only the result in Linux, which causes the coverage rate decrease.

Julian:
Hi @willson-chen thanks! This looks like a nice start.

So, couple of things:

The `jsonschema/schemas` directory is where "metaschemas" are cached -- in other words, the schemas that go there are essentially just the draft schemas for each draft. That's likely not obvious to tell, but here instead where these schemas would go is somewhere inside the test suite.

However! In these cases where we're dealing with a bug in the implementation of the specifications, what we really need to do is have the test be in the [official language-agnostic test suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite) (which I also maintain, and which you'll find a copy of in the `json` directory in the root of this repository).

The suite may, even likely *does* already contain a test case that covers this behavior -- you'll find I skip some tests with "known bugs" -- e.g. see here: https://github.com/Julian/jsonschema/blob/master/jsonschema/tests/test_jsonschema_test_suite.py#L131-L133

What we need to do is reduce the bug here to a minimal example so we can see precisely what the issue is, then see if there is an already-skipped test for that behavior. There should be. But if there isn't, then we should add the bug to the official test suite, written as JSON, and minimized. I can help with that if it turns out to be the case. But definitely the first 2 steps are to see whether you can minimize the bug to it's smallest form and then see if one of the existing skipped tests fails before your patch and then succeeds afterwards.

(As a specific example of minimization -- the bug definitely doesn't relate to having 4 fields in the point, likely just 1 will exercise the same issue, so the goal is to simplify the test as much as possible)

I'm pretty skeptical the fix should go in `_validators.ref` -- this is the implementation of the `"ref"` validator, and it really should be relying on `RefResolver` for all of the actual logic in resolving refs, but that may become more obvious after the above.

Let me know if the above makes sense! Much appreciated again!

willson-chen:
@Julian Test cases have been added to the upstream test library. The link is as follows:
[subschema nested ref in remote ref](https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/421)

Issue #708, opened 2020-07-22
cleanup of `pretty` output renderer


sloanlance:
Based on a conversation started in the comments of #706, clean up the appearance of the new `pretty` output renderer.  Pertinent points from the conversation…

* [x] Fix length of the output dividers (current length is arbitrary; the length of `=` and `-` lines do not match)
* [x] Change to Unicode box characters from ASCII
* [x] Add vertical space between each error
* [x] Change exception messages from Python object `repr`s (with single-quoted strings) to true JSON (with double-quoted strings)

<details>
<summary>Details</summary>
I think more important than adding even more output renderers is essentially cleaning up the one new one that was added. Specifically, here's some example output from the new pretty renderer:

```
⊙  printf '7' | jsonschema --output pretty <(printf '{"minimum": 9, "multipleOf": 3, "not": {"const": 7}}')

===[ValidationError]===(<stdin>)===

7 is less than the minimum of 9

Failed validating 'minimum' in schema:
    {'minimum': 9, 'multipleOf': 3, 'not': {'const': 7}}

On instance:
    7
-----------------------------
===[ValidationError]===(<stdin>)===

7 is not a multiple of 3

Failed validating 'multipleOf' in schema:
    {'minimum': 9, 'multipleOf': 3, 'not': {'const': 7}}

On instance:
    7
-----------------------------
===[ValidationError]===(<stdin>)===

{'const': 7} is not allowed for 7

Failed validating 'not' in schema:
    {'minimum': 9, 'multipleOf': 3, 'not': {'const': 7}}

On instance:
    7
-----------------------------
```

where some things that rub me a bit wrong visually are:

* the width of the ascii dividers (right now it's a bit arbitrary and you can see it doesn't match between the `=` and `-` lines -- particularly so if it's a `SchemaError` vs a `ValidationError`)
* not using unicode box characters
* needing a bit of vertical space between each error
* the fact that, though it's not obvious there, in the exception message, what's there is reprs of Python objects. Even though usually that's what I _want_ shown because jsonschema normally deals with Python objects, not JSON, here that's _not_ the case -- here in fact the reverse is true -- the CLI takes JSON, and does the deserialization itself. So when showing errors back to users of the CLI, it should show JSON, not Python objects.
</details>

sloanlance:
What do you think of this output, @Julian, with divisions padded to a width of 79 characters per line?

```
╒══[ValidationError]═══(<stdin>)══════════════════════════════════════════════╕
[...]

PR #709, opened 2020-07-22
[compat] Drop support of PY2

Djailla:
Allowing to drop six requirement

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,14 +17,6 @@ jobs:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
         python-version:
-          - name: pypy2
-            toxenv: pypy2-build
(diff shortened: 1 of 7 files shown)


Djailla:
Hello @Julian 

The CI is marked as red due to coverage, but you could be interested ?

Julian:
@Djailla hi! Yes, thanks for sending it.

This likely is something I'll want to do in the next few months, but there's likely a few things higher on the list for limited time :/ -- the one thing I see is we need to set a `python_requires` in the `setup.cfg` as part of this change, so that we don't break Python 2 for any existing versions.

But beyond that yes would be interested, just may take me a bit to get to this.

Thanks for sending it!

Julian:
Merged, thanks!

sloanlance:
Awesome!  This solves a lot of the headaches I was having with #708!

PR #712, opened 2020-07-26
[DRAFT] #708 - Make pretty formatter output prettier

### LINKED ISSUE #708: cleanup of `pretty` output renderer

sloanlance:
Fixes #708.

This makes the pretty output look better by making error header and footer lines equal lengths and to use Unicode characters in them.  It also shows schema JSON in error messages as valid JSON, not Python data structure representations.  (Mostly a difference in quote marks used for strings.)

For the error messages, I wanted to continue using the strings with `str.format()`.  However, I couldn't find a clean way of using the left justification and padding that's available in the formatting micro-language.  If the technique I used isn't satisfactory, I can try for another approach.

To get the schema instance to display as true JSON, I thought encoding the schema back to JSON was the best approach.  Using newline with the separator and sorting keys seems to give close to the same output format as `pprint.pformat()`.  The difference is that it ends up being a little prettier, because there are always line breaks, but in my testing, `pprint` only broke the lines when the schema was longer than 72 characters.  I didn't test extensively, so I don't know how well the `json.dumps()` technique compares with `pprint` for lines that longer than 72 characters and aren't easily broken.

Please feel free to be critical with the code review.  I'm very much open to updating my changes to fit with the style of the project.


sloanlance:
Oops.  Obviously I forgot to update the tests.  I'll do that.  I'll convert this PR to draft until that's done.

Julian:
Awesome thanks! I will take a closer look at this (possibly after you get to the tests) but just to save you from one wrong turn -- the "use JSON" desire I had should *only* apply to the CLI, not to general formatting of the exception -- in other words, we can't change it within e.g. `ValidationError.__str__`, otherwise it will apply everywhere.

We have some options there. The least coupled is to just construct a custom way to format the errors within the CLI (using all the attributes present on a `ValidationError`, same as `ValidationError.__str__` does). But we also could factor out the body of `ValidationError.__str__` into a parametrized helper function, and then call it in 2 different ways from `ValidationError.__str__` and from the CLI.

But yeah want to save you from that otherwise you'll find you have to change lots of tests (and then find out later that I didn't mean that :D).

But yeah thanks for all you've done already! Seems like we're on the way here...

sloanlance:
Thanks for the info, @Julian.  I'll update the way the JSON gets formatted before addressing the tests.

sloanlance:
I've been thinking about the use of true JSON in this case.  When `_PrettyFormatter.validation_error()` gets `error`, if it just uses the output of `_Error.__unicode__()`, it could be tricky to convert all single quotes to double quotes before printing it.  Now the text may not have any single quotes other than in the JSON data structure's `repr`, but that could change in the future.

I'm thinking of adding a `format_error()` method to `_Error` that takes an optional Boolean argument, `pretty`, which is `False` by default.  When the  `_PrettyFormatter` code wants to use the error, rather than just letting Python call its `__str__` or `__unicode__` by default, it will call the error's `format_error()` method with `pretty=True`.

I think this is much like what you've described, but at the `_Error` class level instead of a specific exception type.  (And also in my own quirky way of wording it. 😉)

Julian:
Cool. Yes, think you're getting to the same place I was describing -- I'd probably want it to just be a private helper function, so consider something like:

```
class _Error:
    ....

    def __unicode__(self):
        return _formatted_error(self, to_str=lambda thing: pformat(thing, width=72))
```

and then in the CLI you'd use `_formatted_error(some_error, to_str=lambda thing: json.dumps(thing, indent=True, ...))`

sloanlance:
OK.  I see, passing in a lambda for formatting will make the code more flexible.  Would it be helpful to set `pformat()` as the default formatter in `_formatted_error()`?  Then when called from `__unicode__()`, we wouldn't need to specify the formatter.

Also, I'm going to look at the documentation about running the tests locally, but if you can tell me off the top of your head the command for testing with my local environment, that'd be helpful.  Tox is very cool, but it's more information than I need for initial testing.

Julian:
> OK. I see, passing in a lambda for formatting will make the code more flexible. Would it be helpful to set pformat() as the default formatter in _formatted_error()? Then when called from __unicode__(), we wouldn't need to specify the formatter.

Sounds reasonable!

> Also, I'm going to look at the documentation about running the tests locally, but if you can tell me off the top of your head the command for testing with my local environment, that'd be helpful. Tox is very cool, but it's more information than I need for initial testing.

Maybe this helps?:

```
python3.8 -m venv venv
venv/bin/python -m pip install ./ twisted
PYTHONPATH=./ venv/bin/trial jsonschema
```

(In theory you can replace `twisted` and `trial` there with any test runner, including pytest if you're more familiar with that, but the above is what I use.)

sloanlance:
@Julian, I'm trying to pass more of the CI checks.  I don't know whether you require 100% passing.  I made another commit that I think will satisfy the style checks that failed.

I wanted to know whether the pypy2 checks are important to you.  It's complaining about my use of unicode characters as an argument to `ljust()`, etc.  I have a possible solution in mind, but I think I'd need to do it blind.  (Make the change locally, then commit it and see if it passes CI checks.)

sloanlance:
I thought the changes I made, explicitly marking strings as Unicode, would do the job.  That solved the problems when using standard Python 2.7 (CPython) on my computer.  The CLI and the tests would work as expected.  (I also have Python 2.4, which would give errors that Unicode strings didn't have a `format()` method.)

So, I've installed pypy (for Python 2.7) via Homebrew.  When I run the CLI or tests with that interpreter, it gives the same error as the CI checks here, `TypeError: ljust() argument 2 must be a single character`.  In order to be Python 2 compatible, maybe I won't be able to use `ljust()` to make the Unicode header the desired length.

Julian:
Hey! Sorry, I haven't really gotten a chance to look at the error you were hitting -- let me know though if you're still having difficulty -- replacing ljust with just the manual calculation sounds fine to me, but if it continues to be an issue I can have a look (maybe tomorrow)

jsonschema/cli.py
```diff
@@ -70,18 +70,25 @@ def validation_success(self, **kwargs):
...
-    )
-    _SUCCESS_MSG = "===[SUCCESS]===({path})===\n"
+    @classmethod
+    def _json_formatter(cls, x):
+        return json.dumps(x, separators=(',\n', ': '), sort_keys=True)
+
+    def _simple_msg(self, path, type, header=False):
+        begin_end_chars = u'╒╕' if header is True else u'══'
+        return u'{}══[{}]═══({})'.format(begin_end_chars[0], type, path) \
+            .ljust(self._WIDTH - 1, u'═') + begin_end_chars[1]
```
sloanlance, line 82: What I changed was:

```suggestion
        return u'{}══[{}]═══({}){}'.format(begin_end_chars[0], type, path, u'═' * self._WIDTH)[:self._WIDTH - 1] \
            + begin_end_chars[1]
```

That resolved the errors when I run the CLI.  However, the output gets mangled by some other part of pypy2:

```text
âââ[ValidationError]âââ(<stdin>)ââââââââââââ
7 is less than the minimum of 9

Failed validating u'minimum' in schema:
    {"minimum": 9,
    "multipleOf": 3,
    "not": {"const": 7}}

On instance:
    7
ââââââââââââââââââââââââââââââââââââââââââââââââââââââââââââââ
```

I think this is specific to pypy2 because the code works without that change under CPython 2.7.

I haven't gotten the test runner to work under pypy2 yet.

I'm beginning to wonder whether this shouldn't be considered a known issue with pypy2.  The pretty output wouldn't be recommended for that interpreter and those tests would be skipped, too.

sloanlance: Ha!  This may be resolved by adding a `# -*- coding: UTF-8 -*-` line to the beginning of the file.  Checking on it…

sloanlance: Adding a "coding" declaration for the file resolves the problem on my computer when I run the CLI with the 2.* versions of PyPy and CPython.  I still don't have the tests working in my 2.* environments.  I probably need to add that declaration to the CLI tests file, too.

sloanlance:
Adding a "`coding`" declaration line to the beginning of `cli.py` handled much of the problems.  With that and the proper `u`-notation on strings containing Unicode characters, there were no more errors from `ljust()`, etc.

Now what remains appear to be problems with the `self._stderr.write()` and `self._stdout.write()` lines.  We get errors from the CI tests that "'unicode' does not have the buffer interface".  That's a surprise to me because I do **_not_** get that error when I run the CLI with v2.* of CPython/PyPy on my computer.  If that error were to occur, I'd expect it to happen when the CLI runs, not just during the test.

sloanlance:
What had been confusing me is that when I run the CLI with pretty output under CPython/PyPy 2.x, it succeeds.  However, the CI tests fail with the error about `unicode` not having buffer support.

I had trouble getting twisted installed in my PyPy environment, so I opted to use pytest instead.  I got the same error from that test harness.  Then I realized why this error happens.  The test harnesses, pytest and twisted trial, both capture stdout and stderr, which requires buffers.

Working around the problem in the test is not a good solution, because it's conceivable that someone may want to use jsonschema under Python 2 with stdout/stderr going to buffers.  They would get the same error.  So it seems the right thing to do is to update the code in cli.py to handle the output in a specific way under Python 2.  That would satisfy the tests and that use case.

(It would probably be better to check for whether `unicode` supports buffers in each scenario rather than checking for specific Python versions.)

I searched for other mentions of this error under Python 2 and pytest.  I was amused to find this:  https://github.com/tqdm/tqdm/issues/589

I don't have an exact solution in mind, but I understand the problem a bit better.

Julian:
Will have a look at this in a few hours I hope -- but are you sure this passes locally for you? It certainly fails here.

From glancing at the diff, it seems like you probably want native strings (i.e. no `u` prefix). `stdout`/`stderr` accepts native strings -- so str on both Py2 and Py3. Here it looks like you're giving it unicode on both, which yeah will blow up with that exception.

Will have some more specific guidance after having a look.

And thanks for the work so far it's much appreciated!

sloanlance:
> Will have a look at this in a few hours I hope -- but are you sure this passes locally for you? It certainly fails here.

No, it fails the tests via pytest for me, too.  However, if I run the CLI without the test harness (for example, using the input described in #708), it works perfectly.

> From glancing at the diff, it seems like you probably want native strings (i.e. no `u` prefix). `stdout`/`stderr` accepts native strings -- so str on both Py2 and Py3. Here it looks like you're giving it unicode on both, which yeah will blow up with that exception.

👍 

> And thanks for the work so far it's much appreciated!

🙇‍♂️

sloanlance:
This is very close to done now.  I believe all the functional tests are successful.  Even though they're not all done in CI here on GH, pytest on my local machine tells me all tests pass with python3, pypy3, and pypy2.

Now that it works in all those environments, I'm disappointed to see that the CI checks that have failed so far are for coverage and style.

* **_Style_**: I believe I've addressed the long line problems on my local machine, but before I commit them, I'd like to resolve the problem about `/Users/runner/work/jsonschema/jsonschema/jsonschema/cli.py:99:9: F811 redefinition of unused '_message_line' from line 85`.  Suggestions?  I was inspired to code it that way by the PY3 test I saw in the project's `exceptions.py`.  Should I do it a different way?
* **_Coverage_**: I'm not exactly sure of what to do to resolve the coverage problem.  It also involves the `_message_line()` method that is redefined for Python interpreters with weak Unicode support.

jsonschema/exceptions.py
```diff
@@ -61,14 +61,18 @@ def __init__(
...
-    def __unicode__(self):
+    def _formatted_message(self, formatter=None):
         essential_for_verbose = (
             self.validator, self.validator_value, self.instance, self.schema,
         )
         if any(m is _unset for m in essential_for_verbose):
             return self.message
 
-        pschema = pprint.pformat(self.schema, width=72)
+        if (callable(formatter)):
```
Julian, line 71: Would prefer to just always have formatter take a callable -- if this is just to handle the `None` case, just using `formatter=pformat` or `partial(pformat, width=72)` should make it go away I think?

jsonschema/cli.py
```diff
@@ -69,19 +70,49 @@ def validation_success(self, **kwargs):
 
 @attr.s
 class _PrettyFormatter(object):
+    _MESSAGE_BAR_CHAR = '═'
```
Julian, line 73: Minor: style-wise `jsonschema` uses double quotes (unless there's a double quote in the string, then single quotes).

jsonschema/cli.py
```diff
@@ -69,19 +70,49 @@ def validation_success(self, **kwargs):
...
+    _MESSAGE_CORNER_CHARS = ('╒', '╕')
```
Julian, line 74: Can ditch the parens here.

jsonschema/cli.py
```diff
@@ -69,19 +70,49 @@ def validation_success(self, **kwargs):
...
+    _MESSAGE_FORMAT = '{}══[{}]═══({})'
+    _MESSAGE_MAX_LENGTH = 79
+
+    @classmethod
+    def _json_formatter(cls, x):
```
Julian, line 79: This doesn't necessarily seem like it needs to be a classmethod -- any reason it shouldn't just be a regular function?

Also, rather than adding newlines to the separators, probably instead we just want `indent=True`?

jsonschema/cli.py
```diff
@@ -69,19 +70,49 @@ def validation_success(self, **kwargs):
...
+        return json.dumps(x, separators=(',\n', ': '), sort_keys=True)
 
-    _ERROR_MSG = dedent(
-        """\
-        ===[{type}]===({path})===
+    def _message_end_chars(self, header=False):
+        return self._MESSAGE_CORNER_CHARS if header is True else [self._MESSAGE_BAR_CHAR] * 2
```
Julian, line 83: Ideally I'd love to remove the boolean flag for header (and have separate paths for the header and the body lines, i.e. `_header_line` and `_body_line`) -- but as-is if we stick with what's here this should be equivalent to `self._MESSAGE_CORNER_CHARS if header else self._MESSAGE_BAR_CHAR * 2`, no?

jsonschema/cli.py
```diff
@@ -69,19 +70,49 @@ def validation_success(self, **kwargs):
...
+
+            return self._MESSAGE_FORMAT.format(begin_char, type, path) + \
+                self._MESSAGE_BAR_CHAR * bar_length + end_char
+
+
+    def _error_msg(self, path, type, body):
+        HEADER = self._message_line(path, type, header=True)
+        FOOTER = '└' + '─' * (self._MESSAGE_MAX_LENGTH - 2) + '┘'
+
+        return '\n'.join((HEADER, str(body), FOOTER, '\n'))
```
Julian, line 112: What type is body here? The `str` call looks suspicious

jsonschema/cli.py
Julian: Ah, I see ok, you're unpacking it below, so you needed the list for Py2, ok. (But yeah the `if True` is unnecessary).

Julian:
Cool!

Gave it a quick first pass and left some small comments.

> Style: I believe I've addressed the long line problems on my local machine, but before I commit them, I'd like to resolve the problem about /Users/runner/work/jsonschema/jsonschema/jsonschema/cli.py:99:9: F811 redefinition of unused '_message_line' from line 85. Suggestions? I was inspired to code it that way by the PY3 test I saw in the project's exceptions.py. Should I do it a different way?

Probably there's 2 decent options -- the manual way works on both versions no? If so we can likely just use that on both versions, and switch to using ljust later, rather than having 2 versions in the code.

Otherwise probably doing

```
if PY3:
    top_version
else:
    bottom_version
```

also may be a bit clearer and should make this go away.

> Coverage: I'm not exactly sure of what to do to resolve the coverage problem. It also involves the _message_line() method that is redefined for Python interpreters with weak Unicode support.

Yeah this we can likely ignore -- some day I'll implement combining coverage but I haven't gotten to it... So yes it does fail when code for just one testenv is introduced.

Julian:
Or choice C is I just merged the other PR with the Py2-dropping, since it'll make life easier I just can do that first :) -- sorry, probably should have done so before you were having trouble, but now hopefully that helps things a bit.

jsonschema/exceptions.py
sloanlance: Ah, so use that in place of the `formatter=None` argument in the declaration?

I put the `callable()` check there in case a non-callable formatter gets passed in.  It seems safer to me.

jsonschema/cli.py
sloanlance: Ah, sorry.  I hadn't noticed.  My personal habit is to use single quotes.  I'll update that.

jsonschema/cli.py
sloanlance: OK.  It passed through the style check.  Maybe it needs a config change?  I'll update this.

jsonschema/cli.py
sloanlance: I used `is True` to explicitly check for a Boolean value, rather than depend on a truth-y value.  But I can remove it.

sloanlance:
> Or choice C is I just merged the other PR with the Py2-dropping, since it'll make life easier I just can do that first :) -- sorry, probably should have done so before you were having trouble, but now hopefully that helps things a bit.

Excellent choice!  Sure, it would have been nice to avoid the Py2 annoyances I had with handling Unicode strings.  On the other hand, it's been a learning experience.

I'll remove the Py2 support I'd added and address the other things you commented about, then commit again later.

sloanlance:
@Julian, I just pulled upstream master into my branch to get the removal of Py2 support.  I didn't make any other changes yet.  I didn't look at all of them, but some of the CI checks seem to be failing during their set-up, before the tests even run.  Looks like the problem is during a `pip install`.  I know practically nothing about configuring tox, so maybe someone else should take a look at that.

jsonschema/cli.py
sloanlance: I think I had that there to address a problem with Py2.  I've removed it.

Julian:
The CI failures are https://github.com/python-hyper/hyperlink/issues/133 which is being resolved one way or another (either upstream or if not quickly then I'll shove a fix in here somehow)

jsonschema/exceptions.py
sloanlance: I'll replace `formatter=None` and the other logic with `formatter=partial(pformat, width=72)`, even though checking with `callable()` seems safer to me.  If you change your mind, tell me and I can put it back the other way.

jsonschema/cli.py
sloanlance: I could swear I replied about `@classmethod` earlier, but I don't see it here now.

If the method isn't annotated as `@classmethod` or `@staticmethod`, the tests fail.  If I use `@staticmethod`, the style tests tell me that methods should never be static, so I guess it needs to be `@classmethod`.

I agree about the JSON formatting.  I think I'd done it that way to make the `json.dumps()` output somewhat match the `repr` output that was used before.  The `indent` argument takes a number of spaces to indent, so I'll specify 4, to match the Python indentation style.  (JSON doesn't have an official standard indent size.)

Julian: Yes but why is it on the class at all, instead of just being a function?

Julian:
sorry, to be clearer about what I mean there, since maybe that was too short -- that function doesn't use anything relevant to an instance, or to the class itself, so it should just be a global function, completely unattached to the class (and then you won't need neither `@classmethod` nor `@staticmethod`)

sloanlance:
> …it should just be a global function, completely unattached to the class (and then you won't need neither `@classmethod` nor `@staticmethod`)

I see.  It makes sense to me for it to be a class method of `_PrettyFormatter` because it's not used anywhere else in the project.  I guess it has potential to be used elsewhere, though.

Which module do you recommend it be moved to?  `_util`, `_format`, or something else?

Julian:
Generally I'd just leave it right alongside `_PrettyFormatter` -- it's fine to have a function just used in one place, putting it next to the class is just as good as having it on it. As you say, it's possible it'll be used again, if it is, it can move to a module that makes sense for more locations.

sloanlance:
I just moved this PR from draft status to ready.  I should have done that yesterday when I left my [earlier comment](#issuecomment-668303741).

I believe I've addressed all the concerns raised during the code review.

On my local machine (macOS 10.14.5 with Python 3.8.3 for tests and Python 3.6.9/PyPy 7.3.1 for style), the tests all succeed and the style checks pass.  So I think the CI failures seen here are probably all due to the hyperlink problem mentioned before.

sloanlance:
I see there's more conflicts.  I'll pull from master and resolve them.

sloanlance:
All CI checks except coverage passed.  Automated coverage check isn't something I know much about, but I'd like to learn.  According to the [raw logs of the check](https://pipelines.actions.githubusercontent.com/b5J3bxR0zPJnc2fMMsqJBsvVM9U2Zzkt8obsYHRuV1GYDmdUgd/_apis/pipelines/1/runs/733/signedlogcontent/3?urlExpires=2020-08-07T11%3A25%3A19.0989666Z&urlSigningMethod=HMACV1&urlSignature=vllM78%2FCBJSP0naf532Z2sYWXLs75NVoM1hxJ7gBFiE%3D), it looks like a configuration file for running the check is missing.  Is that right?

I saw a [coverage report](#issuecomment-665256587) from closer to the time I opened this PR.  There have been several changes to my code since then, so it's probably out of date.  Is there any value to using that report?  I'm unsure how to interpret the report anyway.  🙈

Julian:
The failure there is codecov.io (which is the free service for hosting those reports) being flaky, likely nothing you did/changed.

I just re-ran the job, it's unfortunately common for it to barf. Let's see if it passes this time.

sloanlance:
Yes, they ran this time.  The only failure was to hit the 100% coverage target.  Looks like there's an overall slight increase in coverage.  Does this satisfy your requirements for merging, or would you like me to look into increasing coverage for code I added?

Julian:
Yeah don't mind the coverage being confused not sure why codecov.io has begun to do that as well recently but you're fine there -- will give this another review I hope later today!

jsonschema/exceptions.py
```diff
@@ -61,14 +61,18 @@ def __repr__(self):
...
             self.validator, self.validator_value, self.instance, self.schema,
         )
         if any(m is _unset for m in essential_for_verbose):
             return self.message
 
-        pschema = pprint.pformat(self.schema, width=72)
-        pinstance = pprint.pformat(self.instance, width=72)
+        pschema = formatter(self.schema)
+
+        pinstance = pformat(self.instance, width=72)
```
Julian, line 75: We're only using the formatter for the schema, not the instance here, so there's going to be inconsistency in the output.

Can we add a test that covers the instance too and then fix this to make sure the formatter is used for both?

jsonschema/tests/test_cli.py
```diff
@@ -1,12 +1,13 @@
...
-from json import JSONDecodeError
-from textwrap import dedent
-from unittest import TestCase
+# -*- coding: UTF-8 -*-
 import errno
 import json
 import os
 import subprocess
 import sys
+from io import StringIO
```
Julian, line 7: Can you put these imports back in the order they were at the top?

jsonschema/cli.py
```diff
@@ -1,14 +1,14 @@
...
 """
 
-from json import JSONDecodeError
-from textwrap import dedent
 import argparse
 import errno
 import json
 import sys
 import traceback
+from json import JSONDecodeError
```
Julian, line 11: Can you put this import back at the top?

jsonschema/cli.py
```diff
@@ -66,21 +66,32 @@ def validation_success(self, **kwargs):
...
+    _MESSAGE_BAR_CHAR = "═"
+    _MESSAGE_CORNER_CHARS = "╒", "╕"
+    _MESSAGE_FORMAT = "{}══[{}]═══({})"
+    _MESSAGE_MAX_LENGTH = 79
 
-    _ERROR_MSG = dedent(
-        """\
-        ===[{type}]===({path})===
+    def _message_line(self, path, type, header=False):
+        begin_char, end_char = self._MESSAGE_CORNER_CHARS if header \
```
Julian, line 81: Minor style: jsonschema follows PEP8 / generally avoids using backslashes for continuation. Can you switch these to use parentheses instead?

Though, I think this will be a lot clearer if we split it into `_header_line` and `_non_header_line` -- can you give that a shot? It should remove a bunch of the conditional behavior here.

jsonschema/exceptions.py
```diff
@@ -1,16 +1,16 @@
 """
 Validation errors, and some surrounding helpers.
 """
-from collections import defaultdict, deque
 import itertools
-import pprint
 import textwrap
+from collections import defaultdict, deque
```
Julian, line 6: Can you move these imports to the top?

jsonschema/exceptions.py
```diff
@@ -1,16 +1,16 @@
...
+from functools import partial
+from pprint import pformat
 
 import attr
 
 from jsonschema import _utils
 
-
```
Julian, line 13: Can you put this newline back?

jsonschema/tests/test_cli.py
```diff
@@ -309,17 +310,18 @@ def test_invalid_schema_pretty_output(self):
 
         with self.assertRaises(SchemaError) as e:
             validate(schema=schema, instance="")
-        error = str(e.exception)
+        error = str(e.exception._formatted_message(
```
Julian, line 313: Minor style: jsonschema generally uses this style for continuation:

```suggestion
str(
    e.exception._formatted_message(formatter=cli._json_formatter),
)```

Julian:
In case it helps, I added a pre-commit config for you/others so you should be able to fix some of the style comments here by just running pre-commit (which will e.g. restore the import order).

jsonschema/cli.py
sloanlance: I usually don't like backslashes, either.  I got in the habit of using it because my coworkers do like them.  I'll change this code.

sloanlance:
> In case it helps, I added a pre-commit config for you/others so you should be able to fix some of the style comments here by just running pre-commit (which will e.g. restore the import order).

What is the definition of the order of imports preferred for this project?  I often use IntelliJ IDEA's reformatting feature.  In addition to cleaning up lines of code, it reorganizes imports.  Off the top of my head, I think the order is something like sections of core Python modules, third-party modules, and project modules, each section alphabetized.

Julian:
> What is the definition of the order of imports preferred for this project? I often use IntelliJ IDEA's reformatting feature. In addition to cleaning up lines of code, it reorganizes imports. Off the top of my head, I think the order is something like sections of core Python modules, third-party modules, and project modules, each section alphabetized.

It's exactly that in fact :), just actually alphabetized (some tools, e.g. `isort` by default, and it sounds like IntelliJ), will put from imports on the bottom rather than on the top, whereas `"from" < "import"`.

Issue #710, opened 2020-07-22
Schema with $id / id does not have the id automatically cached for $ref resolution


FuhuXia:
We have an all-in-one [schema file](https://github.com/GSA/ckanext-datajson/blob/master/ckanext/datajson/pod_schema/federal-v1.1/dataset.json#L722-L724), within which the property `subOrganizationOf` is referencing to its parent `organization` by `"$ref": "#"`. But when validators.py hits this $ref, it tries to follow the url in the id and fetch external url https://project-open-data.cio.gov/v1.1/schema/organization.json. If the url is out-dated, it raise RefResolutionError(exc).

We are expecting validators stays with local definition, not relies on external url to validate this recursive self-referenced definition.

FuhuXia:
sample.json:
```
{
  "organization": {
    "name": "Data.gov",
    "subOrganizationOf": {
      "name": "Technology Transformation Service"
    }
  }
}
```
sample.schema:
```
{
  "$schema": "http://json-schema.org/draft-04/schema#",

  "definitions": {
    "organization": {
      "id": "https://project-open-data.cio.gov/v1.1/schema/organization.json#",
      "required": [
        "name"
      ],
      "properties": {
        "name": {
          "type": "string"
        },
        "subOrganizationOf": {
          "$ref": "#"
        }
      }
    }
  },

  "properties": {
    "organization": {"$ref": "#/definitions/organization"}
  }
}
```
Run the following command and here is what we see.
`$ jsonschema -i sample.json sample.schema` 

1. It is validated but jsonschema actually makes an (unnecessary?) external request to the URI string 'https://project-open-data.cio.gov/v1.1/schema/organization.json#'.
2. If we change the URI string slightly, it gives error `jsonschema.exceptions.RefResolutionError: HTTP Error 404: Not Found`
3. We can remove the `id` field from the schema, but then thw `name` property requirement is not being validated any more.

We expect the jsonschema treats the `id` field as unique identifier, not follows its URI to fetch definition remotely, since the definition is in the local schema already.

willson-chen:
@FuhuXia The PR #717 could solve your problem.

FuhuXia:
@willson-chen
I installed v3.2.0 and cherry-picked the three commits from the PR #717. It still has same issue.

willson-chen:
[...]

Issue #693, opened 2020-05-26
Make it easier to create ref resolvers that don't do any remote lookup


jaredgrubb:
I enjoy using the `$ref` stuff (and `$id`) but I really never want my validators to _actually_ access the network. I used a RefResolver instance to map some schemas together, but I can't  figure out how to permanently disable actual network access (and only use the  local store provided). 

I looked through the sources  and didn't see anything obviously.  If this is already possible, maybe  it'd be worth a FAQ entry or something?

Julian:
It's possible, though not easily and not in a great way. Doing so more easily is part of #346. Happy to keep this open as a way to track the request individually though, it's a common request to make this easier.

willson-chen:
@jaredgrubb Maybe the PR #717 could solve your problem.

Julian:
Hello there!

This, along with many many other `$ref`-related issues, is now finally being handled in #1049 with the introduction of a new [referencing library](https://github.com/python-jsonschema/referencing/) which is fully compliant and has APIs which I hope are a lot easier to understand and customize.

The next release of `jsonschema` (v4.18.0) will contain a merged version of that PR, and should be released shortly in beta, and followed quickly by a regular release, assuming no critical issues are reported.

The new APIs do not have this bad behavior -- they don't make HTTP requests by default!

If you still care to, I'd love it if you tried out the beta once it is released, or certainly it'd be hugely helpful to immediately install the branch containing this work (https://github.com/python-jsonschema/jsonschema/tree/referencing) and confirm. You can in the interim find documentation for the change in a [preview page here](https://python-jsonschema--1049.org.readthedocs.build/en/1049/referencing/).

I'm going to close this given it indeed seems like it is addressed by #1049, but feel free to follow up with any comments. Sorry for the delay in getting to these, but hopefully this new release will bring lots of benefit!

PR #717, opened 2020-08-05
Add id with ref support, solve bug(371).

willson-chen:


jsonschema/validators.py
```diff
@@ -862,6 +868,44 @@ def resolve_remote(self, uri):
...
+
+            last_schema:
+
+                The referring upper level schema.
+
+            last_url:
+
+                Save the last URL.
+        """
+        if not isinstance(schema, dict) \
```
willson-chen, line 889: Keep  ```"id" : "http://json-schema.org/draft-03/schema#", "id" : "http://json-schema.org/draft-04/schema#", "id" : "http://json-schema.org/draft-06/schema#", "id" : "http://json-schema.org/draft-07/schema#"```  unchanged in self.store

jsonschema/validators.py
```diff
@@ -862,6 +868,44 @@ def resolve_remote(self, uri):
...
+                or self.resolution_scope in list_schema:
+            return
+
+        for k in schema.keys():
+            if k in [u"id", u"$id"] and isinstance(schema[k], str_types):
+
+                last_url = urljoin(last_url, schema[k], allow_fragments=True)
```
willson-chen, line 896: Splice the value and the url of the upper level, and save it into last_url at the same time.

jsonschema/validators.py
```diff
@@ -862,6 +868,44 @@ def resolve_remote(self, uri):
...
+                url, fragment = urldefrag(last_url)
+
+                if last_schema:
+                    self.store[url] = last_schema
```
willson-chen, line 900: Save the subschema of the upper level into self.store in the form of `"url": { last_schema }`

jsonschema/validators.py
```diff
@@ -862,6 +868,44 @@ def resolve_remote(self, uri):
...
+
+                if fragment:
+                    self.store[url][fragment] = schema
```
willson-chen, line 903: Add fragment element and save it into `url": { last_schema }` which saved above, in the form of  `"url": { last_schema, "fagment": { schema } }`

jsonschema/validators.py
```diff
@@ -862,6 +868,44 @@ def resolve_remote(self, uri):
...
+
+            backup_schema = copy.deepcopy(schema)
+            if isinstance(schema[k], dict):
+                self.store_subschema(schema[k], backup_schema, last_url)
```
willson-chen, line 907: Recursively traverse the dictionary.

jsonschema/validators.py
```diff
@@ -862,6 +868,44 @@ def resolve_remote(self, uri):
...
+                last_url = urljoin(last_url, schema[k], allow_fragments=True)
+                url, fragment = urldefrag(last_url)
+
+                if last_schema:
+                    self.store[url] = last_schema
+
+                if fragment:
+                    self.store[url][fragment] = schema
+
+            backup_schema = copy.deepcopy(schema)
```
willson-chen, line 905: Deep copy prevents references.

jsonschema/validators.py
willson-chen: @Julian This if statement block is replaced with the following statement, there will be no problem with python3 and above， but python2 will report  error `exceptions.AttributeError: NoneType' object has no attribute 'find'`
```
if not isinstance(schema, dict) or schema.get(u"$schema") in meta_schemas:
     return
```
In order to maintain compatibility with python2, I keep the current if statement block

willson-chen:
@Julian Hello, I have tested this PR could solve the following four skipped scenes，at the same time #710 #693 #341 all have been solved.
```
        or skip(
            message=bug(),
            subject="ref",
            case_description="Recursive references between schemas",
        )(test)
        or skip(
            message=bug(371),
            subject="ref",
            case_description="Location-independent identifier",
        )(test)
        or skip(
            message=bug(371),
            subject="ref",
            case_description=(
                "Location-independent identifier with absolute URI"
            ),
        )(test)
        or skip(
            message=bug(371),
            subject="ref",
            case_description=(
                "Location-independent identifier with base URI change in subschema"
            ),
        )(test)
```

WilliamHeaven:
@Julian This PR can fix some skipped test cases. Could you please check it to see whether it meets the expectation when you have time?

jsonschema/validators.py
```diff
@@ -862,6 +864,40 @@ def resolve_remote(self, uri):
...
+
+            schema:
+
+                The referring schema.
+
+            last_url:
+
+                The last URL.
+        """
+        if not isinstance(schema, dict):
```
Julian, line 881: Will have a closer look, but the schema does not have to be a dict (it's an [arbitrary collections.Mapping](https://github.com/Julian/jsonschema/blob/main/jsonschema/validators.py#L933)) and does not necessarily even need to be copy-able.

jsonschema/validators.py
```diff
@@ -862,6 +864,40 @@ def resolve_remote(self, uri):
...
+            return
+
+        for k in schema.keys():
+            if k in [u"id", u"$id"] and isinstance(schema[k], str):
```
Julian, line 885: This should use `Validator.ID_OF` in some way.

jsonschema/validators.py
```diff
@@ -862,6 +864,40 @@ def resolve_remote(self, uri):
             self.store[uri] = result
         return result
 
+    def store_subschema(self, schema, last_url=None):
```
Julian, line 867: To start I'd not want this to be a public method / addition to the API.

Julian:
Awesome! Left 3 quick comments from a first read on the approach. Will have a closer read but let me know if any of those seem unclear.

WilliamHeaven:
@willson-chen I think this function is useful. Could you please modify it based on the review comments?

PR #718, opened 2020-08-06
fix hostname bug

willson-chen:
Fix skipped test scenarios where the host name ends with character

### DIFF
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -209,7 +209,9 @@ if hasattr(socket, "inet_pton"):
         return socket.inet_pton(socket.AF_INET6, instance)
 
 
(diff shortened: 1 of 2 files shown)


Julian:
Thanks! This is a great thing to fix!

Though I'd much prefer if we found something that implemented this validation -- that regex is hairy and won't be easy to maintain -- could you possibly take a quick look to see if there's a suitable library function somewhere that may implement host name validation?

willson-chen:
> -- could you possibly take a quick look to see if there's a suitable library function somewhere that may implement host name validation?

@Julian I have found the fqdn library that can achieve the purpose of validating hostname, please see PR #719 for specific implementation, if no problem i will close this PR.

Julian:
Cool! Yep definitely looks worth continuing this on the other PR. Thanks again!

PR #719, opened 2020-08-07
fix bug about hostname by import fqdn

willson-chen:
The fdqn module is used to replace the regular implementation of PR #718

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -9,6 +9,7 @@ babel==2.8.0              # via sphinx
 certifi==2020.4.5.1       # via requests
 chardet==3.0.4            # via requests
 docutils==0.16            # via sphinx
+fqdn==1.4.0               # via requests
 idna==2.9                 # via requests
 imagesize==1.2.0          # via sphinx
 jinja2==2.11.2            # via sphinx
@@ -31,7 +32,6 @@ sphinxcontrib-qthelp==1.0.3  # via sphinx
 sphinxcontrib-serializinghtml==1.1.4  # via sphinx
 sphinxcontrib-spelling==5.0.0  # via -r requirements.in
 urllib3==1.25.9           # via requests
-fqdn==1.4.0               # via requests
 
 # The following packages are considered to be unsafe in a requirements file:
 # setuptools
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -31,6 +31,7 @@ sphinxcontrib-qthelp==1.0.3  # via sphinx
 sphinxcontrib-serializinghtml==1.1.4  # via sphinx
 sphinxcontrib-spelling==5.0.0  # via -r requirements.in
 urllib3==1.25.9           # via requests
+fqdn==1.4.0               # via requests
 
 # The following packages are considered to be unsafe in a requirements file:
 # setuptools
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -6,12 +6,9 @@
 #
 alabaster==0.7.12         # via sphinx
 babel==2.8.0              # via sphinx
-cached-property==1.5.1    # via fqdn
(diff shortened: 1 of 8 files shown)


docs/requirements.txt
```diff
@@ -31,6 +31,7 @@ sphinxcontrib-qthelp==1.0.3  # via sphinx
 sphinxcontrib-serializinghtml==1.1.4  # via sphinx
 sphinxcontrib-spelling==5.0.0  # via -r requirements.in
 urllib3==1.25.9           # via requests
+fqdn==1.4.0               # via requests
```
Julian, line 34: Did you add this here, or was it added by running `pip-compile`?

jsonschema/_format.py
```diff
@@ -4,6 +4,7 @@
 import struct
 
 from jsonschema.exceptions import FormatError
+from fqdn import FQDN
```
Julian, line 7:
```suggestion
import struct

from fqdn import FQDN

from jsonschema.exceptions import FormatError
```

setup.cfg
```diff
@@ -37,12 +37,14 @@ format =
     rfc3987
     strict-rfc3339
     webcolors
+    fqdn
```
Julian, line 40: Can you sort these two alphabetically into the list?

docs/requirements.txt
willson-chen: The CI is reproting an error if I'm not adding this.
The error information is as follows:
```
docs-html run-test-pre: PYTHONHASHSEED='2481213764'
docs-html run-test: commands[0] | /Users/runner/work/jsonschema/jsonschema/.tox/docs-html/bin/python -m sphinx -b html /Users/runner/work/jsonschema/jsonschema/docs/ /Users/runner/work/jsonschema/jsonschema/.tox/docs-html/tmp/build -a -n -q -T -W

  File "/Users/runner/work/jsonschema/jsonschema/.tox/docs-html/site-packages/sphinx/config.py", line 319, in eval_config_file
    execfile_(filename, namespace)
  File "/Users/runner/work/jsonschema/jsonschema/.tox/docs-html/site-packages/sphinx/util/pycompat.py", line 81, in execfile_
    exec(code, _globals)
  File "/Users/runner/work/jsonschema/jsonschema/docs/conf.py", line 10, in <module>
    import jsonschema
  File "/Users/runner/work/jsonschema/jsonschema/.tox/docs-html/site-packages/jsonschema/__init__.py", line 14, in <module>
    from jsonschema._format import (
  File "/Users/runner/work/jsonschema/jsonschema/.tox/docs-html/site-packages/jsonschema/_format.py", line 7, in <module>
    from fqdn import FQDN
ModuleNotFoundError: No module named 'fqdn'
````
The link is as follows:
[error](https://github.com/willson-chen/jsonschema/runs/956671609?check_suite_focus=true)

Comment (Julian) [reply docs/requirements.txt]: Got it yeah so the docs will now depend on fqdn, but that file is autogenerated (by running pip-compile), which will automatically overwrite the file the next time someone runs it.

If you `pip install pip-tools` and then run `pip-compile`, you should get a properly regenerated requirements.txt.

Comment (willson-chen) [reply docs/requirements.txt]: Thank you very much for your comments, I have run the `pip-compile` command to regenerate the requirements.txt.
The following content is added:
```
cached-property==1.5.1    # via fqdn
colorama==0.4.3           # via sphinx
fqdn==1.4.0               # via -r requirements.in
```

willson-chen:
@Julian I have revised all points based on the review comments, and I hope you will take the time to check them.

Julian:
It looks like I sent you on a bit of a wild goose chase with the requirements file, I didn't notice what you were saying there until just now. Explaining the right way forward was actually a bit tricky though. If you're curious you can look at the commits I put on your branch to fix it another way.

Otherwise, think this is ready to merge.

Appreciated!

PR #721, opened 2020-08-13
Fix issue #669 and add test case.

willson-chen:


WilliamHeaven:
Add test cases based on PR #680  and fix the base_uri layer problem. 

The code before the rectification is as follows: 
```
os.path.abspath(arguments["schema"])) 
```
Code after the rectification: 
```
os.path.abspath(os.path.dirname(arguments["schema"]))
```

willson-chen:
Add test cases based on PR #680  and fix the base_uri layer problem. 

The code before the rectification is as follows: 
```
os.path.abspath(arguments["schema"])) 
```
Code after the rectification: 
```
os.path.abspath(os.path.dirname(arguments["schema"]))
```

Julian:
Thanks for trying to tackle this one as well. Have a look at the comment [here](https://github.com/Julian/jsonschema/pull/680#issuecomment-615903589) from that PR -- specifically, I'd like to support this with a `--base-uri` flag instead, and if someone wants to use the working directory as that base URI, you'd specify `--base-uri .` essentially.

willson-chen:
@Julian Sorry, I ignored some of the details in the comment #680, now I have replaced -r with --base-uri.

jsonschema/cli.py
```diff
@@ -178,6 +179,15 @@ def _namedAnyWithDefault(name):
         of the class.
     """,
 )
+parser.add_argument(
+    "--base-uri",
+    action="store_true",
```
Julian, line 184: Thanks for changing this -- the idea here though was that this could/should be a URI, not just a flag.

I.e. you should say `--base-uri .` (note the `.`) if you mean "use the current directory as the base URI". But you could equally well pick any other directory as your base URI if you'd like as well.

jsonschema/tests/test_cli.py
```diff
@@ -683,6 +683,23 @@ def test_successful_validation__of_just_the_schema_pretty_output(self):
             stderr="",
         )
 
+    def test_successful_validation_with_specifying_base_uri(self):
+        schema = """\
+                {"type": "object", "properties": {"KEY1":
+                {"$ref": "schema.json#definitions/schemas"}},
+                "definitions": {"schemas": {"type": "string"}}}
+                """
+        fp = open("schema.json", "w+")
```
Julian, line 692: Can we use `tempfile.NamedTemporaryFile` here, bit clearer, and doesn't depend on the current directory.

Also, as-is this would leak a file onto the disk if the test failed somewhere in the middle, another thing the above should fix.

willson-chen: Can we add a schema.json file in the tests directory like tests/schema.json, or in other directory, because I found that if `tempfile.NamedTemporaryFile` is not closed in windows, there will be permission problems if it is opened again, 
so I have to close it before using, at this time, the file will disappear. If the  `delete` option  is set to `False`, the generated temporary files still need to be deleted manually.

jsonschema/cli.py
willson-chen: Ok, I have modified this part of the function. After the temporary file problem is modified, push together!

jsonschema/tests/test_cli.py
Julian: You can do those via:

```
schema_file = tempfile.NamedTemporaryFile(delete=False)
self.addCleanup(os.remove, schema_file.name)
```

or thereabouts.

willson-chen:
@Julian I have already pushed the repair part, comments are welcome.

jsonschema/cli.py
```diff
@@ -251,7 +260,24 @@ def load(_):
                 raise _CannotLoadFile()
         instances = ["<stdin>"]
 
-    validator = arguments["validator"](schema)
+    if arguments["base_uri"] is None:
+        resolver = None
+    elif "http:" in arguments["base_uri"] or "https:" in arguments["base_uri"]\
```
Julian, line 265: This behavior doesn't look like it's tested, but also it doesn't seem exactly correct.

There are many schemes (many URIs) that may be provided here -- jsonschema allows any scheme, including user defined ones, and will retrieve them today via `requests`, which itself allows defining handlers for any scheme someone would like.

But the behavior here shouldn't differ across schemes -- what's making us split file out?

jsonschema/cli.py
```diff
@@ -251,7 +260,24 @@ def load(_):
...
+            or "urn:" in arguments["base_uri"]:
+        resolver = RefResolver(
+            base_uri=arguments["base_uri"],
+            referrer=schema,
+        )
+    else:
+        file_prefix = "file:///{}/" if "nt" == os.name else "file://{}/"
+        resolver = RefResolver(
+            base_uri=file_prefix.format(
+                os.path.abspath(arguments["base_uri"])
```
Julian, line 275: Is it this that's causing file to be split out separately? We shouldn't call `abspath` ourselves unless we have good reason, but we should think it through to find out. If the person using the CLI wants an absolute path, why can't they just pass one?

willson-chen:
We can specify a **remote** or **local** path through `--base-uri`, where the **local** path can be a **relative** path or an **absolute** path. Through `scheme = urllib.parse.urlsplit(arguments["base_uri"]).scheme`, if the **scheme** is an empty string or a drive letter identifier from 'a' to 'z' is considered to be a **local** path, otherwise it is considered a **remote** path.
@Julian How about this solution?

Julian:
I'd have to look more carefully myself but that doesn't sound right to me either I think.

I don't really want to change any default behaviors of things (what behavior folks get should match what behavior they get from normal other things).

Or, said differently, I don't want to maintain code that's trying to guess what kind of URI was provided as the base URI.

The base URI as provided needs to be suitable for use exactly as is, or be unambiguous.

Will leave a quick comment on the PR in the section that looks "suspicious" to me, let me know what you think.

jsonschema/cli.py
```diff
@@ -252,7 +262,34 @@ def load(_):
                 raise _CannotLoadFile()
         instances = ["<stdin>"]
 
-    validator = arguments["validator"](schema)
+    if arguments["base_uri"] is None:
+        resolver = None
+    else:
+        scheme = urllib.parse.urlsplit(arguments["base_uri"]).scheme
```
Julian, line 268: There's way too much code here. What goes wrong when this whole else block disappears? (And when the value as given on the command line is used exactly as-is?)

willson-chen:
@Julian Modified to the following simple implementation, let users pass in the correct URI, for example `--base-uri file:///c:/` `--base-uri http://example`, also supports `--base-uri .`
Is this implementation possible?
```
    if arguments["base_uri"] is None:
        resolver = None
    elif arguments["base_uri"] == ".":
        file_prefix = "file:///{}/" if "nt" == os.name else "file://{}/"
        resolver = RefResolver(
                    base_uri=file_prefix.format(os.getcwd()),
                    referrer=schema,
                )
    else:
        resolver = RefResolver(
                        base_uri=arguments["base_uri"],
                        referrer=schema,
                    )
```

jsonschema/cli.py
```diff
@@ -252,7 +261,21 @@ def load(_):
                 raise _CannotLoadFile()
         instances = ["<stdin>"]
 
-    validator = arguments["validator"](schema)
+    if arguments["base_uri"] is None:
```
Julian, line 264: I think GitHub somehow ate my comment here, so rewriting it in case you didn't see it -- but, basically I'm hoping the entirety of this change should be `resolver = RefResolver(base_uri=arguments["base_uri"], referrer=schema) if arguments["base_uri"] is not None else None` (or thereabouts).

What goes wrong / what functionality is missing if that's what we have?

willson-chen: > I think GitHub somehow ate my comment here, 

Maybe it's a bug of GitHub. Let's ignore it and keep looking at this PR.

>so rewriting it in case you didn't see it -- but, basically I'm hoping the entirety of this change should be `resolver = RefResolver(base_uri=arguments["base_uri"], referrer=schema) if arguments["base_uri"] is not None else None` (or thereabouts).
> 
> What goes wrong / what functionality is missing if that's what we have?

If here is changed to `resolver = RefResolver(base_uri=arguments["base_uri"], referrer=schema) if arguments["base_uri"] is not None else None`, relative paths, such as `--base-uri .`, cannot be supported. If `--base-uri .` and other relative paths are not considered, there will be no problems.

willson-chen:
@Julian I have modified it according to your last review, could you please take a look again?

jsonschema/tests/test_cli.py
```diff
@@ -683,6 +684,73 @@ def test_successful_validation_of_just_the_schema_pretty_output(self):
...
         )
 
+    def test_successful_validate_with_specifying_base_uri_absolute_path(self):
+        absolute_path = os.getcwd()
+        try:
+            schema_file = tempfile.NamedTemporaryFile(
+                mode='w+',
+                prefix='schema',
+                suffix='.json',
+                dir=absolute_path,
```
Julian, line 694: Not sure I follow why we're specifying all these options to `NamedTemporaryFile`. If you use `tempfile.NamedTemporaryFile(delete=False)` it should just give you an (absolute) path to a temporary file.

jsonschema/tests/test_cli.py
```diff
@@ -683,6 +684,73 @@ def test_successful_validation_of_just_the_schema_pretty_output(self):
...
+                delete=False
+            )
+            self.addCleanup(os.remove, schema_file.name)
+            schema = """
+                    {"type": "object", "properties": {"KEY1":
```
Julian, line 699: Can we simplify the schema we're using to test this behavior? It's a bit hard to follow what's happening with all the string formatting. (I don't understand why we need to format in `"\""`, but even more than that is a bit hard to follow).

We should in theory be able to test this behavior by having a schema with one single validator, a `$ref` with a relative reference, and then we ensure that that relative reference was used when resolving -- e.g. say, a temporary directory with two files, bar with `{"$ref": "foo.json#definitions/num"}` and then another, `foo.json`, with `{"definitions": {"num": {"type": "integer"}}`, and then setting `--base-uri` to the enclosing directory already does it, no?

That'd seem a bit clearer, but if I'm just not seeing how this test is working maybe if you remove some of the unneeded properties in this schema it'd be clearer.

jsonschema/tests/test_cli.py
```diff
@@ -683,6 +684,73 @@ def test_successful_validation_of_just_the_schema_pretty_output(self):
...
+            argv=[
+                "-i", "some_instance",
+                "--base-uri", absolute_path,
+                "some_schema",
+            ],
+            stdout="",
+            stderr="",
+        )
+
+    def test_failure_validate_with_specifying_base_uri_absolute_path(self):
```
Julian, line 720: We should have a test for specifying the wrong base URI as well, and possibly one for an invalid base URI.

willson-chen:
@Julian Sorry, I have been away for a while, but I never forget that there is still something unfinished. Is it possible to modify as follows? https://github.com/Julian/jsonschema/pull/721/commits/1bfbb9064c0c2da5df5ea0f7ed03623f8418f278

Julian:
Hey! No worries (trust me I understand the busyness...)

From a quick look that definitely looks more like what I was expecting! Will have a closer look and leave any comments but think we may be close.

willson-chen:
@Julian Can this PR be merged? Ask a rather abrupt question, do you welcome new maintainers to take care of the community. I am willing to participate if needed.

Julian:
Apologies for the delay again here...

There were a few things I wanted to write up comments on which I didn't get a chance to, so I just fixed them in a few commits.

I think this should be good to merge now as soon as CI passes.

On your offer, 100%! I definitely could use more help. If you see pull requests from other folks that aren't yet fully mergeable, or waiting themselves for review, or issues that could use triaging, etc. it would be super super appreciated.

RomainTT:
A year passed and I did not thank you both yet !

Thank you @willson-chen and @Julian for finishing this work! That is a very useful feature.

Issue #655, opened 2020-02-11
email and hostname format validation fails


davaya:
I disagree with the rationale used to close #403: "validating email addresses isn't really a reasonable concept".  If the authors of JSON Schema believed that, they would not have created an email format.   There is significant discussion about the *intent* of validating that a recipient actually exists at a particular email address vs. validating syntactic conformance with RFC 6854, but the goal of any schema is to validate syntax.  Saying "not a reasonable concept" disrespects the existence of RFCs that define the syntax required for interoperability.  I appreciate that handrews participated in the discussion of format as a second class citizen, but some users clearly need for it to behave as a first class assertion in order to implement the purpose of https://tools.ietf.org/html/draft-handrews-json-schema-validation-02#section-7.1:
```
   The "format"
   annotation keyword is defined to allow schema authors to convey
   semantic information for a fixed subset of values which are
   accurately described by authoritative resources, be they RFCs or
   other external specifications.
```

I suggest having a look at https://stackoverflow.com/questions/201323/how-to-validate-an-email-address-using-a-regular-expression.

```
        try:
            jsonschema.validate(
                instance=val,
                schema={'type': 'string', 'format': name},
                format_checker=jsonschema.draft7_format_checker)
        except jsonschema.exceptions.ValidationError as e:
            raise ValueError(e.message)
        return val
```
When calling jsonschema 3.2.0 with jsonschema.draft7_format_checker, some invalid hostname and email strings are rejected but others are accepted.   The Checker Notes table indicates that no additional packages are required for date, email, hostname, ipv4, or regex formats.

ValidationError is raised for email format on `"http://www.foo.com/index.html"` and `"john"`, but is not raised for `"john@"`.

ValidationError is raised for hostname format on `"_http._sctp.www.example.com"` (underscores are allowed in DNS service names but not in hostnames), but is not raised on `"tag-.example.com"` (label cannot begin or end with a hyphen.

https://jsonschemalint.com/#!/version/draft-07/markup/json correctly rejects the above bad email and hostname examples.  The jsonschemalint authors apparently have a different opinion about the "reasonableness" of performing syntax validation on email addresses and hostnames.  I submit that returning a boolean answer for an authoritatively-defined fixed subset of values is both possible and reasonable.  My code does it, including for types other than string, but I wanted to delegate JSON Schema defined formats to your package.  That's when the unit tests started failing.

Julian:
Hi.

[...]

PR #723, opened 2020-08-15
Add validate email by regex.

willson-chen:
Using regex RFC 5322 python by [link](https://www.codegrepper.com/code-examples/csharp/regex+RFC+5322+python).

### DIFF
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -174,12 +174,27 @@ def _checks_drafts(
     return wrap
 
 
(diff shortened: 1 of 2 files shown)


Julian:
Thanks again.

This specifically is the one thing I don't want (or intend) to add.

Have a look at e.g. https://github.com/Julian/jsonschema/issues/655#issuecomment-584846854 for the reasoning, there are some previous issues in the backlog as well.

If someone wants this, they can easily plug it in themselves.

Do appreciate it though!

PR #725, opened 2020-08-20
Fix cli bug about getting arguments["validator"].

willson-chen:
As commented in https://github.com/Julian/jsonschema/issues/710#issuecomment-676851156.
By adjusting the position of `arguments["validator"]`, we can solve this bug, but because `arguments["validator"]` is not processed in `parse_args`, it will cause the `test_latest_validator_is_the_default` test case to fail, so I deleted this test case.

### DIFF
--- a/jsonschema/cli.py
+++ b/jsonschema/cli.py
@@ -191,8 +191,6 @@ parser.add_argument(
 
 def parse_args(args):
     arguments = vars(parser.parse_args(args=args or ["--help"]))
-    if arguments["validator"] is None:
-        arguments["validator"] = validator_for(arguments["schema"])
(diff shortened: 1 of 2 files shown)


Julian:
Thanks! Much appreciated.

This does look likely that it fixes the bug, but we do need a test, both for the new bug (that we weren't ensuring the right validator was used), and also to preserve the test that as you say may need to move now from `parse_args` to elsewhere, that we use the latest draft when unspecified.

willson-chen:
Okay, next I will add some test cases.

Julian:
Thanks, merged this one after fixing a few minor issues.

Issue #748, opened 2020-10-07
ValidationError __str__ is far too verbose on large objects


naught101:
I have some objects that are not particularly complex, but have a large number of elements (a few nested arrays with up to 150 elements at some levels).

If I hit a ValidationError (e.g. during testing), then the error is massive - more lines than my terminal retains. I've never seen such a large error in python, and it makes testing difficult, because I can't see the backtrace.

A few possibilities for solutions:
- Just remove all of the extra information in the [__str__ return](https://github.com/Julian/jsonschema/blob/e48d56cbd2b6c25d4a9483f53f79ec537b9a4a34/jsonschema/exceptions.py#L71). Personally I'd think this was fine - the user can always get the values from `error.schema` and `error.instance`, or
- Truncate the objects - ideally by sub-selecting the elements that are causing the error.
    - Also hard-truncate the string representations of the objects (maybe 100 chars?), and add an ellipsis at the end. This is often still useful information, because if a JSON object has any kind of ID, it will usually be at the top.

Happy to submit a PR for either of these, after your feedback on the ideas.

Julian:
Hi there. This is essentially a dupe of #243, at least https://github.com/Julian/jsonschema/issues/243#issuecomment-125668925

Certainly happy to review a PR that uses `reprlib.repr`.

Julian:
(Error messages already should subselect the relevant portion of an object)

naught101:
Yep, fair enough. Let's close this, and I'll have a proper read over there and see if I can figure out a solution.

naught101:
Yeah, they do sub-select, I was just looking at a top-level error.

ablakey:
For posterity, if anyone wants a short workaround, just catch the error, remove instance, and re-raise:

```python
    try:
        validate(instance=jsonfile, schema=schema, resolver=resolver)
    except jsonschema.exceptions.ValidationError as e:
        e.instance = "<Omitted>"
        raise
```

Thanks Julian for writing and maintaining this library!

Issue #243, opened 2015-07-28
Make json look prettier in errors


arianvp:
I'm not sure if this is a real issue or not, but something that bothered me is that validation errors use `"%r"` as a format string which causes your json to look pretty ugly in the error message with all strings looking like   `{u'key':u'value'}` instead of  `{'key':'value'}`.

In my json API when I return an error it might be confusing for the developer using my API that  there's suddenly  and `u` in a string because that's not valid JSON.

Of course I could build a custom error message from all the errors I get, but that would be a bit tedious. Would it be something to consider to print the schemas and validated data as json strings instead of python dicts in error messages?

Perhaps it's something worth discussing.

Julian:
I'll follow up with a longer reply when I get a sec (sorry) -- the general answer I've had for this question has been "error messages are for developers, not end users" -- showing reprs is important for developers because reprs are what's generally informative to know what's going on. 

That being said, I think it's come up to parametrize the actual thing used here because of other reasons (truncating long instances via `reprlib.repr`) and so possibly adding that feature would make you be able to do this too if you felt like it.

Also FWIW it'd be less than tedious I think depending on exactly what you're after, but yes a slight bit of copy pasting from `ValidationError.__str__` if you wanted the exact same format.

I can follow up with some more, definitely worth discussing.

arianvp:
> Also FWIW it'd be less than tedious I think depending on exactly what you're after, but yes a slight bit of copy pasting from ValidationError.__str__ if you wanted the exact same format.

The problem with this is that the validators in `jsonschema._validators`  also user  `__repr__` in their error messages if I recall correctly. So sadly enough a bit less trivial.

Currently now I hacked in some regex that removed the `u` which is of course a bit of a fragile hack.. 

I'm curious about your idea to parameterize stuff and would love to hear more!

nostalgiaz:
+1

martino:
+1

ramanshah:
+1! I'm brand new as a user of `jsonschema`; I used it yesterday to set up a validation for some data I'd refactored out of the HTML in a legacy static website I'm maintaining. The package validates correct data beautifully, so many thanks for that! But the errors that come out when the data do not conform to the schema will be intimidating to the often unsophisticated users who will be modifying the data in the future. I appreciate that the goal of this project may not to be to serve as a lightweight data validation tool for end-users, but it's very close to achieving that goal as is.

mattig:
[...]

PR #749, opened 2020-10-09
Truncated pretty printing

naught101:
This fixes my problem at https://github.com/Julian/jsonschema/issues/748, as well of some of the problems at https://github.com/Julian/jsonschema/issues/243

This does not use reprlib, because reprlib converts nested objects into a single string, which makes pretty printing impossible.

Instead, it implements a nested list truncator, with a nice `...` formatting.

I would have liked to also truncate dictionaries after ~10 elements, but this is not really viable, as pformat pre python 3.8 unavoidably sorts dictionary keys, and the format looks horrible if you truncate it with `...: ...`. I left a comment in the code to this effect.

I suspect that list truncation is far more important than dictionary truncation anyway, and this definitely fixes my problem (thousands of lines trimmed down around a screenful.

Happy to take suggestions.

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -3,10 +3,10 @@ Validation errors, and some surrounding helpers.
 """
 from collections import defaultdict, deque
 import itertools
-import pprint
 import textwrap
 
 import attr
+import prettyprinter
 
 from jsonschema import _utils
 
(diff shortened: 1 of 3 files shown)


Julian:
Thanks for giving this a whack, it's again very much appreciated.

I'll give it a closer look, but just initially, I wouldn't want to maintain this much code for this. Essentially the extent I'd want as a default is literally passing a single (existing or other wise <3 line) function that pretty-formats instances. Even if that was somewhere in an external lib. Anyone perhaps such as yourself with a more specific desire to make stuff look a particular way can just stub that in then.

Probably it will require a bit of thinking to see how to propagate that down to the right place, but yeah as-is this is more than I'd want.

Will though take a closer look and try to give more specific suggestions but yeah the backlog seems ever growing...

naught101:
OK. It probably looks a bit bigger than it actually is, it's heavily documented. You could delete the try-except and just suffer with pformat's dictionary sorting (necessary pre 3.8).

I don't think you'll get anything that does this in 3 lines, because pformat doesn't do truncation, and reprlib doesn't do indentation, so something custom is required. I would think that a third-party dependency would be more effort than it's worth, but if you really prefer it that way, I'd be happy to move this to a separate library (or see if one exists).

Julian:
Yep, I have to believe a better version of `pformat` has been written at this point. Worth looking to see. I vaguely recall even coming across one or two in the past year or two.

naught101:
No, I'm wrong. [`prettyprinter`](https://github.com/tommikaikkonen/prettyprinter) can do it. Just pushed a new version that uses that. Much simpler code, but with the added dependency.

naught101:
Not sure if this is the best though, it sometimes results in things like this:

```python
            On instance:
                {
                    'level1': {
                        'level2': {
                            'level3': {...},
                            'level3.1': str(...),
                            'level3.2': str(...),
                            'level3.3': str(...),
                            'level3.4': str(...)
                            # ...and 1 more elements
                        },
                        'longlist': [
                            int(...),
                            int(...),
                            int(...),
                            int(...),
                            int(...),
                            # ...and 15 more elements
                        ]
                    }
                }
```

Not sure why the int/str printing. Will ask on the prettyprinter page.

naught101:
https://github.com/tommikaikkonen/prettyprinter/issues/79

PR #752, opened 2020-10-16
Add extend_with_default method

DaniilAnichin:
Intend is to make validator set default properties on mappings; It seems like an extremely common request

### DIFF
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -1,7 +1,7 @@
 """
 Creation and extension of validators, with implementations for existing drafts.
 """
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
 from functools import lru_cache
 from urllib.parse import unquote, urldefrag, urljoin, urlsplit
 from urllib.request import urlopen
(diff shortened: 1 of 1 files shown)


DaniilAnichin:
I'd be happy to also write tests covering this new method, however right now I'm having some troubles setting up test pipeline locally, and also want to understand your general reaction on such method proposal
Pls contact me, if there is something I can do to enforce such helper appearing in your project

Julian:
Hi! Thanks, much appreciated!

This is indeed a common request, so common it appears in [jsonschema's FAQ](https://python-jsonschema.readthedocs.io/en/stable/faq/#why-doesn-t-my-schema-s-default-property-set-the-default-on-my-instance)

Just making sure you've seen it there hopefully?

In general as that answer alludes to, I haven't wanted to add this to `jsonschema` itself essentially for 2 reasons -- this library stays close to the specification, and this behavior isn't part of the spec's behavior, and relatedly or perhaps even because of the above, it's significantly different behavior from any other validator "built-into" jsonschema.

There also are a number of choices someone may make here in terms of how they want this to work (e.g. should the value included in default be required to validate under the schema? Should the value there be processed before/after required? etc.)

So in short I'd absolutely love to see this and other related useful things become part of some higher level libraries on *top* of `jsonschema`.

Hope that makes sense? Lemme know what you think.

Julian:
Closing but happy to discuss further if you have any comments on the above!

PR #765, opened 2020-12-20
Add pre-commit hook config

AleksaC:
Hi, I think it would be useful to add support for jsonschema as [pre-commit](https://github.com/pre-commit/pre-commit) hook. I think this would be more convenient than using local hooks or creating a mirror repo. 

With the way I set it up, the hook can be used with a config like this:

```yaml
repos:
  - repo: https://github.com/AleksaC/jsonschema
    rev: 639bc64d82
    hooks:
      - id: jsonschema
        files: |
          (?x)(
            ^valid.json$ |
            ^invalid.json$ |
            ^schema.json$
          )
        args: ["-i", "valid.json", "-i", "invalid.json", "schema.json"]

      - id: jsonschema
        files: |
          (?x)(
            ^valid2.json$ |
            ^invalid2.json$ |
            ^schema2.json$
          )
        args: ["-i", "valid2.json", "-i", "invalid2.json", "schema2.json"]
```

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: jsonschema
+  name: jsonschema
+  description: json schema validation
+  language: python
+  pass_filenames: false
+  entry: jsonschema


Julian:
Hi there. Sounds great, thanks! Will merge and see what feedback we get.

asottile:
I think the interface @janosh suggests is right, but the default configuration should still require the user to set up `files` *and* `args` to filter to the targetted files properly and to select the right schema

something like:

```yaml
-   id: jsonschema-v2
    name: jsonschema
    entry: jsonschema-hook
    language: python
    # require the user to override args to set `[--schema, somefile.json]`
    # require the user to override `files` to filter to the target files
    files: ^$ 
```

or ~something like that

in this `jsonschema-hook` would be a script which has the following "interface"

```
jsonschema-hook --schema SCHEMA [other options] [file [file [...]]]
```

AleksaC:
Hi, as the author of this pull request I'd like to chime in. First of all sorry for not giving more thought to this, it worked well for my use case, but I realize that it isn't ideal. I knew that there wasn't a way to use the current cli with pre-commit in idiomatic way, I just assumed that changing the cli for pre-commit wasn't something you'd want to do.

There is still one thing that isn't addressed - if you change the schema you'd probably want to run the hook on all the files it is supposed to check to make sure there aren't any discrepancies. With the wrapper proposed above it would be solved by defining the hook twice, the second definition being something like this:
```yaml
hooks:
  - id: jsonschema-v2
    files: ^schema.json$
    args: [--schema schema.json, ...files]
    pass_filenames: false
```

Anyway if you want to add a wrapper for pre-commit, I'd be willing to do that as it's sort of my mess to clean up :smile:

Issue #771, opened 2021-01-18
Integrating continuous fuzzing by way of OSS-Fuzz


DavidKorczynski:
Hi,

I was thinking that it would be nice to set up continuous fuzzing of jsonschema, by way of OSS-Fuzz. In this PR: https://github.com/google/oss-fuzz/pull/4996 I have done exactly that, namely created the necessary logic from an OSS-Fuzz perspective to integrate jsonschema. This includes developing initial fuzzers as well as integrating into OSS-Fuzz.

Essentially, OSS-Fuzz is a free service run by Google that performs continuous fuzzing of important open source projects. The only expectation of integrating into OSS-Fuzz is that bugs will be fixed. This is not a "hard" requirement in that no one enforces this and the main point is if bugs are not fixed then it is a waste of resources to run the fuzzers, which we would like to avoid.

If you would like to integrate, could I please have an email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports, coverage reports and more stats. Notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

Julian:
Hi there. Thanks for the offer / sending the PR.

So -- all of jsonschema's code is pure Python at the minute, so I'd be curious whether OSS-Fuzz could say anything interesting without at least some information about how to generate JSON Schema specification -alike objects. But happy to see what pops out too.

The email address that's in the [SECURITY.md](https://github.com/Julian/jsonschema/blob/main/.github/SECURITY.md) file is a decent place to send these.

CC @Zac-HD in case you're interested or have opinions :)

And thanks for raising!

Julian:
I didn't know this, though surely @Zac-HD will, but looks like OSS-Fuzz supports generating data [via Hypothesis](https://google.github.io/oss-fuzz/getting-started/new-project-guide/python-lang/#hypothesis)

Something like that would be a big improvement over random dictionary poking I think. Zac may already be doing that himself as part of hypothesis-jsonschema? But if not, @DavidKorczynski I think that'd be the right kind of integration with OSSFuzz.

DavidKorczynski:
Am happy to set up a property-based approach with Hypothesis if you are happy to integrate with OSS-Fuzz! If I go ahead with the integration I can submit the fuzzers upstream in this repository instead of keeping them on OSS-Fuzz, then we can also get the property-based testing going - does that sound good?

Julian:
Sounds good to me!

[...]

PR #772, opened 2021-01-21
Added fuzzer to be run with OSS-Fuzz

DavidKorczynski:
Adds a fuzzer as a step to integrating into OSS-Fuzz.

Cross-referencing 
- https://github.com/Julian/jsonschema/issues/771
- https://github.com/google/oss-fuzz/pull/4996

@Julian I now added more structure to it by way of Hypothesis and am happy to put more effort in. I think it would be nice to get it up and running on OSS-Fuzz though, so we can start to see when results happen.

@Zac-HD Thanks for your assistance in the other issue. Will keep digging into Hypothesis!


Julian:
> @Julian I now added more structure to it by way of Hypothesis and am happy to put more effort in. I think it would be nice to get it up and running on OSS-Fuzz though, so we can start to see when results happen.

Cool! So the idea here which I think I'm OK with is to get something here that we can run and iterate on without needing to touch the upstream OSS-Fuzz repo, right? I think to @Zac-HD's point what's here so far likely wouldn't "do" much utility wise but with some tweaks hopefully would. So we could just push forward and work on those tweaks.

I do think much as it might seem annoying, that we *should* address the coverage failure -- meaning, it would be good to have a "test" that executes the fuzzer for some extremely tiny amount, just so that we know any API changes don't suddenly silently break integration with OSS-Fuzz.

Does that make sense to you / seem like it'd be non-totally-terrible to do? The hardest part may be that module here which may need stubbing that's OSS-Fuzz specific? Let me know thoughts otherwise I can think myself about what that might look like, but yeah do think it'd be nice if somewhere locally here in CI we knew this script continued to function.

DavidKorczynski:
> > @Julian  I do think much as it might seem annoying, that we _should_ address the coverage failure -- meaning, it would be good to have a "test" that executes the fuzzer for some extremely tiny amount, just so that we know any API changes don't suddenly silently break integration with OSS-Fuzz.
> 
> Does that make sense to you / seem like it'd be non-totally-terrible to do? The hardest part may be that module here which may need stubbing that's OSS-Fuzz specific? Let me know thoughts otherwise I can think myself about what that might look like, but yeah do think it'd be nice if somewhere locally here in CI we knew this script continued to function.

I understand your concern although having integrated loads of projects into OSS-Fuzz it's a first to actually write tests that tests the fuzzers. How about doing the following approach - I added a `main.yml` file that integrates jsonschema into CI-fuzz, i.e. the fuzzer will be run for 600 seconds for every pull request. This might be a good approach to this?

DavidKorczynski:
> @Julian Cool! So the idea here which I think I'm OK with is to get something here that we can run and iterate on without needing to touch the upstream OSS-Fuzz repo, right? 
> 

Yep, no iterations needed from you on OSS-Fuzz. Also, we are collaborating with the OSS-Fuzz folks so you can ping me whenever if you think there is something going that needs fixing in OSS-Fuzz and I will handle it for you. But by default you will not need to touch OSS-Fuzz.

Julian:
> I added a main.yml file that integrates jsonschema into CI-fuzz, i.e. the fuzzer will be run for 600 seconds for every pull request. This might be a good approach to this? Let me know if not and I will then add a test.

I think even if this were 2 seconds rather than 600 seconds that my concern would be addressed. It's just "make sure the thing runs" any time someone makes a change to it.

DavidKorczynski:
The reason I left 600 seconds is because OSS-Fuzz sets it as a minimum (https://google.github.io/oss-fuzz/getting-started/continuous-integration/#integrating-into-your-repository)

Julian:
> The reason I set 600 seconds is because OSS-Fuzz sets it as a minimum (https://google.github.io/oss-fuzz/getting-started/continuous-integration/#integrating-into-your-repository)

It's not really a minimum, that's the docs saying any less than that and it's not very useful for fuzzing not that it will fail the build or anything -- let's set it to 30 seconds I think.

Looks like it's failing now though for some config reason?

DavidKorczynski:
> Looks like it's failing now though for some config reason?

I recon this is because the project is not ready integrated in OSS-Fuzz  (we are waiting for this to merge)

Julian:
Ah fair enough. Can you maybe make that job not fail the build then (that might be reasonable even going forward perhaps)?

After that sounding like we're close to merge.

DavidKorczynski:
> Ah fair enough. Can you maybe make that job not fail the build then (that might be reasonable even going forward perhaps)?
> 
> After that sounding like we're close to merge.

Could you clarify on this one  - am not sure I understand (this is the first time I integrate a CI job). 

Do you mean make it not fail by merging things in OSS-Fuzz, or do you mean not fail by changing something in the `main.yml` file? I put in a `continue-on-error: true` for the fuzzing job, is this perhaps what you were looking for?

DavidKorczynski:
I think I got a solution you were looking for now @Julian - now the CIFuzz job will exit with success in case the `build` step fails.

Julian:
Hooray. Will have a close look in the morning when I've got a clean mind but think that's perfect!

jsonschema/tests/fuzz_validate.py
```diff
@@ -0,0 +1,39 @@
...
+
+from hypothesis import given
+from hypothesis import strategies as st
+import atheris
+
+import jsonschema
+
+PRIM = st.one_of(
+    st.booleans(),
+    st.integers(min_value=-(2 ** 63), max_value=2 ** 63 - 1),
```
Zac-HD, line 11:
```suggestion
    st.integers(),
```

The 64-bit limit is an implementation issue in `ujson`, rather than a general property of JSON objects.

jsonschema/tests/fuzz_validate.py
```diff
@@ -0,0 +1,39 @@
...
+    st.floats(allow_nan=False, allow_infinity=False),
+    st.text(),
+)
+DICT = st.recursive(
+    base=st.dictionaries(st.text(), PRIM),
```
Zac-HD, line 16:
```suggestion
    base=st.booleans() | st.dictionaries(st.text(), PRIM),
```

Booleans are also valid schemas (!), and that's been a fertile source of bugs for my schema-processing logic especially variously nested.

Zac-HD (commented):
One nice way to ensure everything is covered is to run the fuzz harness *as a test function!*  Specifically:

- Rename the file as e.g. `test_foo_fuzz.py`, so that `pytest` collects it;
- Adjust the glob pattern in the OSS-Fuzz config to detect `*fuzz.py` instead of `fuzz_*.py`
- Move `import atheris` into the `if __name__ == "__main__":` block, so that it's a fuzz-time but *not* a test-time dependency

And then it'll be covered as expected (modulo perhaps a coverage pragma on the `__name__` branch...)

jsonschema/tests/fuzz_validate.py
DavidKorczynski: Cheers!

jsonschema/tests/fuzz_validate.py
DavidKorczynski: Cheers!

Julian:
@DavidKorczynski apologies for the delay here, have been dealing with some other stuff first, but hopefully will get to merging this by the end of the weekend.

DavidKorczynski:
No problem @Julian  - please take your time and handle this when it makes sense to you. There is no stress from my side.

Julian:
@DavidKorczynski apologies here that this slipped, but I'm likely about to merge as soon as CI passes (which I put some minor changes in for).

I did re-rename this back to `fuzz_*` rather than being prefixed with `test_` even though I agree that's a good thing in the long term, but for now hypothesis isn't a testing dependency so it was failing the normal CI run when that module was imported, and rather than bother to futz with installing hypothesis to be able to import that file, I just moved it back for now.

You'll obviously see when/if I merge in a bit after I see the build go green.

DavidKorczynski:
Wohoooo - we got it done :)! https://github.com/google/oss-fuzz/pull/4996

Thanks @Zac-HD for the input and thanks @Julian for getting it merged in.

Julian:
Thank you!

PR #774, opened 2021-01-26
Sort extra properties for additionalProperties errors

robherring:
extras is a set which is unordered. Let's sort it so users get
consistent error messages.

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -63,7 +63,7 @@ def additionalProperties(validator, aP, instance, schema):
             yield ValidationError(error)
         else:
             error = "Additional properties are not allowed (%s %s unexpected)"
-            yield ValidationError(error % extras_msg(extras))
+            yield ValidationError(error % extras_msg(sorted(extras)))
 
 
 def items(validator, items, instance, schema):


Julian:
Hi, can you elaborate on what problems the error messages not being in consistent sort order causes?

Julian:
Closing since I'm not sure I follow what the idea here would be, but happy to discuss further.

robherring:
Users (of dtschema) have reported to me that the order changes from run to run and diffing error logs shows differences. So it's problematic to have that noise.

Also note that 'extras' is sorted in the case with 'patternProperties' just above, but not here.

Julian:
The exact contents of error messages [aren't part of jsonschema's public API](https://python-jsonschema.readthedocs.io/en/stable/faq/#how-do-jsonschema-version-numbers-work) -- so `patternProperties`'s extras may become *non-sorted* at any point in time if it somehow becomes required to do so.

It sounds like perhaps these users should instead be logging the [fields off of ValidationErrors that they want to track](https://python-jsonschema.readthedocs.io/en/stable/errors/#jsonschema.exceptions.ValidationError) in a structured way, rather than relying on the precise error message, which as mentioned isn't guaranteed to look any certain way or another.

(Though #119 may be relevant here)

robherring:
> The exact contents of error messages [aren't part of jsonschema's public API](https://python-jsonschema.readthedocs.io/en/stable/faq/#how-do-jsonschema-version-numbers-work) -- so `patternProperties`'s extras may become _unsorted_ at any point in time if it somehow becomes required to do so.

I'm only asking for it to be stable within a given release, not across releases.

> It sounds like perhaps these users should instead be logging the [fields off of ValidationErrors that they want to track](https://python-jsonschema.readthedocs.io/en/stable/errors/#jsonschema.exceptions.ValidationError) in a structured way, rather than relying on the precise error message, which as mentioned isn't guaranteed to look any certain way or another.
> 
> (Though #119 may be relevant here)

Exactly, the data in question isn't in a structured field to extract it.

Julian:
> Exactly, the data in question isn't in a structured field to extract it.

Got it -- then I think the solution is to implement #119 -- there's I believe a specced solution there in the comments if you were interested in giving it a shot (or if further discussion is needed obviously happy to do that).

Even within one release, there's yeah just no guarantees provided here.

PR #780, opened 2021-02-06
Fix test to windows path matching

Milly:
Tests are failed in windows. Because path separator is hard corded in test.


Julian:
Hi there! Thanks for sending this (and apologies for not noticing it previously)

I think CI on the main branch is passing on Windows even before this fix, unless I'm mistaken. What you changed looks reasonable, but yeah did you happen to look at why that is?

Milly:
I'm using *jsonschema* in windows.
When I used it or tried to fix it, I couldn't run the test in my local environment.

PR #799, opened 2021-04-11
allOf required properties

Spintronic:
Added support to specify required properties against inherited properties via allOf. See below.

components:
  schemas:
    Thing:
      type: object
      properties:
        name:
          type: string
        color:
          type: string
	age:
          type: int  
paths:
  /things:
      requestBody:
        description: Thing object
        content:
          application/json:
            schema:
		allOf:
		- $ref: '#/components/schemas/Thing'
		- type: object
	          required: 
	            - name
	            - color
	          properties:
		      species:
		         type: string  
        required: true

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -331,14 +331,12 @@ def allOf(validator, allOf, instance, schema):
         if u"properties" in subschema:
             all_properties.update(subschema["properties"])
         elif u"$ref" in subschema:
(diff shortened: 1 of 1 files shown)


Julian:
Hi there. Can you elaborate on whether this PR is attempting to fix a bug (i.e. something specified by the JSON Schema specification but incorrectly implemented by this library)?

PR #805, opened 2021-04-21
Add support for reading input files from a directory

richin13:
**Description**

This PR adds support for running the CLI against a directory containing JSON files, instead of listing each file individually.


Julian:
Hi there. Thanks for the PR. This seems like reasonable functionality to add. Can you look at adding some tests? (Which should also fix the CI failure)

Julian:
@richin13 more than happy to add this if you add some tests. Closing just to clear the queue, but feel free to follow up if you're still interested!

Issue #612, opened 2019-10-09
3.1.0 Release fails to validate strings via regex


amisevsk:
A CI build of ours pulled the new release (3.1.0) and all validation is failing with messages like
```
u'go' does not match u'^[^\\s]+$'
```
Building with `<3.1.0` does not have this issue.

Reproducer:

test.json:
```json
{
  "test": "testString"
}
```

schema.json;
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "Reproducer",
  "type": "object",
  "properties": {
    "test": {
      "type": "string",
      "pattern": "^[^\\s]+$"
    }
  }
}
```

dockerfile (or just install 3.1.0 locally)
```dockerfile
FROM alpine:3.10
RUN apk add --no-cache py-pip jq && pip install jsonschema

COPY . /build
RUN jsonschema --version
RUN jsonschema /build/schema.json -i /build/test.json
```

Julian:
Having a look...

amisevsk:
Thanks @Julian, from a somewhat hasty check, looks to be related to character classes (`"^[^ \t]+$"` does not have the issue)

dslagle:
This appears to be related to the switch to js_schema. Specifically in relation to character classes where something like `[\w0-9]` gets translated to `[[a-zA-Z]0-9]` inside js_schema. At least that's what happened in my case.

Julian:
OK, given that there's 2 separate issues here (#612 and #611) and that this is fairly fundamental functionality, I've temporarily pushed out v3.1.1 reverting the switchover.

We'll add some tests for the two cases here (and anything else we missed), and then there likely will be a v3.1.2 that goes back again to js-regex again.

Zac-HD:
> This appears to be related to the switch to `js_regex`. Specifically in relation to character classes where something like `[\w0-9]` gets translated to `[[a-zA-Z]0-9]` inside `js_regex`.

Yep, that's a bug in `js-regex`.  Sorry for the trouble this has caused everyone 😕

amisevsk:
Thanks all, FWIW I re-read the jsonschema draft docs and it *does* suggest not using anything outside of pretty limited range of the regex spec, so my issues are partly my own doing ;)

tschmidtb51:
@Julian: What is the current status on that issue? Does jsonschema still use `js_regex`?

Julian:
[...]

PR #817, opened 2021-06-14
Add support for draft 2020-12

nezhar:


Julian:
Hello there!

Thanks very much for this, at first glance it looks like it's great, definitely in the right ballpark.

The first thing I think we'll need to address are the changes to the `json/tests` folder -- those come from the [upstream JSON-Schema-Test-Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite) (which I also help co-maintain with others from cross-language communities).

If there are gaps you noticed there, we need to basically take them upstream (and then we'll be able to pull those changes into that folder, which is a `git subtree`).

Could you perhaps consider pulling those out as PRs to that repository? Ideally in self-contained chunks for each piece.

I'll have to look more carefully at the changes to give more specific comments, but from my first skim it seemed exactly like what I'd expect.

Let's see whether CI passes as well, which I just manually enabled.

And thanks again!

nezhar:
Thanks :slightly_smiling_face:, good to know it's going the right way.

Sure, I can remove that commit and create a dedicated PR if this fits better.
I usually use a `git submodule` to include another repository, this can be also done with https://github.com/json-schema-org/JSON-Schema-Test-Suite. What do you think?

The implementation is still in progress, the curent state is of the `draft2020-12` tests is `1334 total, 168 failed, 111 ignored, 1055 passed`. The `ignored` tests should lower once the missing format validations are implemented.

I added [isodate](https://pypi.org/project/isodate/) new dependency for the `duration` validation, but it seems that the lib does not cover some cases: https://github.com/Julian/jsonschema/pull/817/files#diff-0953f1b9ffe16c7d4aa18ae8bf21287c552dcd9b241f66fb51e099724af8b722R616. Any sugestion here? There seems no beter implementation of the standard. My idea was to cover the missing cases innside the validation function as the library had no release since 2017.

Julian:
> Sure, I can remove that commit and create a dedicated PR if this fits better.
I usually use a git submodule to include another repository, this can be also done with https://github.com/json-schema-org/JSON-Schema-Test-Suite. What do you think?

That's already the case here functionally, that directory is a `git-subtree` (https://www.atlassian.com/git/tutorials/git-subtree), which is essentially a better version of git submodules.

> The implementation is still in progress, the curent state is of the draft2020-12 tests is 1334 total, 168 failed, 111 ignored, 1055 passed. The ignored tests should lower once the missing format validations are implemented.

Cool! That sounds promising, and yeah matches my experience when implementing earlier drafts.

> Any sugestion here? There seems no beter implementation of the standard. My idea was to cover the missing cases innside the validation function as the library had no release since 2017.

I'll have to do some research myself. If we can't find something, the alternative is to possibly just not support the format. I don't really want to maintain a ton of format-specific code internally in the library. But let me see what I can find myself.

tschmidtb51:
> I added [isodate](https://pypi.org/project/isodate/) new dependency for the `duration` validation

Not quite sure, but could [isoduration](https://pypi.org/project/isoduration/) do the job?

nezhar:
> > I added [isodate](https://pypi.org/project/isodate/) new dependency for the `duration` validation
> 
> Not quite sure, but could [isoduration](https://pypi.org/project/isoduration/) do the job?

Already tried `isoduration`, but the implementation seems to be worst than the one in `isodate`, so  I decied to go back as there are only some cases that are considered invalid in the tests, but are parsed by `isodate`. They are collected in `duration_format_validation`.

nezhar:
Inside `ecmascript_regex_validation` there are several skips related to `ecmascript-regex` tests that where introduced with `draft2020-12`. I saw there was an issue once related to this topic: https://github.com/Julian/jsonschema/issues/612 and [js_regex](https://github.com/Zac-HD/js-regex) tried to solve such issues due to the imcompatiblity of implementation of regular expression across languages.

The package was latter removed due to https://github.com/Zac-HD/js-regex/issues/4 and now the repository is archived. Are there any plans to cover this?

nezhar:
Some tests are skiped in `format_validation_annotation` as the format validation is a bit confusing. Is there any dirference beetween https://github.com/Julian/jsonschema/blob/main/json/tests/draft2020-12/format.json#L775 and https://github.com/Julian/jsonschema/blob/main/json/tests/draft2020-12/optional/format/duration.json#L12? In terms of testing they both define the same schema and use the same data instance, but expect diferent results.

tschmidtb51:
> Some tests are skiped in `format_validation_annotation` as the format validation is a bit confusing. Is there any dirference beetween https://github.com/Julian/jsonschema/blob/main/json/tests/draft2020-12/format.json#L775 and https://github.com/Julian/jsonschema/blob/main/json/tests/draft2020-12/optional/format/duration.json#L12? In terms of testing they both define the same schema and use the same data instance, but expect diferent results.

If I remember correctly that is due to the new rules for Format Vocabulary (see the description in [2019-09](https://json-schema.org/draft/2019-09/release-notes.html).

nezhar:
> > > I added [isodate](https://pypi.org/project/isodate/) new dependency for the `duration` validation
> > 
> > 
> > Not quite sure, but could [isoduration](https://pypi.org/project/isoduration/) do the job?
> 
> Already tried `isoduration`, but the implementation seems to be worst than the one in `isodate`, so I decied to go back as there are only some cases that are considered invalid in the tests, but are parsed by `isodate`. They are collected in `duration_format_validation`.

Tried [isoduration](https://pypi.org/project/isoduration/) and it seems to work well, maybe I defined the wrong exepction in the `raises` param at the first run. bf4b288 is now switching form `isodate` to `isoduation`.

Julian:
> The package was latter removed due to Zac-HD/js-regex#4 and now the repository is archived. Are there any plans to cover this?

It's not a priority for me personally at the minute -- if a library comes along in the Python ecosystem that provides JS regexes, it'll be considered, but yeah there are way more important things to address (including this PR!)

jsonschema/validators.py
```diff
@@ -438,7 +438,8 @@ def extend(validator, validators=(), version=None, type_checker=None):
         u"anyOf": _validators.anyOf,
         u"const": _validators.const,
         u"contains": _validators.contains,
-        u"dependencies": _validators.dependencies,
```
robherring: AIUI, 'dependencies' is still a valid keyword even if deprecated. It's still in the meta-schema.

robherring (commented):
I gave your branch a spin (using https://github.com/devicetree-org/dt-schema) and found a few issues.

Ultimately, 202012 is too incompatible for me to test this further easily. I'm primarily interested in 'unevaluatedProperties'. The big issue is 'items' lists have been replaced with prefixItems and that's used *everywhere*. So I need 201909 support. Is that something you plan?

jsonschema/validators.py
```diff
@@ -161,12 +161,7 @@ def iter_errors(self, instance, _schema=None):
             if scope:
                 self.resolver.push_scope(scope)
             try:
-                ref = _schema.get(u"$ref")
-                if ref is not None:
-                    validators = [(u"$ref", ref)]
-                else:
-                    validators = _schema.items()
-
+                validators = _schema.items()
```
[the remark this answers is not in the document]
nezhar: Thanks for the hint, will look into this, probably `Draft7` and earlier tests are already failing.

jsonschema/validators.py
nezhar: I'm not sure, but according to the [migration guide](https://json-schema.org/draft/2019-09/release-notes.html) to `2019-19` this has been split to `dependentSchemas` and `dependentRequired`. The test cases have also been removed from [JSON-Schema-Test-Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite) for the newer drafts.

The meta-schema reserves the keyword for backwards compatibility, so I assume that the implementation decides if this should be supported or not. @Julian any suggestion?

jsonschema/_validators.py
```diff
@@ -78,7 +79,11 @@ def items(validator, items, instance, schema):
             if len(instance) > len(schema['prefixItems']):
                 yield ValidationError("%r has more items than defined in prefixItems" % instance)
             else:
-                for error in validator.descend(instance, {'prefixItems': schema['prefixItems']}, path='items__prefixItems'):
+                for error in validator.descend(
+                        instance,
+                        {'prefixItems': schema['prefixItems']},
+                        path='items__prefixItems',
+                ):
```
robherring, line 86: Belongs in another commit.

jsonschema/_validators.py
```diff
@@ -496,9 +501,6 @@ def if_(validator, if_schema, instance, schema):
             yield error
 
 
-
-
-
 def unevaluatedItems(validator, unevaluatedItems, instance, schema):
```
robherring, line 504: Belongs in another commit. Though, it's up to @Julian on how clean the commits are.

jsonschema/_validators.py
```diff
@@ -514,6 +516,22 @@ def unevaluatedItems(validator, unevaluatedItems, instance, schema):
...
+    evaluated_property_keys = find_evaluated_property_keys_by_schema(validator, instance, schema)
+    for property, subschema in instance.items():
+        if property not in evaluated_property_keys:
+            for error in validator.descend(
+                    instance[property],
+                    unevaluatedProperties,
+                    path=property,
+                    schema_path=property,
+            ):
+                yield error
```
robherring, line 528: With 'additionalProperties: false', you get a message saying what unexpected properties were found. With 'unevaluatedProperties: false', you get 'false schema does not allow: <value of unexpected property>'. It would be nice to follow the 'additionalProperties' code here if possible.

I didn't check, but I'd guess the same thing applies to unevaluatedItems WRT additionalItems.

robherring (commented):
I managed to hack in 2019-09 support and test out 'unevaluatedProperties'. Works well for me, but a few comments.

jsonschema/validators.py
Julian: It's meant to still be supported via an optional compatibility mode, not really by default IIRC.

So yeah I'd remove them for now. We can choose to add the compatibility mode later if we want, but I also think I agree (with the spec) that it's trivial to use the new properties so it doesn't seem very high value even as an opt-in.

jsonschema/_validators.py
Julian: We're going to have to redo some of these commits anyhow to get the tests upstream into the test suite -- @nezhar maybe it's a decent time to start doing some of that? Otherwise they'll start to pile up.

Overall though I don't mind much if the commits aren't 100% clean, just keep them as clean as you can and if I can't decipher them I'll ask.

nezhar: Currenlty all tests are fetched form the external test suite via https://github.com/Julian/jsonschema/pull/817/commits/d371dc53007ee947e81e049d2624ff46a8674551 - I can move this commit up in the history or create a dedicated PR so it does not polute the list of changes.

Would it make sense to squash the commits the implementation is done? I was thinking to keep them like this for the review, but certain keywords need to be extended when implementing others.

Julian:
@nezhar I pulled in the subtree changes from the test suite.

If you rebase or merge your branch, it should now remove those changes from this PR, so any outstanding changes to the `json/` folder (i.e. changes you made to tests) we should take to the test suite repo.

jsonschema/validators.py
nezhar: In this case, should we move the `dependencies` validation function to `_legacy_validators.py`

Julian: Yes. Basically any validator that doesn't apply to the latest draft I move to the legacy file.

bolsote:
@nezhar Hello, I'm the author of `isoduration`. First of all, thank you for picking the library.

Since there's not been much activity recently, I'd like to mention I'm currently working on these two issues:

- https://github.com/bolsote/isoduration/issues/9: In this one we are making sure all known `isodate` issues do not affect `isoduration`. The test suite in `isoduration` should be a strict superset of that of `isodate`. If you find it is not, please report it, as that would be a bug.
- https://github.com/bolsote/isoduration/issues/5: As of right now we are not really ISO/DIS 8601-1 compliant, as you can have decimal points at any place, while they are only accepted at the very last element of each segment. This attempts to fix it.

The first of those issues already has a PR attached, adding a few tests. No functional changes. The second one will have a PR soon (most likely tomorrow). I've reserved a healthy chunk of time in the next few weeks to solve all other issues, particularly the one related to supporting repeating intervals.

I also understand there were some teething troubles while integrating the library, maybe due to unexpected interfaces? If so, please do let me know, and we can see if the situation can be improved.

jsonschema/validators.py
```diff
@@ -161,15 +161,22 @@ def iter_errors(self, instance, _schema=None):
...
-                validators = []
-
-                # We make sure $ref is evaluated first if available in schema
-                ref = _schema.get(u"$ref")
-                if ref is not None:
-                    validators = [(u"$ref", ref)]
-
-                # Add all remaining validators
-                validators += [(x, _schema[x]) for x in _schema if x not in ["$ref"]]
+                if isinstance(self, Draft202012Validator):
```
robherring, line 164: This will have to be updated for every new validator such as 2019-09. That seems error prone. I think the validator needs some sort of flag for this or perhaps a version that can be compared using =, <, or >.

jsonschema/validators.py
```diff
@@ -161,15 +161,22 @@ def iter_errors(self, instance, _schema=None):
...
+                    validators = []
+                    # We make sure $ref is evaluated first if available in schema
+                    ref = _schema.get(u"$ref")
+                    if ref is not None:
+                        validators = [(u"$ref", ref)]
```
robherring, line 169: These 3 lines could be factored out of the if/else.

jsonschema/validators.py
nezhar: I agree. For now I made a list of older classes in `legacy_cls` in the new `schema_validators` method, so newer versions should not be affected.

jsonschema/validators.py
nezhar: Done

jsonschema/validators.py
nezhar: The function has been moved to `_legacy_validators.py`

jsonschema/validators.py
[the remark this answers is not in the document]
nezhar: Done

jsonschema/_legacy_validators.py
```diff
@@ -27,6 +27,30 @@ def dependencies_draft3(validator, dependencies, instance, schema):
                     yield ValidationError(message % (each, property))
 
 
+def dependencies_draft4_draft6_draft7(validator, dependencies, instance, schema):
+    """
+    The dependencies keyword has been deprecated since draft 2019-09 and has been split into dependentRequired
```
Julian, line 32:
```suggestion
    Support for the ``dependencies`` validator from pre-draft 2019-09.
    
    In later drafts, the validator was split into separate ``dependentRequired``
    and ``dependentSchemas`` validators.
```

jsonschema/_format.py
```diff
@@ -412,18 +444,43 @@ def is_relative_json_pointer(instance):
 
 
 try:
-    import uritemplate.exceptions
+    import uri_template
```
Julian, line 447: Can you pull this out into its own PR (which likely can merge immediately).

nezhar:
> @nezhar Hello, I'm the author of `isoduration`. First of all, thank you for picking the library.
> 
> Since there's not been much activity recently, I'd like to mention I'm currently working on these two issues:
> 
> * [Solve all isodate issues bolsote/isoduration#9](https://github.com/bolsote/isoduration/issues/9): In this one we are making sure all known `isodate` issues do not affect `isoduration`. The test suite in `isoduration` should be a strict superset of that of `isodate`. If you find it is not, please report it, as that would be a bug.
> * [Properly support decimal points bolsote/isoduration#5](https://github.com/bolsote/isoduration/issues/5): As of right now we are not really ISO/DIS 8601-1 compliant, as you can have decimal points at any place, while they are only accepted at the very last element of each segment. This attempts to fix it.
> 
> The first of those issues already has a PR attached, adding a few tests. No functional changes. The second one will have a PR soon (most likely tomorrow). I've reserved a healthy chunk of time in the next few weeks to solve all other issues, particularly the one related to supporting repeating intervals.
> 
> I also understand there were some teething troubles while integrating the library, maybe due to unexpected interfaces? If so, please do let me know, and we can see if the situation can be improved.

Hello @bolsote, thanks for creating and maintaining the library :1st_place_medal:

As mentioned in https://github.com/Julian/jsonschema/pull/817#issuecomment-863963902 it works well and covers all test cases specified in https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/master/tests/draft2020-12/optional/format/duration.json and it is already part of this PR.

jsonschema/_format.py
nezhar: Moved to https://github.com/Julian/jsonschema/pull/819

jsonschema/_utils.py
```diff
@@ -158,10 +158,46 @@ def ensure_list(thing):
...
+            return False
+
+    return True
+
+
 def equal(one, two):
     """
     Check if two things are equal, but evade booleans and ints being equal.
     """
+    if isinstance(one, list) and isinstance(two, list):
```
nezhar, line 195: @Julian should I also extract `list_equal` and `dict_equal` into a dedicated PR for #686?

nezhar:
> > Some tests are skiped in `format_validation_annotation` as the format validation is a bit confusing. Is there any dirference beetween https://github.com/Julian/jsonschema/blob/main/json/tests/draft2020-12/format.json#L775 and https://github.com/Julian/jsonschema/blob/main/json/tests/draft2020-12/optional/format/duration.json#L12? In terms of testing they both define the same schema and use the same data instance, but expect diferent results.
> 
> If I remember correctly that is due to the new rules/bahavior for Format Vocabulary (see the description in [2019-09](https://json-schema.org/draft/2019-09/release-notes.html)).

There is something maybe that I miss here. Having this schema:

```
{ "format": "duration" }
```

This test should pass (format.json)
```
{
    "description": "invalid duration string is only an annotation by default",
    "data": "PT1D",
    "valid": true
}
```

But this test should fail (optional/format/duration.json)
```
{
    "description": "an invalid duration string",
    "data": "PT1D",
    "valid": false
}
```

Same schema and same data, and this applies for all format tests. Is this suposed to be controled with an extra param or something?

jsonschema/validators.py
```diff
@@ -644,11 +639,37 @@ def resolving(self, ref):
         finally:
             self.pop_scope()
 
+    def _finditem(self, schema, key):
```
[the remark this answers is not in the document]
nezhar: Done

Julian:
> Is this suposed to be controled with an extra param or something?

It's a bit clunky, but the `format.json` file is intended to be ran with format-as-annotation (in this library that means with no configured `FormatChecker()`) and the latter with one enabled. It's indeed the same test, but ran with two different configurations, which changes the expected result.

It may be we need a small tweak to the way we load the test suite to accommodate that.

nezhar:
> > Is this suposed to be controled with an extra param or something?
> 
> It's a bit clunky, but the `format.json` file is intended to be ran with format-as-annotation (in this library that means with no configured `FormatChecker()`) and the latter with one enabled. It's indeed the same test, but ran with two different configurations, which changes the expected result.
> 
> It may be we need a small tweak to the way we load the test suite to accommodate that.

Splitting the tests seems to be working fine: https://github.com/Julian/jsonschema/pull/817/commits/1a65f81f2ea81d251eb7ede2d6f61617e238b38e#diff-0953f1b9ffe16c7d4aa18ae8bf21287c552dcd9b241f66fb51e099724af8b722R493

jsonschema/_utils.py
Julian: Yes please! Especially because this implementation will need some tweaking, it will break for arbitrary mappings and sequences, which we support (so we need to add tests for that).

But let's take that to a separate PR, appreciated!

jsonschema/tests/test_format.py
```diff
@@ -86,3 +86,100 @@ def test_repr(self):
...
+
+try:
+    import isoduration
+
+    from jsonschema._format import is_duration
+except ImportError:
+    pass
+else:
+    class TestDurationFormat(TestCase, FormatTestMixin):
+        valid_instances = [
```
Julian, line 140: All these should be moved upstream.

nezhar: I will remove them, I was not sure why the `duration` format was not covered by the tests. Turns out this is related to `pypy3` as this runs on default with Python 3.6 and the library requires Python >=3.7. I changed the gihub action to use Python 3.9 instead. We can also use `pypy-3.7` if pypy is a requirement,

Julian: Running on 3.9 as you did is fine!

nezhar: Test cases moved to https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/494

jsonschema/_utils.py
nezhar: Moved to https://github.com/Julian/jsonschema/pull/820. Test cases will be currently ignored in draft2020-12, once merged I can rebase an remove the commit.

jsonschema/_validators.py
nezhar: I made an update for this in https://github.com/Julian/jsonschema/pull/817/commits/9200165c40f463e0b7b7fa1e5fba5ef5eacdd33e

jsonschema/_validators.py
nezhar: Squashing the commits before merging is a better option here. @Julian what do you think?

jsonschema/_validators.py
```diff
@@ -70,13 +73,31 @@ def items(validator, items, instance, schema):
...
-                item, subschema, path=index, schema_path=index,
+                    instance,
+                    {'prefixItems': schema['prefixItems']},
+                    path='items__prefixItems',
             ):
                 yield error
-    else:
+
+            # Remove evaluated prefixItems indexes
+            del instance[0:len(schema['prefixItems'])]
```
Julian, line 99: We can't ever mutate an instance we're given (it may not even be a mutable type)

jsonschema/_validators.py
```diff
@@ -70,13 +73,31 @@ def items(validator, items, instance, schema):
...
+        if not items:
+            if len(instance) > len(schema['prefixItems']):
+                yield ValidationError(
+                    "%r has more items than defined in prefixItems" % instance
+                )
+            else:
+                for error in validator.descend(
+                        instance,
+                        {'prefixItems': schema['prefixItems']},
+                        path='items__prefixItems',
```
Julian, line 86: This path doesn't look right -- it should be an indexable thing in the instance -- i.e. `instance[path]` should exist.

(I honestly don't know yet what prefixItems does though, so I'll have to check before making a specific recommendation, but if I can guess at what it does, the path should be a list index, probably the first one that's invalid)

jsonschema/_validators.py
Julian: I won't squash commits as part of merging -- definitely will preserve the commit history exactly as it was. But if the point here is that this is unrelated functionality, definitely split it into its own PR, but I suspect it isn't given prefixItems is new in these drafts.

jsonschema/validators.py
```diff
@@ -495,10 +585,7 @@ def __init__(
         self.handlers = dict(handlers)
 
         self._scopes_stack = [base_uri]
-        self.store = _utils.URIDict(
-            (id, validator.META_SCHEMA)
-            for id, validator in meta_schemas.items()
-        )
+        self.store = get_store()
```
Julian, line 588: Why are we turning what was previously local state to mutable global state?

jsonschema/_validators.py
nezhar: After refactoring I noticed that the `descend` is not reuired here, as this is already handled in the `prefixItems` function.

jsonschema/_validators.py
nezhar: Fixed

jsonschema/validators.py
nezhar: Since `draft2020-12` relies on references in takes some time to initialize the validator as it fetches all this URLs form remote. The tests for `draft2020-12` require over 30 min without this approach and takes several hours with the current pipeline configuration as there is a limit on paralel github actions.

Julian: That sounds like there are some URLs missing from Draft202012Validator's own schema store.

IIRC from the last time I did this myself, basically we need to be caching the vocabulary schemas as well (globally), not just the metaschema.

But yeah we shouldn't move to global state here for all URLs.

nezhar: Yes, the meta schema is loaded form the local file located in `draft2020-12.json`.  All references are loaded afterwards:

```
{
    "$id": "https://json-schema.org/draft/2020-12/schema",
    ...
    "allOf": [
        {"$ref": "meta/core"},
        {"$ref": "meta/applicator"},
        {"$ref": "meta/unevaluated"},
        {"$ref": "meta/validation"},
        {"$ref": "meta/meta-data"},
        {"$ref": "meta/format-annotation"},
        {"$ref": "meta/content"}
    ],
    ...
}
```

In this case I will try to load only the meta schemas in a global store, everything else will be loaded in the validator store.

nezhar: Now the vocabulary schemas are also stored in the `schemas` directory and loaded additionally to the store by adding `vocabulary_schemas` when creating a validator.

.github/workflows/coverage.yml
```diff
@@ -14,14 +14,14 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: pypy3
+        python-version: 3.9
```
Julian, line 17: Hi! Sorry for the delay on this one. I'll get you a full review soon, but this change (moving coverage to run on 3.9) is another good one to extract into its own tiny PR please! Appreciated.

nezhar: Moved to new PR https://github.com/Julian/jsonschema/pull/823

Also I cleaned the commits for the fix that has been applied with https://github.com/Julian/jsonschema/pull/820

Julian: Fantastic thanks! I'll have a look there too.

nezhar:
I made another contribution on the tests repostiory to cover some cases on unevaluated properties https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/500

The other contribution I made is kind of reverted, as the checks where moved to format.json - https://github.com/json-schema-org/JSON-Schema-Test-Suite/commit/ba1f1a772df864ce77d82a6da2aec7028bbccf02. This means that the coverage will drop as there are no more non string tests for the functions inside `_format.py`. Any sugestion on how to handle this better? I was thinking to add some new unit tests specific to the functions in this repository.

Julian:
Nice, thanks!

> This means that the coverage will drop as there are no more non string tests for the functions inside _format.py

I'll look closer, but from remembering that PR upstream, the tests are still there, just in a different file, no? It may be codecov is confused, but I don't think we should need any additional internal tests to cover that behavior, they should still all be there, but let me check.

setup.cfg
```diff
@@ -47,13 +48,14 @@ format_nongpl =
...
     uri_template
+    isoduration;python_version>'3.6'
 
 [options.entry_points]
 console_scripts =
     jsonschema = jsonschema.cli:main
 
 [options.package_data]
-jsonschema = schemas/*.json
+jsonschema = schemas/*.json, schemas/draft2020-12/*.json
```
Julian, line 58: I have to double check, but if you see one exists, we should just use the syntax that recursively includes everything in the `schemas` directory -- we'll have more drafts there.

Julian (commented):
Left another round of comments. Let me know if they make sense!

Thanks again for being so thorough.

nezhar:
> I'll look closer, but from remembering that PR upstream, the tests are still there, just in a different file, no? It may be codecov is confused, but I don't think we should need any additional internal tests to cover that behavior, they should still all be there, but let me check.

The tests suit for draft 2020-12 is split in `TestDraft202012` and `TestDraft202012Format` and the `format_checker` is defined only for the format tests. It's true, the tests are still available, but in `format.json`, so the non string case for duration or uuid won't be covered anymore.

Julian:
Ah. Maybe we should just run the file twice then? It's valid for both cases (or it should be)

nezhar:
> Ah. Maybe we should just run the file twice then? It's valid for both cases (or it should be)

Not really, there are some test cases that have an invalid format, but should be marked as valid, since it should run only as annotation: https://github.com/Julian/jsonschema/pull/817#issuecomment-868440805

I can add the `format.json` file to the format test and skip the test cases the contradiction, but in our case it would be enough to have just the few test cases added extra to `duration.json` and `uuid.json`

jsonschema/_utils.py
```diff
@@ -55,6 +58,19 @@ def load_schema(name):
     return json.loads(data.decode("utf-8"))
 
 
+def load_vocabulary(name):
+    """
+    Load all schema files from ./schemas/``name`` and return them as a list.
+    """
+    vocabulary = []
+    base_path = os.path.dirname(sys.modules["jsonschema"].__file__)
```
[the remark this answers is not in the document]
nezhar: This function is suposed to cover the loading of all the references provided in the meta schema (otherwise the URLs will be fetched from remote on every test, which cases them to run about 30 min/pipeline). Would it be an option to be explicit about which vocabularies should be loaded? This way the paths can be provided when creating the validator.

jsonschema/_utils.py
```diff
@@ -232,3 +248,148 @@ def uniq(container):
...
+    then, else, 'contains', 'unevaluatedItems', 'allOf', 'oneOf', 'anyOf'
+    """
+    if validator.is_type(schema, "boolean"):
+        return []
+    evaluated_indexes = []
+
+    if 'items' in schema:
+        return list(range(0, len(instance)))
+
+    if '$ref' in schema:
```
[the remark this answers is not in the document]
nezhar: I agree, this would also have a huge impact on performance and code reuse as `find_evaluated_item_indexes_by_schema` and `find_evaluated_property_keys_by_schema` are kind of repetitiv for certain tasks. I'm just afraid that this is a bigger change that has to be applied over all validator funtions somehow but let me know if you find a better approach :slightly_smiling_face:

jsonschema/_validators.py
```diff
@@ -111,10 +117,59 @@ def contains(validator, contains, instance, schema):
...
+        min_contains = schema['minContains']
+
+    if 'maxContains' in schema:
+        max_contains = schema['maxContains']
+
+    # minContains set to 0 will ignore contains
+    if min_contains == 0:
+        return
+
+    matches = len(list(
```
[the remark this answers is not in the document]
nezhar: The list is not required here and the sum is cleanner and better readable, I will change this :wink:

jsonschema/tests/test_format.py
```diff
@@ -86,3 +86,22 @@ def test_repr(self):
             repr(checker),
             "<FormatChecker checkers=['bar', 'baz', 'foo']>",
         )
+
+    def test_duration_format(self):
```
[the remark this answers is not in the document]
nezhar: Not anymore, this is related to the discusion we started [here](https://github.com/Julian/jsonschema/pull/817#issuecomment-880520779)

jsonschema/_legacy_validators.py
```diff
@@ -27,6 +27,37 @@ def dependencies_draft3(validator, dependencies, instance, schema):
                     yield ValidationError(message % (each, property))
 
 
+def dependencies_draft4_draft6_draft7(
+        validator,
```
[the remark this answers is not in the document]
nezhar: Fixed

jsonschema/_utils.py
```diff
@@ -232,3 +248,148 @@ def uniq(container):
...
+    Get all indexes of items that get evaluated under the current schema
+
+    Covers all keywords related to unevaluatedItems: items, prefixItems, if,
+    then, else, 'contains', 'unevaluatedItems', 'allOf', 'oneOf', 'anyOf'
+    """
+    if validator.is_type(schema, "boolean"):
+        return []
+    evaluated_indexes = []
+
+    if 'items' in schema:
```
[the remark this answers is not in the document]
nezhar: No problem, fixed it :wink:

jsonschema/_validators.py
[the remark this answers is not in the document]
nezhar: Done

jsonschema/_validators.py
```diff
@@ -111,10 +117,59 @@ def contains(validator, contains, instance, schema):
...
+    if not matches:
         yield ValidationError(
             "None of %r are valid under the given schema" % (instance,)
         )
+        return
+
+    if min_contains and max_contains is None:
+        if matches < min_contains:
+            yield ValidationError(
+                "Invalid number or matches of %r under the given schema, "
```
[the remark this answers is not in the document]
nezhar: Changed the message

jsonschema/_validators.py
```diff
@@ -111,10 +117,59 @@ def contains(validator, contains, instance, schema):
...
+                "expected min %d, got %d" % (
+                    instance, min_contains, matches
+                )
+            )
+        return
+
+    if min_contains is None and max_contains:
+        if matches > max_contains:
+            yield ValidationError(
+                "Invalid number or matches of %r under the given schema, "
```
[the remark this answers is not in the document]
nezhar: Done

jsonschema/_validators.py
```diff
@@ -111,10 +117,59 @@ def contains(validator, contains, instance, schema):
...
+                "expected max %d, got %d" % (
+                    instance, max_contains, matches
+                )
+            )
+        return
+
+    if min_contains and max_contains:
+        if matches < min_contains or matches > max_contains:
+            yield ValidationError(
+                "Invalid number or matches of %r under the given schema, "
```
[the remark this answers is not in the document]
nezhar: Changed the message, does this fit?

jsonschema/_validators.py
```diff
@@ -233,24 +288,35 @@ def maxLength(validator, mL, instance, schema):
         yield ValidationError("%r is too long" % (instance,))
 
 
-def dependencies(validator, dependencies, instance, schema):
+def dependentRequired(validator, dependentRequired, instance, schema):
+    """
+    Split from dependencies
```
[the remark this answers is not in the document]
nezhar: Since all functions are named after the spec in JSON Schema only a link to the Spec would make sense, but this is self-explanatory. So I removed the docstring.

jsonschema/_validators.py
```diff
@@ -233,24 +288,35 @@ def maxLength(validator, mL, instance, schema):
...
-            for error in validator.descend(
+        for each in dependency:
+            if each not in instance:
+                message = "%r is a dependency of %r"
+                yield ValidationError(message % (each, property))
+
+
+def dependentSchemas(validator, dependentSchemas, instance, schema):
+    """
+    Split from dependencies
```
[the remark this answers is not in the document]
nezhar: Removed

jsonschema/_validators.py
```diff
@@ -279,6 +345,41 @@ def ref(validator, ref, instance, schema):
...
 
+def dynamicRef(validator, dynamicRef, instance, schema):
+    _, fragment = urldefrag(dynamicRef)
+    scope_stack = validator.resolver.scopes_stack_copy
+
+    for url in scope_stack:
+        lookup_url = urljoin(url, dynamicRef)
+        with validator.resolver.resolving(lookup_url) as lookup_schema:
+            if "$dynamicAnchor" in lookup_schema \
+                    and fragment == lookup_schema["$dynamicAnchor"]:
```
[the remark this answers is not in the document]
nezhar: Done

jsonschema/_validators.py
```diff
@@ -279,6 +345,41 @@ def ref(validator, ref, instance, schema):
...
+
+
+def defs(validator, defs, instance, schema):
+    if not validator.is_type(instance, "object"):
+        return
+
+    if '$defs' in instance:
+        for definition, subschema in instance['$defs'].items():
+            for error in validator.descend(
+                    subschema,
```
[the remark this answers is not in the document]
nezhar: Fixed

jsonschema/_validators.py
```diff
@@ -384,3 +485,51 @@ def if_(validator, if_schema, instance, schema):
...
+
+def unevaluatedProperties(validator, unevaluatedProperties, instance, schema):
+    evaluated_property_keys = find_evaluated_property_keys_by_schema(
+        validator, instance, schema
+    )
+    unevaluated_property_keys = []
+    for property, subschema in instance.items():
+        if property not in evaluated_property_keys:
+            for error in validator.descend(
+                    instance[property],
```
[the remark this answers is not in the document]
nezhar: Fixed

jsonschema/_validators.py
```diff
@@ -384,3 +485,51 @@ def if_(validator, if_schema, instance, schema):
...
+        yield ValidationError(error % extras_msg(unevaluated_property_keys))
+
+
+def prefixItems(validator, prefixItems, instance, schema):
+    if not validator.is_type(instance, "array"):
+        return
+
+    for k, v in enumerate(instance[:min(len(prefixItems), len(instance))]):
+        for error in validator.descend(
+                v, prefixItems[k], schema_path="prefixItems"
```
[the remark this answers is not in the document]
nezhar: Fixed

setup.cfg
nezhar: I could not find a clear definition on the recursive syntax, the one I added now seems to be working.

jsonschema/validators.py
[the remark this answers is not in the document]
robherring: I'm still seeing an issue with this.

jsonschema/validators.py
```diff
@@ -692,14 +681,12 @@ def resolving(self, ref):
 
     def _finditem(self, schema, key):
         results = []
+        if key in schema:
+            results.append(schema)
 
-        if isinstance(schema, dict):
```
robherring: Looks like you reintroduced the issue with handling schema is None here.

This commit history where you add code and then refactor it again in later commits makes it harder for reviewers and harder to trace the history once merged. Make your commits clean in that they do 1 thing. Put any preparatory refactoring that shouldn't change behavior and any fixes first. You're rebasing the branch anyways. As @Julian mentioned, all this won't be squashed when merged.

jsonschema/validators.py
```diff
@@ -145,16 +145,23 @@ def schema_validators(self, schema):
...
 
             ref = schema.get(u"$ref")
             if ref is not None:
                 # We make sure $ref is evaluated first if available in schema
                 validators = [(u"$ref", ref)]
 
                 # Apply all remaining validators for newer drafts
-                if not any(isinstance(self, x) for x in legacy_cls):
-                    validators += [(x, schema[x]) for x in schema if x not in ["$ref"]]
+                if not any(isinstance(self, x) for x in legacy_validator_cls):
```
robherring, line 177: This still doesn't work for me as I have a class extended from Draft7Validator. But Draft7Validator is not a type. It's type is 'Validator', right? So does this ever work?

The relevant code for me is here:
https://github.com/devicetree-org/dt-schema/blob/master/dtschema/lib.py#L668

jsonschema/validators.py
nezhar: Sorry for that, that happened as I tried to fix some things in order to push the coverage of the patch. I will go over the history and also fix that part.

nezhar: Fixed the mentioned issue, also I cleaned the commit history :slightly_smiling_face:

jsonschema/validators.py
nezhar: The existing test cases don't fail with this construct, but I will check this deeper. Maybe you have a sugestion how to handle this better :slightly_smiling_face:

robherring: I think you need a parameter for create() to handle the different behavior. Then any validator created with extend() inherits the same behavior if not specified.

robherring:
I've noticed another issue with the $ref handling and error schema paths. If you have something like this:

{
  "$defs": { 
    "foo": { "required": [ "bar" ] }
  },
  "properties": {
    "aprop": {
      "$ref": "#/$defs/foo",
      "required": [ "baz" ]
    }
  }
}

If either 'required' fails, the schema path is the same: (properties, aprop, required)

With draft7 and earlier, the $ref was always by itself, so the error path would always be unique. Now the schema path is ambiguous.  I think '$ref' needs to be stored in the schema path. But that probably needs to be conditional on the version to not break existing users.

jsonschema/validators.py
nezhar: Moved this into two functions which can now be used when creating a validator. What do you think about this approach?

@Julian can `_schema_validators_draft7_draft6_draft_4_draft3` be moved to `_legacy_validators.py` or would it maked sence to create a `_legacy` module?

Julian: If I follow, the code's changed, in the current version what you have looks good to me @nezhar, but let's call it `applicable_validators`, and then yes you can move the one that ignores non-`$ref` to `_legacy_validators`.

While I see it though, @robherring just FYI, subclassing validators isn't part of their public API (and isn't supported), so there's no promises stuff continues to work there. If it works now though I guess great.

jsonschema/validators.py
```diff
@@ -120,8 +155,10 @@ class Validator:
 
         VALIDATORS = dict(validators)
         META_SCHEMA = dict(meta_schema)
+        VOCABULARY_SCHEMAS = list(vocabulary_schemas)
         TYPE_CHECKER = type_checker
         ID_OF = staticmethod(id_of)
+        SCHEMA_VALIDATORS = staticmethod(schema_validators)
```
Julian, line 161: We can make this private as well.

nezhar:
> I've noticed another issue with the $ref handling and error schema paths. If you have something like this:
> 
> ```
> {
>   "$defs": { 
>     "foo": { "required": [ "bar" ] }
>   },
>   "properties": {
>     "aprop": {
>       "$ref": "#/$defs/foo",
>       "required": [ "baz" ]
>     }
>   }
> }
> ```
> 
> If either 'required' fails, the schema path is the same: (properties, aprop, required)
> 
> With draft7 and earlier, the $ref was always by itself, so the error path would always be unique. Now the schema path is ambiguous. I think '$ref' needs to be stored in the schema path. But that probably needs to be conditional on the version to not break existing users.

I created some test cases for this https://github.com/Julian/jsonschema/pull/817#issuecomment-881550313

@robherring Are there any other cases that I'm missing in the tests?

jsonschema/validators.py
nezhar: Renamed to `applicable_validators` and moved to old code to `_legacy_validators`

jsonschema/validators.py
nezhar: Done

jsonschema/validators.py
```diff
@@ -568,17 +652,6 @@ def base_uri(self):
         uri, _ = urldefrag(self.resolution_scope)
         return uri
 
-    @contextlib.contextmanager
```
Julian, line 571: Is there a reason this is being removed?

jsonschema/validators.py
```diff
@@ -312,6 +336,8 @@ def extend(validator, validators=(), version=None, type_checker=None):
     type_checker=_types.draft3_type_checker,
     version="draft3",
     id_of=lambda schema: schema.get(u"id", ""),
+    applicable_validators=_legacy_validators
```
Julian, line 339: Let's call this `_legacy_validators.ignore_ref_siblings`.

jsonschema/validators.py
```diff
@@ -63,12 +68,30 @@ def _id_of(schema):
...
+
+
 def create(
     meta_schema,
+    vocabulary_schemas=(),
     validators=(),
     version=None,
     type_checker=_types.draft7_type_checker,
     id_of=_id_of,
+    applicable_validators=_applicable_validators
```
Julian, line 94: It's short enough to just inline this.

```suggestion
    applicable_validators=lambda schema: schema.items(),
```

jsonschema/validators.py
```diff
@@ -26,6 +26,7 @@
 
 validators = {}
 meta_schemas = _utils.URIDict()
+vocabulary_schemas = _utils.URIDict()
```
Julian, line 29:
```suggestion
_VOCABULARIES = _utils.URIDict()
```

jsonschema/validators.py
nezhar: I think I removed this in order to increase the coverage as this code is not used anywhere in the library. I can create a dedicated PR for this or simply revert it if this is a common use case of the lib.

jsonschema/validators.py
nezhar: Done

jsonschema/validators.py
nezhar: I agree, code is updated

jsonschema/validators.py
nezhar: Also changed to your recomandation :slightly_smiling_face:

Julian:
Thanks @nezhar! This is likely getting close to merging, I'll have another look and perhaps leave one more round of comments but then likely merge and move forward from there.

There are definitely some changes I want to make (specifically around vocabulary support), but we can do so after things are merged, and perhaps after a beta release.

jsonschema/validators.py
Julian: It's part of the public API of the object, so we can't remove it without deprecating it.

jsonschema/tests/test_jsonschema_test_suite.py
```diff
@@ -398,3 +457,65 @@ def leap_second(test):
...
+    DRAFT202012.optional_tests_of(name="ecmascript-regex"),
+    DRAFT202012.optional_tests_of(name="float-overflow"),
+    DRAFT202012.optional_tests_of(name="non-bmp-regex"),
+    DRAFT202012.optional_tests_of(name="refOfUnknownKeyword"),
+    Validator=Draft202012Validator,
+    skip=lambda test: (
+        narrow_unicode_build(test)
+        or ecmascript_regex_validation(test)
+        or skip(
+            message="ToDo: Extend validation",
```
Julian, line 474: I'm ok if we are missing support for something here, but can you change the message or add a ticket to indicate what we are missing and what'd be required to get it?

jsonschema/tests/test_jsonschema_test_suite.py
```diff
@@ -125,6 +128,62 @@ def leap_second(test):
     )(test)
 
 
+def ecmascript_regex_validation(test):
```
Julian, line 131: Other drafts I think just don't run this file at all (because we don't support it), so I think it's a bit less verbose to just remove all this and not run that file for 2020 as well.

jsonschema/validators.py
nezhar: I agree, this can be done in another PR. I restored the method.

jsonschema/tests/test_jsonschema_test_suite.py
nezhar: This is true, I just saw now that `ecmascript` was ignored for all the other drafts. I removed it from the optional tests.

jsonschema/tests/test_jsonschema_test_suite.py
nezhar: There are two additional cases that fail that where introduced with the latest update of the tests. I'm kind of having a hard time understanding how the dynamic scopes are suposed to work, maybe they need to be split when loading `anschors` and `dynamicAnchors`. I'm still lokking into this, for now I fixed the message.

setup.cfg
mik-laj: Recursive glob is not supported by setuptools: https://github.com/pypa/setuptools/issues/1806

Julian:
@nezhar I've merged this just now -- apologies for this taking so long, and quite grateful for the work!

There definitely still are things I want to fix, you'll see I've just sent a PR upstream because the UUID formatting validation here isn't correct, and I think the relative pointer validation may not be either.

But I want to get this merged and fix things on the main branch going forward, so we at least aren't sitting here with the PR waiting.

I'll follow up perhaps here if there are any major updates/changes that need doing, just in case you have comments on them, but thanks again!

jsonschema/_validators.py
```diff
@@ -279,6 +336,41 @@ def ref(validator, ref, instance, schema):
...
+                    yield error
+                break
+    else:
+        with validator.resolver.resolving(dynamicRef) as lookup_schema:
+            subschema = lookup_schema
+            for error in validator.descend(instance, subschema):
+                yield error
+
+
+def defs(validator, defs, instance, schema):
```
Julian, line 359: In playing around myself here the past few days, I'm pretty sure something is wrong here, but I don't know what yet.

`defs` doesn't really have any behavior, it's just a rename of `definitions` (and thereby just a standard place to put definitions used elsewhere in the schema), so there's no validation that should happen when encountering it.

But I see tests fail if I remove this -- I *think* that's an issue with the `dynamicRef` support, not with `def`, the test that fails is one that validates an invalid schema against the metaschema.

If anyone sees where the issue is, let me know, otherwise will keep playing.

PR #819, opened 2021-06-25
Fixes validation for uri-template format

nezhar:
Fixes the validation of `uri-template` using https://pypi.org/project/uri-template/


Julian:
Thanks!

PR #820, opened 2021-06-29
Fixes unique and equal checks

nezhar:
Extends the equal check for dictionaries and lists.

This solves #686, but it's also usefull for newer spec implementations such as draft 2020-12

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -158,10 +158,46 @@ def ensure_list(thing):
     return thing
 
 
+def dict_equal(one, two):
+    """
+    Check if two dicts are the same using `equal`
+    """
+    if len(one.keys()) != len(two.keys()):
+        return False
+
+    for key in one:
+        if key not in two:
+            return False
+        if not equal(one[key], two[key]):
+            return False
+
+    return True
+
+
+def list_equal(one, two):
+    """
+    Check if two lists are the same using `equal`
+    """
+    if len(one) != len(two):
+        return False
+
+    for i in range(0, len(one)):
+        if not equal(one[i], two[i]):
+            return False
+
+    return True
+
+
 def equal(one, two):
     """
     Check if two things are equal, but evade booleans and ints being equal.
     """
+    if isinstance(one, list) and isinstance(two, list):
+        return list_equal(one, two)
+
+    if isinstance(one, dict) and isinstance(two, dict):
+        return dict_equal(one, two)
+
     return unbool(one) == unbool(two)
 
 
@@ -192,14 +228,22 @@ def uniq(container):
         try:
             sort = sorted(unbool(i) for i in container)
             sliced = itertools.islice(sort, 1, None)
+
             for i, j in zip(sort, sliced):
+                if isinstance(i, list) and isinstance(j, list):
+                    return not list_equal(i, j)
                 if i == j:
                     return False
+
         except (NotImplementedError, TypeError):
             seen = []
             for e in container:
                 e = unbool(e)
-                if e in seen:
-                    return False
+
+                for i in seen:
+                    if isinstance(i, dict) and isinstance(e, dict):
+                        if dict_equal(i, e):
+                            return False
+
                 seen.append(e)
     return True
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -217,33 +217,27 @@ def uniq(container):
     """
     Check if all of a container's elements are unique.
 
-    Successively tries first to rely that the elements are hashable, then
-    falls back on them being sortable, and finally falls back on brute
-    force.
+    Successively tries first to rely that the elements are being sortable
+    and finally falls back on brute force.
     """
-
     try:
-        return len(set(unbool(i) for i in container)) == len(container)
-    except TypeError:
-        try:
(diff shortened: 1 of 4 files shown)


Julian:
There's some interactions with `TypeChecker` we need to account for -- specifically, we need a new test (which I'm pretty sure will fail) where someone does `type_checker.redefine("array", some_collections.Sequence_that_does_not_inherit_from_list)`, which is supported use of the library, and then which still contains `1` and `True` in that non-list-inheriting sequence.

nezhar:
> There's some interactions with `TypeChecker` we need to account for -- specifically, we need a new test here internally (which I'm pretty sure will fail) where someone does `type_checker.redefine("array", some_collections.Sequence_that_does_not_inherit_from_list)`, which is supported use of the library, and then which still contains `1` and `True` in that non-list-inheriting sequence.
> 
> And then the same for dict.

Does [this](https://github.com/Julian/jsonschema/pull/820/files#diff-4f792d03ba8407d680a5588f2d82b77029c3ecd169c26c46aec73fff47daa930R1091) go in the right direction? Is there anything similar in the code you can point me to?

Julian:
Sorry for being terse -- yes that's in the right direction, but now the point is not to test the behavior in that test case itself (which is already covered elsewhere in the suite), it's whether #686 is fixed there. I.e. whether `deque([[False], [0]]), {'type': 'array', 'uniqueItems': True})` also fails properly now (because it contains nonunique items). And same for `[MyMapping({"a": 0}), MyMapping({"a": False})]`

Let me know if that makes sense, if not will write something longer up. And thanks again.

nezhar:
> Sorry for being terse -- yes that's in the right direction, but now the point is not to test the behavior in that test case itself (which is already covered elsewhere in the suite), it's whether #686 is fixed there. I.e. whether `deque([[False], [0]]), {'type': 'array', 'uniqueItems': True})` also fails properly now (because it contains nonunique items). And same for `[MyMapping({"a": 0}), MyMapping({"a": False})]`
> 
> Let me know if that makes sense, if not will write something longer up. And thanks again.

Thanks for the explanation. I added the examples you mentioned in the test :slightly_smiling_face: 

Maybe I miss something, but according to these tests this should be valid: https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/master/tests/draft2020-12/uniqueItems.json#L21

This is what actually fails
```
validator.validate([deque([False]), deque([0])])

validator.validate([MyMapping('a', 0),  MyMapping('a', False)])
```

According to the spec this should be valid, right?

Julian:
Yes, sorry, I was typing too quickly -- valid, not invalid!

I didn't try them with the change here but yeah we want to make sure the fix covers those as well.

nezhar:
> Yes, sorry, I was typing too quickly -- valid, not invalid!
> 
> I didn't try them with the change here but yeah we want to make sure the fix covers those as well.

I removed the first check in `uniq` as `set` would interfere in the equal check. Tests seem to be ok now, what do you think?

jsonschema/_utils.py
```diff
@@ -181,25 +217,27 @@ def uniq(container):
...
-        try:
-            sort = sorted(unbool(i) for i in container)
-            sliced = itertools.islice(sort, 1, None)
-            for i, j in zip(sort, sliced):
-                if i == j:
+        sort = sorted(unbool(i) for i in container)
+        sliced = itertools.islice(sort, 1, None)
+
+        for i, j in zip(sort, sliced):
+            return not list_equal(list(i), list(j))
```
Julian, line 228: I think this will now likely fail for triply-nested non-list containers.

It also will now do a lot of copying (every object will always be copied).

It's a bit better if we instead just use `isinstance(collections.Sequence` or `collections.Mapping` I suspect.

But yeah please check the additional nesting, I suspect we need even better tests that uncover that. Really the nice thing would be a hypothesis test that generates arbitrarily nested containers which it puts `0`s and `False`s inside and ensures they're always considered unique. If you know how to do that that'd be great, otherwise yeah I suspect at least another manual test is needed.

jsonschema/_utils.py
```diff
@@ -181,25 +217,27 @@ def uniq(container):
...
+
+    except (NotImplementedError, TypeError):
+        seen = []
+        for e in container:
+            e = unbool(e)
+
+            for i in seen:
+                if isinstance(i, dict) and isinstance(e, dict):
```
Julian, line 236: This line too I'm suspicious of if it still contains `isinstance(, dict)` -- it'll fail for non-dict mappings then, so we need to exercise that case as well.

Let me know if that makes sense.

jsonschema/_utils.py
nezhar: You are absolutely right, a tripple nesting caused new failures, this was related to the `equal` function where the behavior was not extended. I refactored the tests and added some more cases as I'm not really sure how to create a data generator here and I'm afraid this makes debugging harder.

jsonschema/_utils.py
nezhar: This has been removed and it makes use of the `equal` function instead.

Julian:
Apologies for the delay here. I made some small tweaks, but have merged this.

Much appreciated!

I'll get you some feedback on the other PR next!

Issue #569, opened 2019-06-12
Make $schema lookup more liberal for empty fragment URLs


jswinarton:
I have a valid schema document that contains the following `$schema` value:

```
{
  "$schema": "http://json-schema.org/draft-07/schema/#"
}
```

When validating an instance against this schema, I get the following warning: 

```
/usr/local/lib/python3.7/site-packages/jsonschema/validators.py:893: DeprecationWarning: The metaschema specified by $schema was not found. Using the latest draft to validate, but this will raise an error in the future.
```

I discovered that I can suppress the error by changing removing the trailing slash and hash from the schema URI, e.g. `http://json-schema.org/draft-07/schema`. 

I think the validator should allow for some variations on URI formats--the JSON Schema documentation specifies the `$schema` value _with_ the trailing slash and hash, so I believe that will be the most common format users attempt.

Julian:
This looks reasonable -- I'm not sure we want to accept multiple variations on its own, but really it depends what the JSON Schema spec says (if anything) about whether those URLs are the same or not as far as its concerned. Probably you're right that they are.

So yeah definitely thanks for filing.

scottj97:
I second this request. The "official" guide [recommends `http://json-schema.org/schema#`](https://json-schema.org/understanding-json-schema/reference/schema.html) but that gives this warning from jsonschema. In this case, removing the trailing hash mark does not help. It looks like I must specify an explicit version (and also remove the trailing slash+hash).

Julian:
That... I think is actually a bad recommendation by our tutorial.

JSON Schema does not make any forwards compatibility guarantees, and has broken some significant compatibility in recent versions, so I can't think of any good reason a schema author should declare that URL.

We should discuss upstream in the tutorial repo, but yeah that one is way less reasonable sounding to me than the original request here.

jayeye:
Since a json schema is itself a json document that should match the "metaschema" in `http://json-schema.org/schema`, I should be able to use the validator to make sure my schema is correct (at least syntactically). I can validate json data against a schema I wrote, but I cannot validate my schema against the metaschema. 

Also: maybe my understanding of JSON Schema is lacking, but shouldn't validating the metaschema against itself work?
`curl -s http://json-schema.org/schema > ms`
then in python: `jsonschema.validate(instance=json.load(open('ms')), schema=json.load(open('ms')))`

I get an exception:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

Issue #692, opened 2020-05-25
Q: Is it possible to retrieve a json schema with refs resolved?


somiandras:
I want to generate some documentation/description (probably some input forms) from a json schema that contains `$ref`s.  Is there a way in jsonschema to retrieve a json schema with `$ref`s resolved and inlined, so I can iterate through the schema?

So from a schema like this:

```json
{
  "definitions": {
    "some_object": {
      "type": "object",
      "title": "My definition",
      "description": "Some description",
      "properties": {
        "name": {
          "type": "string"
        },
        "value": {
          "type": "number"
        }
      }
    }
  },
  "properties": {
    "list_of_objects": {
      "type": "array",
      "title": "Some stuff",
      "description": "List of stuff",
      "items": {"$ref": "#/definitions/some_object"}
    }
  }
}
```

I want to have this structure in my Python object (notice that the `{$ref:...}` section is replaced by the actual definition):

```json
{
  "definitions": {
    "some_object": {
      "type": "object",
      "title": "My definition",
      "description": "Some description",
      "properties": {
        "name": {
          "type": "string"
        },
        "value": {
          "type": "number"
        }
      }
    }
  },
  "properties": {
    "list_of_objects": {
      "type": "array",
      "title": "Some stuff",
      "description": "List of stuff",
      "items": {
        "type": "object",
        "title": "My definition",
        "description": "Some description",
        "properties": {
          "name": {
            "type": "string"
          },
          "value": {
            "type": "number"
          }
        }
      }
    }
  }
}
```

I am aware that with self-referencing schemas this will lead to an infinite structure, but hopefully this is feasible with reasonable limitations (eg. excluding self-refs).

calvin:
jsonref https://pypi.org/project/jsonref/ resolves references so that you can have fully expanded schemas.

jpmckinney:
I think @Julian previously decided that this belongs in a separate library: https://github.com/Julian/jsonschema/pull/419

I would like a helper, though, as https://github.com/gazpachoking/jsonref seems to not be actively maintained (all its tests are failing on CI, so it's not clear that it would even be easy to take maintenance of).

Issue #252, opened 2015-10-22
Add a helper for extending a validator to reject unknown formats


jmagnusson:
It would be nice if a strict mode could be turned on for `FormatChecker`. By _strict_ I mean that `jsonschema` will raise an error if it encounters a format that it doesn't recognize.

Julian:
Hey! Thanks for the report.

I'm generally not in favor of `strict` modes (and more generally of boolean flags). Specifically here, how will someone know to use it unless they've already guessed that their problem is a missing format checker?

Also, this can already be done -- use `jsonschema.validators.extend`, and add a `oneOf / enum` to the metaschema for format. I can elaborate if you'd like, shouldn't be more than ~3 lines of code.

jmagnusson:
Hi Julian! Thanks for your input.

Not sure I understand your reasoning, but it's early morning here so that might be why :wink:

The reason I opened this issue is because I have a REST API where users push in data, according to a specific JSON schema. I want them to get a 400 Bad Request if I encounter data that I will not be able to parse, like a date not formatted according to ISO 8601.

If this can already be done I would love to see how. Thanks man!

Julian:
Your users are sending you schemas to use or instances to validate? It sounds like you're saying instances?

jmagnusson:
Yup, they are sending instances for me to validate.

Julian:
Hm, OK, then I'm confused :), how does that relate to this ticket?

jmagnusson:
That's because I was confused... Ignore my previous messages.

What I'm looking for as I stated originally is a way to become aware of formats that will not be validated. As a developer it's hard to notice if jsonschema just ignores the `format`. Example:

```
In [1]: import jsonschema
In [2]: print(jsonschema.validate("xy", {"type": "string", "format" : "time-hour"}, format_checker=jsonschema.FormatChecker()))
None
```
How would I catch this? As an author of the schemas I want to be aware of which format checks I need to create.

Julian:
I see, OK, so now possibly my first comment might make sense:

The reason I don't generally think strict modes are useful is (among other things), as a developer, how would you ever know that you *need* strict mode unless you already knew that your instance contained formats that didn't exist? It sounds like for you that you just want to do this once possibly? Or once when you get a new schema, to see if there's any formats that someone put in there that you need to implement?

You can extend a metaschema to define that formats other than the known ones are invalid. E.g.:

```
>>> Validator = jsonschema.validators.create(
...     meta_schema={
[...]

PR #822, opened 2021-07-10
Ignore scheme and trailing slashes in meta_schemes key

lcvisser:
This PR intends to resolve https://github.com/Julian/jsonschema/issues/569 by ignoring the scheme (http, https) and any trailing slashes in the `$schema` uri.

### DIFF
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -3,6 +3,7 @@ Creation and extension of validators, with implementations for existing drafts.
 """
 from collections.abc import Sequence
 from functools import lru_cache
+from re import A
 from urllib.parse import unquote, urldefrag, urljoin, urlsplit
 from urllib.request import urlopen
 from warnings import warn
@@ -810,3 +811,35 @@ def validator_for(schema, default=_LATEST_VERSION):
             stacklevel=2,
         )
     return meta_schemas.get(schema[u"$schema"], _LATEST_VERSION)
+
+
+def set_validator_for(schema, validator, meta_schema_id=None):
+    """
+    Register a validator class for the given schema.
+
+    Uses the `$schema` property of the given schema as key in the validator
+    lookup table, unless `meta_schema_id` is given to override this.
+
+    This method allows you to set a validator for a particular schema in cases
+    where automatic lookup fails, for example when a schema specificies a
+    `$schema` that is not resolvable.
+
+    Arguments:
+
+        schema (collections.abc.Mapping):
+
+            the schema to register a validator for
+
+        validator:
+
+            the validator to use for this schema
+
+        meta_schema_id:
+
+            if not `None`, the given value is used as key in the lookup table.
+    """
+    if meta_schema_id is None:
+        meta_schema_id = schema.get(u"$schema", u"")
+
+    if meta_schema_id:
+        meta_schemas[meta_schema_id] = validator
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -3,7 +3,6 @@ Creation and extension of validators, with implementations for existing drafts.
 """
 from collections.abc import Sequence
 from functools import lru_cache
-from re import A
 from urllib.parse import unquote, urldefrag, urljoin, urlsplit
 from urllib.request import urlopen
 from warnings import warn


lcvisser:
Should I do something else?

Julian:
Hey! Sorry for not responding here -- the key thing we need to verify is whether this change is actually fully correct under the spec.

Specifically, `$id`s in the spec are, as far as I vaguely recall, meant to be identifiers, nothing more -- and the URI specifications leave how you treat trailing slashes up to the individual use case -- so whether `https://example.com/foo` and `https://example.com/foo/` are indeed the same is up to the spec I believe.

I think what we need to do here is ask this question upstream, in perhaps the [slack](json-schema.slack.com/) or discussion board.

lcvisser:
Thanks for the reply.

I understand the motivation to stick to the spec, but the original issue reported a URI with trailing slash and hash and [my motivation](https://docs.zhinst.com/hdawg/commandtable/v2/schema) has a trailing hash in it, so there is a practical aspect to consider also.

If it turns out to be against spec to have trailing slashes/hashes, would it be acceptable to have something like a `uri-strictness` flag or something, to deal with stubborn reality?

Julian:
If it's against the spec, it'd seem quite easy to do external from the validation process. I.e. `def canonicalize_urls(schema) -> schema` which just walks all the `$id` fields and removes trailing slashes and empty fragments and returns the resulting normalized schema.

I could possibly be convinced that'd be worth adding somewhere (there's some precedence in wanting this kind of thing in e.g. #692 or #252) but yeah not as a flag.

lcvisser:
I asked on their [Slack channel](https://json-schema.slack.com/archives/C5CF75URH/p1628006832022300); I'll reproduce the response here for completeness:

> Stripping/ignoring the scheme is definitely against spec (specifically RFC 3986). It happens to be safe (most of the time) to assume that "https" URIs are referencing the same URIs as their HTTPS counterparts.
But a) that's not guaranteed in all cases, and b) there are more schemes than just http/https. I've seen people do goofy things with schemes. Best not to make assumptions about those things.

> IMO it would only be valid to accept other forms for those URIs if the evaluator actually fetched that document via the network with that URI and parsed it to be sure it’s the same thing. it should not simply assume that a URI that looks similar is referencing the same metaschema as the official URI.
..and not all evaluators support fetching documents via the network (nor are they required to)

> Yeah, plus the standard metaschema URIs (including the ones in their http forms) are "well known", so most implementations I've seen don't even need to fetch them over the network.
So if you're worried about making insecure http requests, it's generally not an issue for the official meta schemas (since implementation libraries often ship with the metaschemas pre-fetched for your convenience.)
Not a guarantee though, so double check any implementation you're using.

I'll think about your proposal for a `canonize_uri()` utility-like function.  Would it be better to make a new MR for that or reuse this one?

Julian:
Did you ask specifically about trailing fragments too?

Ignoring schemes I was sure wasn't going to fly, so unsurprised there, but those seemed more likely to be allowed. Is an empty fragment considered the same URL as fragmentless?

As for the helper function, either way is fine with me on reusing this one or creating a new one -- whatever is easiest for you.

And thanks!

lcvisser:
Sorry, should have included my question also:

> Hi; the JSON-schema Python package has an open issue (#569) regarding what should be accepted for $schema . json-schema.org uses e.g. http://json-schema.org/draft-07/schema#, but "in the wild" there may be https instead of http and/or trailing slash before the # (i.e. http://json-schema.org/draft-07/schema/# )  and/or a missing # at the end, and combinations thereof. Would it be against spec to be liberal in what to accept for $schema? In particular: this pull request proposes to use a canonical URI, stripping scheme and empty fragments - would that violate anything? Seems a safe change to me, but better safe than sorry and double check.

I don't know enough about this, to be honest. It seems that the safest way forward is to only accept the format that is in the schema `http://json-schema.org/draft-07/schema#` and then provide a utility function that liberally "converts" to that. The doctoring of that utility could provide the necessary warnings and caveats as put forward in the Slack channel.

lcvisser:
Sorry for the lack of updates.

The longer I thought about it, the more I got convinced I was getting myself into a pickle. I agree that the least-intrusive way to solve this is to add an utility function that allows the user to override a particular `$schema` lookup. The rationale being that the provider of the schema should stick to the standard, hence silently eating schemes or trailing slashes is indeed wrong, as you argued.

I've updated the PR with the utility function. Here is an MWE:

```python
import jsonschema


schema = {
    "$schema": "https://json-schema.org/draft-04/schema/#",
    "definitions": {
        "my_type": {
            "type": "number",
            "maximum": 10,
            "exclusiveMaximum": True
        }
    },
    "properties": {
        "my_number": {
            "$ref": "#/definitions/my_type"
        }
    }
}

instance = {
    "$schema": "https://json-schema.org/draft-04/schema/#",
    "my_number": 7
}


# Fails
try:
    jsonschema.validate(instance, schema)
except jsonschema.exceptions.SchemaError as exc:
    print("failed")

# Succeeds
jsonschema.validators.set_validator_for(schema, jsonschema.validators.Draft4Validator)
jsonschema.validate(instance, schema)
print("success")
```

I will fix the merge conflict and add unit tests together with other comments you may have.

lcvisser:
Any feedback @Julian ?

Julian:
Apologies, I am inordinately busy, even busier than usual, as academic semester is starting -- I'll try to get you some feedback but it likely won't be before middle of the week or early next. Sorry -- please do ping me if you have to wait any longer.

lcvisser:
> Apologies, I am inordinately busy, even busier than usual, as academic semester is starting -- I'll try to get you some feedback but it likely won't be before middle of the week or early next. Sorry -- please do ping me if you have to wait any longer.

No worries! Just wanted to make sure that it was noticed.

Julian:
@lcvisser -- looking at this briefly, my honest reaction is -- is this really any better than just recommending people do `schema["$schema"] = Draft4Validator.META_SCHEMA["$id"]`? To me it seems obviously not personally, but willing to hear what you think.

lcvisser:
@Julian Personally, I think the utility functions is more user friendly, as it is discoverable and documented via the docstring. But it's a personal preference, so if you'd like the code base as minimal as possible, than that's a valid argument too of course. In that case, I'd propose to reject and close this MR (and arguably also the linked issue).

Julian:
I think for now will close -- I appreciate your effort and research. We can revisit in the future if needed.

PR #823, opened 2021-07-12
Use python 3.9 for coverage

nezhar:
Move coverage pipeline from PyPy (which currently implements Python 3.6) to CPython with Python 3.9.


Julian:
Thanks again!

PR #833, opened 2021-08-23
Remove shebangs from nonexecutable scripts

hrnciar:
When packaging jsonschema and poetry-core (which bundles it) in Fedora, we have
realised that there are nonexecutable files with a shebang line.

It seems that the primary purpose of these files is to be imported from Python code
and hence the shebangs appear to be unnecessary.

Shebangs are hard to handle when doing downstream packaging, because it makes
sense for upstream to use `#!/usr/bin/env` python while in the RPM package,
we need to avoid that and use a more specific interpreter. Since the
shebangs were unused, I propose to remove them to avoid problems.

### DIFF
--- a/jsonschema/benchmarks/issue232.py
+++ b/jsonschema/benchmarks/issue232.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 """
 A performance benchmark using the example from issue #232.
 
--- a/jsonschema/benchmarks/json_schema_test_suite.py
+++ b/jsonschema/benchmarks/json_schema_test_suite.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 """
 A performance benchmark using the official test suite.
 


Julian:
Hi.

I'm happy to merge this I suppose, though I'm not sure I 100% follow -- if these were `+x`'ed (which I suppose I didn't notice they weren't), they'd be easier to package? They're not really meant to be imported, they're indeed meant to be executed -- it so happens I apparently executed them via an [explicit interpreter](https://github.com/Julian/jsonschema/blob/cabb008b027bce9d9fa7fac97390ba2b4fb3f57f/tox.ini#L41-L42) but on the command line locally I'd be equally likely to run them directly (which of course I've apparently never done given they're yeah, not `+x`'ed, but that's mostly a sign of how poor the current benchmarking situation is unfortunately).

Anyways, yeah, happy to merge this since it's fairly low impact, but I can't promise they won't come back somewhere in the future (pip rewrites shebangs -- I know for better distro support there's tools being encouraged like [build](https://pypi.org/project/python-build/) and [installer](https://pypi.org/project/installer/) -- it may be tangential but if shebang rewrite support isn't present in them that's probably a good upstream discussion to have).

PR #841, opened 2021-09-28
Don't build universal wheels for Python 3-only

hugovk:
Universal wheels are for when you need to support both Python 2 and 3:

https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels

Before:

* `jsonschema-0.1.dev1+gdbda4f7-py2.py3-none-any.whl`

After:

* `jsonschema-0.1.dev2+g97057ed-py3-none-any.whl`

---

Also ignore some build dirs generated by running `python -m build` locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,8 @@ _cache
 _static
 _templates
 
+*.egg-info/
+build
+dist
+
 TODO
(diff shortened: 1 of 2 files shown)


Julian:
Thanks!

PR #850, opened 2021-10-01
validators: Fix dependentSchemas when instance is not an object

robherring:
Just like 'dependentRequired' and 'dependencies', 'dependentSchemas'
needs to handle instance not being an object.


Julian:
Thanks, appreciated -- we need a test for this upstream in the test suite for this -- it belongs here: https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/8891d81074300ebb5e551cbda8796e686e586c90/tests/draft2020-12/dependentSchemas.json but should also go in the 2019-09 file.

Can you submit a PR there in that repo for that which I can then merge in here?

PR #856, opened 2021-10-12
Add extra_info on ValidationError for required missing property

TsankoBg:
This PR adds an `extra_info` attribute on `ValidationError`, which is used to indicate the missing property in errors from the `required` validator.

This should resolve issue #119, and can be used for adding details to other validators as well, in the future.


Julian:
@TsankoBg with sincere apologies (since you yourself did everything right here), some other things have recently convinced me to bite the bullet on finally doing https://github.com/python-jsonschema/jsonschema/issues/119#issuecomment-256253014 (i.e. doing this via separating out `ValidationError`). I have some code locally which does so, but needs some tidying. I'm going to close this as yeah I think that's a better direction long term, but I sincerely appreciate sending the PR for what is the longest-running issue here :(

PR #862, opened 2021-10-18
Fix prefixItems index reporting in ValidationError

jhashkes:
- Fixed an issue when a prefixItems mismatch was identified, ValidationError was neglecting to report the instance index in json_path and the various deque paths
- Updated unit test


Julian:
Thanks!

Issue #876, opened 2021-11-04
4.2.0 contains versionless dep on importlib_resources


eddie-scio:
Could we add version boundaries [here](https://github.com/Julian/jsonschema/blob/v4.2.0/setup.cfg#L33) like we do for other libraries?  Older versions (e.g. 1.0.2) of `importlib_resources` (if pinned) will throw errors like below

```
  File "/env/lib/python3.7/site-packages/jsonschema/__init__.py", line 29, in <module> 
    from jsonschema.validators import ( 
  File "/env/lib/python3.7/site-packages/jsonschema/validators.py", line 349, in <module> 
    meta_schema=_utils.load_schema("draft3"), 
  File "/env/lib/python3.7/site-packages/jsonschema/_utils.py", line 60, in load_schema 
    path = resources.files(__package__).joinpath(f"schemas/{name}.json") 
AttributeError: module 'importlib_resources' has no attribute 'files' 
```

Julian:
Sure, happy to see a PR pinning importlib-resources from below.

PR #877, opened 2021-11-05
Add version constraint to importlib_resources

sevein:
Based on this recent [change in the docs](https://github.com/python/importlib_resources/commit/ed3b2d30e01121b27a6e0d0a8a478a483e437736) I opted to require `importlib_resources>=1.4.0`, but tests also passed in jsonschema when using importlib_resources v1.1.0.

Let me know if you want to use `importlib_resources>=1.1.0` instead.

For reference: https://importlib-resources.readthedocs.io/en/latest/history.html.

Fixes https://github.com/Julian/jsonschema/issues/876.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ python_requires = >=3.7
 install_requires =
     attrs>=17.4.0
     importlib_metadata;python_version<'3.8'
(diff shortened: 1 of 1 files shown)


Julian:
Thanks.

Issue #818, opened 2021-06-15
Issue in multipleOf validator


vcatpro:
360.57 is not a multiple of 0.0001
74.77 is not a multiple of 0.0001

Julian:
Hello. Please see previous issues: #810, #687, #185, #320, #247.

vcatpro:
Thanks for the quick reply and a great library.

Julian:
Thank you for the kind words!

Issue #810, opened 2021-05-21
Float operations can lead to a false fail of `multipleOf`


lorinkoz:
Hi, much ❤️ to the library.

I found this validation problem when using `multipleOf`:

```
>>> from jsonschema import validate
>>> validate(instance=10.1, schema={"type":"number", "multipleOf":0.1})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/path/to/python/site-packages/jsonschema/validators.py", line 934, in validate
    raise error
jsonschema.exceptions.ValidationError: 10.1 is not a multiple of 0.1

Failed validating 'multipleOf' in schema:
    {'multipleOf': 0.1, 'type': 'number'}

On instance:
    10.1
```

When printing the cotient that results from this code:
https://github.com/Julian/jsonschema/blob/0ea12edea319580473bda7e26ef088ff3a0983c5/jsonschema/_validators.py#L169

The cotient is 100.99999999999999

Therefore, `int(quotient)` can't be `quotient` on https://github.com/Julian/jsonschema/blob/0ea12edea319580473bda7e26ef088ff3a0983c5/jsonschema/_validators.py#L171

What to do?

Julian:
Hi there, thanks!

If you want fixed point precision (or generally non-float behavior), you should use `decimal.Decimal` when deserializing your JSON.

lorinkoz:
Okay, thanks. I guess that's the solution, then. Keep up the good work!

tdamsma:
As this is the first issue I stumbled upon, I'm adding these links for reference: #687, #185, #320, #247

Issue #687, opened 2020-05-19
MultipleOf validation Bug


noelwilsondel:
Hello,

Firstly thanks for this library it's really useful, we have noticed some unusual behaviour on version 3.2.0 validation using `multipleOf` constraint for example:

```
>>> from jsonschema import validate
>>> schema = {
...     "type" : "object",
...     "properties" : {
...         "test" : {"type" : "number", "multipleOf": 0.001},
...     },
... }
>>> # This should be valid but raises validation error
>>> validate(instance={"test" : 0.95}, schema=schema)
```

Some other things we noticed using the above value this passes:
```
>>> # This is valid
>>> validate(instance={"test" : 0.01}, schema=schema)
>>> # This is valid
>>> validate(instance={"test" : 0.1}, schema=schema)
>>> # This is invalid
>>> validate(instance={"test" : 0.95}, schema=schema)
```

Julian:
Hi! Glad to hear.

Sorry you're having trouble -- this issue is just the way float division works -- have a look at some [historical related issues](https://github.com/Julian/jsonschema/search?q=multipleof&type=Issues).

If you don't want float division, use something with the precision you want, e.g. `decimal.Decimal`, when deserializing your JSON.

Julian:
Hope that helps with some ideas, obviously happy to follow up.

jwnwilson:
@Julian We're attempting the `parse_float=decimal.Decimal` route thanks for the suggestion

Issue #320, opened 2017-02-07
multipleOf incorrectly raises ValidationError for some floats


dalehumby:
In ``_validators.py``, the ``multipleOf`` function incorrectly raises a ValidationError for e.g. 606.67 as not a multiple of 0.01, due to floating point errors.

Our use case is using a schema to ensure that amounts (in Dollars) are in increments of 1 cent ($0.01)

Going through the [code](https://github.com/Julian/jsonschema/blob/master/jsonschema/_validators.py#L119), 
```
quotient = 606.67 / 0.01 = 60666.99999999999
failed = int(quotient) != quotient
failed = 60666 != 60666.99999999999
failed = True
```

One number is a multiple of another number when the quotient is an integer. However, in this case, the quotient is _very close_ to an integer due to floating point representation of those numbers.

The solution I see is: ``instance`` is a multiple of ``db`` if ``quotient`` is close to an integer.

I'm happy to submit a pull request, and first would like to 

1. write a unit tests for this case. Where is the best place for this test?
2. implement an isclose method similar to [math.isclose](https://docs.python.org/3/library/math.html#math.isclose) in Python 3. A [previous commit](https://github.com/Julian/jsonschema/commit/2b9a72273edeac33da7d7f9054f676b0e9f37750) removed something similar. Is there a better way of doing this?

Julian:
I'd recommend not using floats to recommend dollar amounts in that case. You could use a string and a simple regex fairly easily I think?

Julian:
Alternatively you can deserialize your JSON using decimals as well.

Julian:
Closing this out, but lemme know if you disagree.

Fufutako:
If we use Decimal in deserialization we get an error:

```
jsonschema/_validators.py", line 124, in multipleOf
    quotient = instance / dB
TypeError: unsupported operand type(s) for /: 'decimal.Decimal' and 'float```

so multipleof is unusable...

This is how I did to workaround:
```
def multipleOf(validator, dB, instance, schema):
    if not validator.is_type(instance, "number"):
        return

    if isinstance(dB, float):
        if isinstance(instance, float):
            quotient = instance / dB
        else:
            quotient = instance / Decimal(str(dB))
        failed = int(quotient) != quotient
    else:
        failed = instance % dB

    if failed:
        yield ValidationError("%r is not a multiple of %r" % (instance, dB))```

[...]

PR #878, opened 2021-11-05
More lenient float multipleOf validation

tdamsma:
Many user have ran into floating point precision errors when validation multipleOf. E.g. #818, #810, #687, #185, #320, #247. Technically it can be argued that it is silly to check if any number is an integer multiple of e.g. 0.1. Ask silly questions, get silly answers. This approach is a bit unhelpful though.

Looking at other implementations in javascript and python, other libraries struggle with [this](https://github.com/ajv-validator/ajv/issues/652) [too](https://github.com/ajv-validator/ajv/issues/1005)

Checking if 10.1 is a multiple of 0.1 yields mixed results:

```
>>> import fastjsonschema
>>> fastjsonschema.validate({"multipleOf": 0.1}, 10.1)
10.1

>>> import jsonschema_rs
>>> jsonschema_rs.JSONSchema.from_str('{"multipleOf": 0.1}').is_valid(10.1)
False
>>> jsonschema_rs.JSONSchema({"multipleOf": 0.1}).is_valid(10.1)

```
https://jsonschema.dev/ fails validation
https://www.jsonschemavalidator.net/ accepts it

And [react-jsonschema-form](https://rjsf-team.github.io/react-jsonschema-form/#eyJmb3JtRGF0YSI6MTAuMSwic2NoZW1hIjp7InR5cGUiOiJudW1iZXIiLCJtdWx0aXBsZU9mIjowLjF9LCJ1aVNjaGVtYSI6eyJmaXJzdE5hbWUiOnsidWk6YXV0b2ZvY3VzIjp0cnVlLCJ1aTplbXB0eVZhbHVlIjoiIiwidWk6YXV0b2NvbXBsZXRlIjoiZmFtaWx5LW5hbWUifSwibGFzdE5hbWUiOnsidWk6ZW1wdHlWYWx1ZSI6IiIsInVpOmF1dG9jb21wbGV0ZSI6ImdpdmVuLW5hbWUifSwiYWdlIjp7InVpOndpZGdldCI6InVwZG93biIsInVpOnRpdGxlIjoiQWdlIG9mIHBlcnNvbiIsInVpOmRlc2NyaXB0aW9uIjoiKGVhcnRoaWFuIHllYXIpIn0sImJpbyI6eyJ1aTp3aWRnZXQiOiJ0ZXh0YXJlYSJ9LCJwYXNzd29yZCI6eyJ1aTp3aWRnZXQiOiJwYXNzd29yZCIsInVpOmhlbHAiOiJIaW50OiBNYWtlIGl0IHN0cm9uZyEifSwiZGF0ZSI6eyJ1aTp3aWRnZXQiOiJhbHQtZGF0ZXRpbWUifSwidGVsZXBob25lIjp7InVpOm9wdGlvbnMiOnsiaW5wdXRUeXBlIjoidGVsIn19fSwidGhlbWUiOiJkZWZhdWx0IiwibGl2ZVNldHRpbmdzIjp7InZhbGlkYXRlIjp0cnVlLCJkaXNhYmxlIjpmYWxzZSwicmVhZG9ubHkiOmZhbHNlLCJvbWl0RXh0cmFEYXRhIjpmYWxzZSwibGl2ZU9taXQiOmZhbHNlfX0=) accepts it.

So there is no real consensus amongst the libraries. Still I find the number of issues a strong indication that the current behaviour of this library is unexpected. This PR's modifies the multipleOf behaviour to allow for float tolerance (epsilon) to be taken into account.

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -186,6 +186,16 @@ def multipleOf(validator, dB, instance, schema):
         quotient = instance / dB
         try:
             failed = int(quotient) != quotient
+            if failed and dB != int(dB):
+                # Checking if floats are integer multiples of non integer
+                # floats is asking for floating point errors. Use a more
+                # lenient validation that probably conforms to the users
+                # expectations where 101 * 0.1 == 10.1 would be true.
+                # This also conforms to behaviour in javascript jsonschema
+                # checkers
+                remainder = instance % dB
+                tolerance = float_info.epsilon * instance
+                failed = remainder > tolerance and dB - remainder > tolerance
         except OverflowError:
             # When `instance` is large and `dB` is less than one,
             # quotient can overflow to infinity; and then casting to int


Julian:
Hi. Thanks for this but I'm going to politely decline.

As I mentioned in the other issues, if someone wants non-float behavior, there is already a way to get it via Decimals.

Otherwise, this change just moves which floats are incorrect multiples from some to others.

If you indeed are interested in this, I'd point you to [this PR](https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/413) which claims there is a better algorithm entirely for us to use.

I don't really think that looking at the number of tickets filed is a good metric for surprisability (and I'm not sure I believe surprise is the right metric at all in this case, this library implements a spec, so for better or worse it's compliance with the spec that trumps).

But just looking at issues filed is going to be biased against the non-current behavior -- people who expect this behavior aren't filing tickets saying things worked as expected :)

But do appreciate the PR nonetheless.

tdamsma:
I understand and sympathize with you points, but am going to try to convince you otherwise nonetheless.

I see now that you have been involved in [discussions](https://github.com/json-schema-org/json-schema-spec/issues/312) on the jsonschema spec itself, without reaching a meaningful course of action. I just see many valid points raised, linked issues, calls for clarification of the spec, other languages sturggling with this etc. 

My take is that it is up to the language to interpret/implement this as they want (correct me if I'm wrong here). The spec offers no real guidance here (and whilest I think that should be improved, that is another discussion), for the python implementation I think python's best practices should be guiding. 

And that is where it becomes a bit unclear I guess. On first sight, python works like this:

```python
>>> 0.1*round(10.1/0.1)
10.100000000000001
>>> 0.1*(10.1/0.1)
10.1
```
And it seems fair to reason that 10.1 is indeed not an integer multiple of 0.1.

But digging a bit deeper, both python itself and the native json library already do some implicit rounding:

```python
>>> import json
>>> v = json.loads("0.100000000000000001")
>>> print(v)
0.1
>>> print(f"{v:.20g}")
0.10000000000000000555
>>> print(json.dumps(v))
0.1
```

There are three occurences of implicit rounding here: 
 - the input is rounded to the nearest float.
 - then printing the float, the value is by default represented as `0.1`. This behaviour was explcitily introduced in python https://docs.python.org/3/tutorial/floatingpoint.html
 - when converting it back into json, the number is also represented as 0.1

So the underlying philosophy seems (to a certain degree) to be to treat float(0.1) as exactly 1/10. 

So even though the behaviour below is strict/correct/explainable:
```
 >>> 1/10 == 0.1
True
>>> 10.1 / 101 == 0.1
False
```

This is also the case:
```
jsonschema.validate(instance=json.loads("10.00000000000000005"), schema={"multipleOf": 1})
```

If I really cared, I should have used a `Decimal` representation:
```python
>>> import simplejson 
>>> import jsonschema
>>> jsonschema.validate(instance=simplejson.loads("10.00000000000000005",use_decimal=True), schema={"multipleOf": 1})

Failed validating 'multipleOf' in schema:
    {'multipleOf': 1}

On instance:
    Decimal('10.00000000000000005')
```

So whilest you have argued that is you want more correct behaviour of multipleOf one should use Decimal and not float, I woul like to argue that I think it is more pythonic to silently add some leniency to multipleOf validation (equal to the float representation error) just like python displays `float(0.1)` differently from `Decimal('0.1')`. If you really care about exactness here use `Decimal`, otherwise do some rounding and get on with it. 

> this change just moves which floats are incorrect multiples from some to others.

Not really, it just checks (or should check) if the number is equal to the nearest representable float. 

Now there are many different ways to implement this, and perhaps one based on Decimal would feel less arbitrary:

```
instance = 10.1
dB = 0.1
instance_d = Decimal(repr(instance))
dB_d = Decimal(repr(dB))
quotient = instance_d / dB_d
failed = int(quotient) != quotient
```

If the users intention actually is to use arbitrary precision decimal representaion they can do so by using the Decimal package explicitly, something they should be doing anyways as this holds:

```python
>>> Decimal(1) / Decimal(0.1)
Decimal('9.999999999999999444888487687')
>>> 1 / 0.1
10.0
```

Which in the current implementation actually leads to the arbitrary behaviour where some numbers are multiples of some irrepresentable numbers and others are not.

Julian:
Apologies for not having more time to respond in detail, but there's no implicit rounding in your examples -- what Python does in recent versions is a very targeted, very specific display change -- https://bugs.python.org/issue1580

It reprs floats by using the shortest input that produces the given float. But no rounding is happening, 0.1 isn't a representable float, so you always have gotten the same float, before or after the repr change. And absolutely no tolerance is added -- it's strictly a display change.

There's as far as I know no precedence anywhere in Python for introducing some implicit tolerance when doing float operations unless the user explicitly asks for it, and what `jsonschema` does matches that -- it does float operations and gives the same answers Python does if you do them yourself:

```
>>> v = 10.00000000000000005 / 1
>>> print(f"{v:.50g}")
10
```

Issue #853, opened 2021-10-05
performance regression from 3.2.0 to 4.0.1


tfranzel:
Hi @Julian, first of all thanks for this library and all the hard work. We rely heavily on it and are delighted to see `Draft 2020-12` support as it is required for OpenAPI 3.1.0

When updating, I noticed our test suite ran noticeably slower. I compiled a quick reproduction for you. On my machine I get a 5x slowdown with the new version. Is this performance hit to be expected due to the new capabilities or is it a regression?

```python
import jsonschema
import yaml
import json
import time

# https://raw.githubusercontent.com/tfranzel/drf-spectacular/master/tests/test_basic.yml
with open('tests/test_basic.yml') as fh:
    data = yaml.load(fh.read(), Loader=yaml.SafeLoader)

# https://raw.githubusercontent.com/tfranzel/drf-spectacular/master/drf_spectacular/validation/openapi3_schema.json
# which comes from:
# https://github.com/OAI/OpenAPI-Specification/blob/6d17b631fff35186c495b9e7d340222e19d60a71/schemas/v3.0/schema.json
with open('drf_spectacular/validation/openapi3_schema.json') as fh:
    openapi3_schema_spec = json.load(fh)

t_acc = 0

for i in range(500):
    t0 = time.time()
    jsonschema.validate(instance=data, schema=openapi3_schema_spec)
    t1 = time.time()
    t_acc += t1 - t0

print(f'{t_acc} sec')
```

``` bash
✗ python --version; pip freeze | grep json; python test.py
Python 3.9.5
jsonschema==3.2.0
5.254251718521118 sec

✗ python --version; pip freeze | grep json; python test.py
Python 3.9.5
jsonschema==4.0.1
28.189855813980103 sec

✗ python --version; pip freeze | grep json; python test.py
Python 3.9.7
jsonschema==4.2.1
27.27832531929016 sec

✗ python --version; pip freeze | grep json; python test.py
Python 3.9.7
jsonschema==4.3.1
8.10183048248291 sec
```

EDIT: included measurement for `4.2.1` release
EDIT: included measurement for `4.3.1` release

tdamsma:
I noticed this too, a set of tests I run for an application that uses jsonschema heavily the test suite now takes 60 minutes vs 14 before.

Julian:
It will take some digging, but there is indeed a "known" performance chance in 4.0.0: https://github.com/Julian/jsonschema/blob/main/CHANGELOG.rst#v400

[...]

PR #886, opened 2021-11-30
Extend dynamicRef keyword

nezhar:
The description in https://stackoverflow.com/a/69774520/5382179 helped me better understand how the keyword should actually work and the PR extends the missing parts of the current implementation of the keyword.

There might be a better way to implement this, but it would require bigger changes in the way the scopes and the validator functions are handled.

### DIFF
--- a/jsonschema/tests/test_jsonschema_test_suite.py
+++ b/jsonschema/tests/test_jsonschema_test_suite.py
@@ -405,15 +405,7 @@ TestDraft202012 = DRAFT202012.to_unittest_testcase(
     skip=lambda test: (
         narrow_unicode_build(test)
         or skip(
-            message="dynamicRef support isn't working yet.",
-            subject="dynamicRef",
-        )(test)
-        or skip(
-            message="These tests depends on dynamicRef working.",
-            subject="defs",
-        )(test)
-        or skip(
-            message="These tests depends on dynamicRef working.",
+            message="These tests require an extension or the url resolver.",
             subject="anchor",
             case_description="same $anchor with different base uri",
         )(test)
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -1,3 +1,4 @@
+from collections import deque
 from collections.abc import Mapping, MutableMapping, Sequence
 from urllib.parse import urlsplit
 import itertools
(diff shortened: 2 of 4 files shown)


jsonschema/tests/test_jsonschema_test_suite.py
```diff
@@ -401,15 +401,7 @@ def leap_second(test):
...
-            message="dynamicRef support isn't working yet.",
-            subject="dynamicRef",
-        )(test)
-        or skip(
-            message="These tests depends on dynamicRef working.",
-            subject="defs",
-        )(test)
-        or skip(
-            message="These tests depends on dynamicRef working.",
+            message="These tests require an extension or the url resolver.",
```
Julian, line 404: Thanks! This looks promising. Can you elaborate on what you mean by an extension of the URL resolver? Looking at this test it doesn't look dynamicRef related (so I must have mis-labelled the message), but I'm not sure I see why this wouldn't work as-is with the location independent ref support you helped add.

nezhar: The `resolve_fragment` method located in `RefResolver` is resolving the anchor by taking the first subschema that matches the anchor name in the current schema. 

```python
for keyword in ["$anchor", "$dynamicAnchor"]:
    for subschema in self._finditem(document, keyword):
        if fragment == subschema[keyword]:
            return subschema
```

This work fine as long as the anchor names are not repeated in other scopes, this is why the test fails and resolves to:

```json
{
    "$id": "child2",
    "$anchor": "my_anchor",
    "type": "number"
}
```

We may need to extend `_finditem` to to not leave the scope while searching for the anchor or take all findings and evaluate if they not break out of the current scope before returning the result.

Julian:
@nezhar can you have a look at resolving the conflicts here now that the performance issues with the implementation have been addressed?

Julian:
Thanks @nezhar. I'm about ready to merge this (sorry for the delay), but to be extra safe I'd like to re-run the benchmarks from @Stranger6667 in #853 -- can you rerun the two they left at the bottom there just to be sure we don't regress on them by merging this?

Stranger6667:
FWIW, some of the work done during ref resolving is only relevant for newer drafts (after Draft 7), so maybe as a follow-up from this PR it would be nice to explore adding some conditional logic to avoid unnecessary work (e.g. the `$dynamicAnchor` keyword is not a part of Draft 7, so could be skipped during resolving) for older drafts. Not sure how much benefit it would yield though

nezhar:
@Julian also sorry for the delay :slightly_smiling_face:. I made another rebase and run the test from the linked issue. This are my results:

```
$ python --version; pip freeze | grep json; python test.py
Python 3.8.10
jsonschema==4.3.3
10.12309193611145 sec

$ python --version; pip freeze | grep json; python test.py
Python 3.8.10
jsonschema @ file:///home/hnezbeda/DEV/jsonschema
10.206285238265991 sec
````

Julian:
No worries! Will merge I think. Thanks! If you manage to figure out the remaining 2 skipped tests in draft2020 a PR would of course would be very much appreciated. So long as they still fail, my suspicion is that the implementation is still not 100% correct (even in the parts that pass), but it's definitely an improvement. Thanks again!

PR #892, opened 2021-12-10
Setup mypy in `tox -e typing` and get it to pass

sirosen:
This is the smallest possible change to get mypy passing on the jsonschema codebase. The goal of this configuration is to enforce type annotations anywhere that they appear.
That is, if a method is added to the codebase,

    def foo(x: int) -> str:
        return str(x)

then usages of `foo` will by type checked. If no annotations are added, `mypy` will not type check functions.

For the most part, this keeps the impact low. The one exceptional case is the use of `pyrsistent.pmap` as an argument to `attr.ib(converter=...)`. Unfortunately, it causes `mypy` to incorrectly deduce the type of the init parameter created by attrs. We need to "explain the type of init" to mypy by creating a callable with a concrete type to act as the converter. The callable in question simply wraps `pmap` with a cast and presents the desired type
information to mypy.

---

A few asides:

I see some whitespace edits crept in when I `black`ed files. I'll back those out when I can work on this more next week.

I personally like `import typing as t` and then using `t.Optional`, `t.Dict`, etc. throughout a codebase. But there are various opinions on this, so I've gone with `import typing` and `typing.Optional`, etc. as what I believe is the least controversial choice.

`tox -e typing` seemed consistent with `tox -e style`: name the env by the thing being checked, not the tool used to check it. `tox -e mypy` would also work.

Type checking enforcement is somewhat necessary if we are to consider publishing type information from `jsonschema`. I don't know if this is really desirable, but the PR shows what it would take to get type checking to pass.

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -1,18 +1,19 @@
 """
 Validation errors, and some surrounding helpers.
 """
+from __future__ import annotations
+
 from collections import defaultdict, deque
 from pprint import pformat
 from textwrap import dedent, indent
 import itertools
-import typing
 
 import attr
 
 from jsonschema import _utils
 
-WEAK_MATCHES: typing.FrozenSet[str] = frozenset(["anyOf", "oneOf"])
-STRONG_MATCHES: typing.FrozenSet[str] = frozenset()
+WEAK_MATCHES: frozenset[str] = frozenset(["anyOf", "oneOf"])
+STRONG_MATCHES: frozenset[str] = frozenset()
 
 _unset = _utils.Unset()
 
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -5,13 +5,14 @@ from collections import defaultdict, deque
 from pprint import pformat
 from textwrap import dedent, indent
 import itertools
+import typing
 
 import attr
 
 from jsonschema import _utils
 
-WEAK_MATCHES = frozenset(["anyOf", "oneOf"])
-STRONG_MATCHES = frozenset()
+WEAK_MATCHES: typing.FrozenSet[str] = frozenset(["anyOf", "oneOf"])
+STRONG_MATCHES: typing.FrozenSet[str] = frozenset()
 
 _unset = _utils.Unset()
 
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -1663,7 +1663,7 @@ class AntiDraft6LeakMixin(object):
 
 class TestDraft3Validator(AntiDraft6LeakMixin, ValidatorTestMixin, TestCase):
     Validator = validators.Draft3Validator
-    valid: typing.Tuple[dict, dict] = {}, {}
+    valid: typing.Tuple[dict, dict] = ({}, {})
     invalid = {"type": "integer"}, "foo"
 
     def test_any_type_is_valid_for_type_any(self):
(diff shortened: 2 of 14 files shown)


jsonschema/exceptions.py
```diff
@@ -5,13 +5,14 @@
...
 import itertools
+import typing
 
 import attr
 
 from jsonschema import _utils
 
-WEAK_MATCHES = frozenset(["anyOf", "oneOf"])
-STRONG_MATCHES = frozenset()
+WEAK_MATCHES: typing.FrozenSet[str] = frozenset(["anyOf", "oneOf"])
```
Julian, line 14: (Ignorance question, I'll have to read myself a bit more before merging) but is there a more generic type here than this? I'd like to not say this is a frozenset, in the future it may (likely will) become a `pyrsistent.pset`. So just something like `collections.abc.Set` or whatever the typing equivalent is would be a bit preferable.

jsonschema/tests/test_validators.py
```diff
@@ -1662,7 +1663,7 @@ def test_False_is_not_a_schema_even_if_you_forget_to_check(self):
 
 class TestDraft3Validator(AntiDraft6LeakMixin, ValidatorTestMixin, TestCase):
     Validator = validators.Draft3Validator
-    valid = {}, {}
+    valid: typing.Tuple[dict, dict] = ({}, {})
```
Julian, line 1666: I take it mypy complains if you don't add these types here? (If not I don't see much value in adding them here for typing the test case fixtures, the real type is in fact the type of schemas, i.e. `Union[dict, bool]` right now, but possibly some larger union type later, but if it's simpler to add than ignore, then yeah leave them).

jsonschema/exceptions.py
sirosen: I think the best we can do is `typing.Collection[str]`, which is the type for `collections.abc.Collection`.

`typing.Set` doesn't work, as it is the type for `builtins.set`, not `collections.abc.Set`.
Most of `collections.abc` is mirrored in `typing`, and even `collections.abc.MutableSet` maps to `typing.MutableSet`. I guess `Set` was skipped because of the likelihood of confusion with the builtin set type.

In py3.9+, all of the `collections.abc` types are usable as generics, so we'd be able to write
```python
WEAK_MATCHES: collections.abc.Set[str] = ...
```
and I would expect it to work.

The big alternative, which I could apply here and elsewhere, is to use `from __future__ import annotations`, now that py3.6 has been dropped. The future import enables several features and behaviors, but the one that is relevant here is that it allows the use of builtin types in annotations. So we can do
```python
from __future__ import annotations

WEAK_MATCHES: frozenset[str] = ...
```

If you prefer the results of the future import, I can apply it to all of the places where I had to add annotations.

jsonschema/tests/test_validators.py
sirosen: Yes, unfortunately. `mypy` generally complains when it sees unannotated empty dict assignments.
Adding this annotation seemed the simplest way to pacify it.

We could also add `# typing: ignore` comments. For tests I think it would be appropriate (I am against the use of ignore comments in most cases, since it's just as verbose as an annotation and leaves the type checker with less information).
If we don't want type checking on tests at all, we can add `exclude = test_*` to the mypy config, and it will exclude all `test_*` module names. I could do that and revert any changes to test modules, if that is preferable.

jsonschema/exceptions.py
Julian: Yeah let's go with the future import! After that I think I'll merge. Thanks again for the help here!

sirosen: I've just gotten around to doing this. Wherever it simplified things, I added a `from __future__ import annotations`.

Let me know if there are any issues with it, but I think we're good to go on this! 😄

sirosen:
I'm going to have to make further tweaks. I forgot about how the future import behaves.
`x = Union[int, str]` is valid in all (current) language versions but `x = int | str` is only valid in py3.10+ even with the future import.

Sorry for the noise; hopefully I'll have it sorted out shortly.

sirosen:
There were a few issues with the doc build which I think I have sorted out. I did some amends and rebases to try to keep a useful commit history.

631fba1 deals with issues arising from sphinx trying to build a cross-reference for `ClassVar[TypeChecker]`. Because it can't resolve `TypeChecker` to `jsonschema.TypeChecker`, it needs some special hand-holding. The commit message explains in more detail. If there's a better way to give sphinx context, I'm happy to adjust. 

17384e7 handles a related but different case (which was not flagged in my local builds because I was running tox under py3.8, and for whatever reason it only shows up on 3.9+). If you write `FooType = Union[int, str]; x: FooType`, that's valid and means what it looks like. However, when sphinx sees `FooType`, it will try to cross-reference that, rather than using the value. The easiest solution is to just avoid these kind of type assignments and use the desired type directly, even if it's a little more verbose.

cdc2807 does the same kind of cleanup as 17384e7, but in a location where sphinx was not seeing the variable names. It's purely precautionary.

811bab2 omits type-checking-only code from coverage reporting, to try to get the coverage check to pass. It doesn't seem to have been sufficient, but I think it still makes sense to include.

As I write this, the build is still going and the coverage check is still red, but I expect all of it to pass.

Julian:
Awesome! Don't worry too much about the coverage check, I'll have a look. Appreciated.

Julian:
Odd (and worrying) that things pass locally in Sphinx even the way you originally had it, and on the same versions of everything via tox. But I don't have time to diagnose why at the minute... Going to merge. Thanks again for the PR!

sirosen:
> Odd (and worrying) that things pass locally in Sphinx even the way you originally had it

I agree.

I'm generally a big fan of the new typing stuff, but this branch seemed to run into almost all of my least-favorite things -- typing-time code in `if TYPE_CHECKING`, idiosyncratic rules about how to do conditional imports, "future annotations" which the runtime can't handle, ...

Please feel free to call me in to help out if any of this creates problems down the line. And thanks for being receptive to these changes!

Julian:
I very well may take you up on that offer, thanks!

PR #894, opened 2021-12-15
perf: Cache reference lookups for subschemas

Stranger6667:
The latest commit cuts the time to 8.08942461013794 sec which is quite close to `3.2.0` version performance (5.38702654838562 sec on my machine)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ v4.3.0
   certain input types (#893)
 * Cache reference lookups for subschemas (#893)
 * Use cached lookups for resolving fragments if the referent document is known (#893)
+* Replace the ``Validator.evolve`` method with an equivalent class attribute
 * Implement a PEP544 Protocol for validator classes (#890)
 
 v4.2.1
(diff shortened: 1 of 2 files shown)


Stranger6667:
Another small thing that cuts the time to `7.816486835479736 sec` & removes ~2M func calls from that example code is replacing `Validator.evolve` method with `evolve = attr.evolve` so there is no extra function call overhead. Let me know if this change would be ok :)

Julian:
Ayy CPython... Yes that's likely fine with me too as long as we preserve the documentation!

CHANGELOG.rst
```diff
@@ -5,6 +5,7 @@ v4.3.0
   certain input types (#893)
 * Cache reference lookups for subschemas (#893)
 * Use cached lookups for resolving fragments if the referent document is known (#893)
+* Replace the ``Validator.evolve`` method with an equivalent class attribute
```
Julian, line 8: This is an implementation detail, we can save mentioning it I think.

And for the other two I'd maybe just say something simpler like "resolving refs has had performance improvements" and have that cover the whole change, someone can look at the PR for details.

Stranger6667: Done! :)

Julian:
Fantastic, thanks again, really appreciate the time! Merging.

PR #896, opened 2021-12-18
perf: Cache subschemas

Stranger6667:
Continuation of #894. This patch introduces a cache for subschemas-related keywords, which reduces the number of calls to the search routine. All subschemas-related keywords are searched in one pass and then stored in a cached dict. With the OP from #893 it reduces the total number of function calls from 61512655 to 56438634 (~9%). The number of calls to the search routine is reduced 5x (from 1090000 to 218000).

I'd be happy to know what do you think :)

### DIFF
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -755,27 +755,21 @@ class RefResolver(object):
         finally:
             self.pop_scope()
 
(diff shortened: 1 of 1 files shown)


Julian:
Wonderful, thanks again, merging!

PR #899, opened 2022-01-01
Move warnings to user context

Kojoley:
Currently the warning location points inside the `jsonschema` code and it takes an additional effort to find where the actual code that triggers it is located.

### DIFF
--- a/jsonschema/__init__.py
+++ b/jsonschema/__init__.py
@@ -46,6 +46,7 @@ def __getattr__(name):
             "removed in a future release. Use importlib.metadata directly "
             "to query for jsonschema's version.",
             DeprecationWarning,
+            stacklevel=2,
         )
 
         try:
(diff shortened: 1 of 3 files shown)


Julian:
Thanks, this is definitely reasonable --  can you update the tests for these deprecations?

Appreciated.

Kojoley:
> Thanks, this is definitely reasonable -- can you update the tests for these deprecations?
> 
> Appreciated.

I could not figure out what should I update, the test suit seems to be fine as-is, may you please explain what do you mean?

Julian:
Each deprecation here has a test asserting that some code triggers the deprecation warning.
That test should also be modified so not only does it assert the deprecation was raised, but also that the stacklevel was correct (which it wasn't before this change).
I'm not at a computer but I can give an example later if needed, just let me know if the above is unclear. (But essentially every change here should have a test ensuring the change was broken before and fixed now)

Kojoley:
Should be done now.

Julian:
Wonderful, thanks!

PR #909, opened 2022-01-27
Cache all jsonschemas in vocabularies.json to make jsonschema work offline

franekbcorp:
There were two jsonschemas missing from the vocabularies.json file: "https://json-schema.org/draft/2019-09/links" and "https://json-schema.org/draft/2019-09/hyper-schema". The PR adds the relevant jsonschemas, which are not downloaded from the internet any more. That allows the library to work offline and may provide performance benefits.


Julian:
Thanks, I'm still very unhappy with having to do things this way, the main reason is `importlib.resources` not supporting directories :/

I'd definitely like to merge this, but can you try to add a test ensuring this is the case which fails before this change and passes after? i.e. a test which perhaps makes a ref resolver that is unable to hit the internet, and then validates draftN schemas (and therefore should likely fail before and pass now)?

Julian:
Please feel free to follow up if you are interested in this still, otherwise will close (and possibly look at this again after vocabulary support is merged).

PR #923, opened 2022-03-07
Handle case when Content-Encoding is gzip

h-hellstrom:
Currently, there is a "magic" behaviour when whether or not `requests` is installed, a dependency which currently is quite hidden.

This PR tries to reduce the magic behaviour, by handling the `gzip` case (ignoring the cases of `compress`, `deflate` and `br`) and raising an error being transparent that the package cannot decode the current encoding and recommending installing request.

### DIFF
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -957,7 +957,7 @@ class RefResolver(object):
             # json over http
             result = requests.get(uri).json()
         else:
-            # Otherwise, pass off to urllib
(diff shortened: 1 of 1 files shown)


Julian:
Thanks. I agree that the current behavior isn't as discoverable as it should be, but I'm more inclined to make requests or now httpx a required dependency and drop the urllib code. Happy to see a PR which does that instead! And appreciated either way.

PR #924, opened 2022-03-08
Implements vocabulary loader for draft 2019-09 and draft 2020-12

nezhar:
This will allow the use of dedicated vocabularies based on the meta schema definition. For this the validators are grouped by the the vacabulary schemas, and only loaded if they are provided in the vocabulary of the meta schema. The default meta schema is used if the "$schema" is not provided.

The implementation still allows to override and provide custom validators.


Julian:
Thanks @nezhar -- not sure how I feel about needing to be this explicit -- let me give playing with this PR locally a bit a shot. Will follow up in a bit. Appreciated!

Julian:
This is going to require some significant changes (and new API), so with apologies (especially for again having these sit here), I'm going to close this for now.

PR #925, opened 2022-03-08
Split contains, maxContains and minContains beetween applicator and validation vocabulary

nezhar:
This is an addition to https://github.com/Julian/jsonschema/pull/924/ in order to provide support vor contains, maxContains and minContains based on the defined vocabulary in the meta schema. Before the 3 keywords where handled in `contains`, but since they are part of diefrent vocabularies, they should be handled on their own.


Julian:
This is going to require some significant changes (and new API), so with apologies (especially for again having these sit here), I'm going to close this for now.

PR #926, opened 2022-03-11
Remove stray double-quote

lurch:
The help-string for --error-format is triple-quoted, so remove unnecessary quote character


Julian:
Thanks!

PR #940, opened 2022-05-05
Ensure proper sorting of list in error message

ssbarnea:
Fixes problem where error message was not consistent between runs, even when the inputs were the same (file to be validated and the schema to use).

This should make the outcomes predictable and eliminate the random factor from it. In fact the sorting already happens in 3 other places inside the library but that specific location was missed.

This was found while working on https://github.com/ansible/ansible-lint/pull/2035 and where we observed that the output did very randomly between two runs:
```
"Additional properties are not allowed ('when', 'pki_certificates', 'pki_realms', 'pki_authorities', 'pki_private_groups_present', 'pki_routes' were unexpected)"
"Additional properties are not allowed ('pki_private_groups_present', 'when', 'pki_authorities', 'pki_certificates', 'pki_realms', 'pki_routes' were unexpected)"
```

Not listing the allowed properties alphabetically is also not human-friendly as it make quite hard to spot specific entries, especially as the size of these list can grow considerably.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -113,7 +113,7 @@ def extras_msg(extras):
         verb = "was"
     else:
         verb = "were"
-    return ", ".join(repr(extra) for extra in extras), verb
+    return ", ".join(repr(extra) for extra in sorted(extras)), verb
 
 
 def ensure_list(thing):
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -594,7 +594,7 @@ class TestValidationErrorMessages(TestCase):
         message = self.message_for(instance=["foo", "bar"], schema=schema)
         self.assertIn(
             message,
-            "Unevaluated items are not allowed ('foo', 'bar' were unexpected)",
+            "Unevaluated items are not allowed ('bar', 'foo' were unexpected)",
         )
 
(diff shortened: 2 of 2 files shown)


ssbarnea:
> +1, but a similar change has been rejected before. See PR #774

Ouch... I hope it was due to some misunderstanding because the ability to provide consistent error message instead of random ones should not even need to be explained.

cidrblock:
If we can do something here I think it would be a real win.  I won't pretend to know the solution, but our use of json schemas continues to increase and we love the library.

I can think of quite a few places where the errors are being presented right back to the user, so if consistent ordering is possible IMHO, it would be great.

Julian:
The 3.11 failure in CI is spurious (and I probably should just disable testing on 3.11), but the other one is not -- can you address that one?

ssbarnea:
@Julian thanks! I also updated the tests. You might want to configure github to require GHA approval only for new-github users, that proved to be more enough for us to prevent CI abuse. That involves editing https://github.com/python-jsonschema/jsonschema/settings/actions and selecting "Require approval for first-time contributors who are new to GitHub".

Julian:
Thanks (I made that GH setting change in the background just as you said it hah.) Will merge in a few. Appreciated.

PR #944, opened 2022-05-08
Fix exception when encountering numeric properties

ssbarnea:
While JSON does not allow integers to be used as map keys, YAML does allow it and trying to validate such document caused an ugly exception.

Fixes: https://github.com/python-jsonschema/check-jsonschema/issues/89

This fix might not be correct but at least it addressing the exception and treating the value as a string. We might want to change the loader to ensure it does that but I was not sure where to do such a change.

Obviously that it will also need some tests so we prevent future regressions but I a was not sure how to add them. Feel free to add them if you know or just give me a hint.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -98,9 +98,9 @@ def find_additional_properties(instance, schema):
     patterns = "|".join(schema.get("patternProperties", {}))
     for property in instance:
         if property not in properties:
-            if patterns and re.search(patterns, property):
+            if patterns and re.search(patterns, str(property)):
                 continue
-            yield property
+            yield str(property)
 
 
 def extras_msg(extras):
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -21,7 +21,7 @@ def patternProperties(validator, patternProperties, instance, schema):
 
     for pattern, subschema in patternProperties.items():
         for k, v in instance.items():
-            if re.search(pattern, k):
+            if re.search(str(pattern), str(k)):
                 yield from validator.descend(
                     v, subschema, path=k, schema_path=pattern,
                 )


Julian:
Hey, thanks -- I've written about this elsewhere in a previous issue that I'll have to find.

I'm somewhat inclined to agree with https://github.com/python-jsonschema/check-jsonschema/issues/89#issuecomment-1120586251 at least in part, but for here specifically:

It's not obvious to me (as I probably said in that previous issue I mentioned) that what's happening here is overwhelmingly more correct than another choice -- specifically, given we're anyways talking about something beyond the scope of JSON Schema, it's not obvious that treating non-str keys as strings is overwhelmingly more behavior than ignoring non-str keys entirely and not applying validation to them.

Which to me makes me want to stick with the current behavior of erroring, even if we do so via a more transparent exception. Open to opinions though.

sirosen:
As a direct user of `jsonschema`, I find it unsurprising that a dict which cannot be serialized to JSON cannot be checked without errors.
I would +1 whatever decision puts the least maintenance strain on this library.

There's a cost to maintaining any feature, even a small one. But if this is something which people ask about repeatedly, there's also cost to explaining the situation.

ssbarnea:
I am not sure what is expected from me on this PR. I still think that validating incoming data to be JSON-compatible would be desired as default.

Maybe we just need to raise a validation error like InputWasNotJSONCompatible if garbage is fed into when jsonify argument is False. If jsonify argument would be True or False by default, i am not sure. I would personally be ok with any of them.

Please let me know what you think and what would be needed.

Julian:
> InputWasNotJSONCompatible

Is this meaningfully different than the current behavior? The current behavior is "if you give something that isn't JSON compatible, you get an exception other than `ValidationError`, so if you're worried about that, `except Exception` and handle that as you wish."

To me that's preferable, which is why it's the current behavior. Wrapping exceptions in another exception doesn't really do much other than hide context (even if it's still in the nested traceback printout), so I'd love to hear what benefits you're hoping for there.

(I don't know what you mean by `jsonify` -- are you proposing such an argument be added somewhere?)

ssbarnea:
`TypeError` is very generic python error and it would be unsafe to assume that this one is always caused by invalid data being fed into the validator.

Catching generic Exception would be even [worse](https://realpython.com/the-most-diabolical-python-antipattern/), as it would likely hide other genuine runtime errors.

Do you have any performance concerns regarding validating input data? I doubt that these would be significant, likely less than 1% of current performance (just a guess).

Julian:
(it would be quite literally how such a thing would even be implemented within the library)

sirosen:
It is not the responsibility of `jsonschema` to make sure that its input is well-formed. That's the caller's responsibility.
There was a bug in `check-jsonschema` (now fixed), but there is no bug in `jsonschema`.

What this PR adds is a nicety for a small class of users, but it isn't being well-justified by this thread. Why should the library incur _any_ performance penalty to handle inputs which are known to be bad?

If you really don't care at all about performance, just `json.loads(json.dumps(my_dict))`.

Julian:
Given that there's multiple possible implementations here, that neither of which is obviously better than the other, both are beyond the specification, that there was a downstream bug here that's been fixed, and this issue is handleable with the current code if you are dealing with data that's potentially of the wrong type, I think it's best to close -- happy to discuss further, but that seems like it's still the right option at the minute.

Issue #959, opened 2022-05-30
Module 'attr' has no attribute 's'


anuran-roy:
There is a recurring error with the aforementioned syntax wherever jsonschema and attrs are a dependency:

## Screenshot:
![image](https://user-images.githubusercontent.com/76481787/171035934-76ec0c30-7dba-4d60-9a17-e158f7867c73.png)

On looking up, I found that jsonschema uses `import attr`,  while the PyPI index of the `attr` package says to use the `import dry_attr`:

![image](https://user-images.githubusercontent.com/76481787/171036682-962c687d-a4cc-4bd7-990f-1481a39b3109.png)

Julian:
The relevant package is called attrs, not attr -- what version do you have installed and how did you install jsonschema, your package manager should have installed attrs automatically.

Julian:
This is an installation issue as I mentioned. May be happy to help if you provide the details above.

PR #960, opened 2022-05-30
Change attr to dry attr

anuran-roy:
Solves issue #959


Julian:
This is incorrect. See my comment on the issue.

Issue #963, opened 2022-06-07
Unable to use recent versions of jsonschema from typed code (mypy)


ssbarnea:
As the wrapper `types-jsonschema` went out of sync, is no longer possible to use it with mypy. Attempts to use the library directly gives errors because `py.typed` marker was removed.

```
src/ansible_compat/schema.py:6: error: Skipping analyzing "jsonschema": module is installed, but missing library stubs or py.typed marker
src/ansible_compat/schema.py:6: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
src/ansible_compat/schema.py:7: error: Skipping analyzing "jsonschema.validators": module is installed, but missing library stubs or py.typed marker
```

Looking at https://pypi.org/project/types-jsonschema/#history I see that these releases are 2 versions behind as the current version is `4.6.0`

Can we please start publishing types with the library?

Related:
* https://github.com/python-jsonschema/jsonschema/pull/954
* https://github.com/sirosen/jsonschema/commit/ded065d324064a28019739f7478d29b021b687dd which reverted the previous change, apparently not via a pull request.

Julian:
https://github.com/python-jsonschema/jsonschema/pull/954#issuecomment-1136342244 was fairly clear about what the next steps are here. Again, help is welcome.

PR #964, opened 2022-06-07
Adopt typing information from typeshed (PEP-561)

ssbarnea:
Avoids problems with mypy by officially declaring the library as typed while copying the current type information from types-jsonschema.

From now on the library itself is typed and we will be able to ensure that this types are correct and always in sync with each version of the library.

Closes: #963


Julian:
@ssbarnea to save you some time I'm going to close this I think.

A PR of 400+ lines touching 13 files doesn't meet any definition of "small" that [I mentioned in the comment](https://github.com/python-jsonschema/jsonschema/pull/954#issuecomment-1148559814).

Separately it seems @sirosen [isn't in agreement](https://github.com/python-jsonschema/jsonschema/pull/954#issuecomment-1148588634) neither with the proposal to copy things in this PR nor that the original report from #963, so you need to discuss that.

More generally, I think you need to pay closer attention to reading comments carefully and making sure there's buy-in before spinning wheels. #963 itself was a good sign of not paying attention to what was previously discussed.

PR #983, opened 2022-08-17
jsonschema deserves a ✨fancy✨ readme

hynek:
So this is a simple starting point.

LMK what you think. I for one would probably cut "Running the Test Suite", "Benchmarks", and "Community" from PyPI.

You can test the output by checking this branch out and running `pipx run hatch-fancy-pypi-readme`.

---

Small aside: One thing I've notices is that your heading chars are a bit off. Your Readme uses over-and-under `=` but the sections use `-`. Then your `index.rst` uses `-` for Contents, but `=` for Indices and tables.

I think you should switch Contents and everything that is `-` in the README to just-under `=` and am happy to make it part of this PR if you want to.


pyproject.toml
```diff
@@ -77,6 +76,26 @@ Tidelift = "https://tidelift.com/subscription/pkg/pypi-jsonschema?utm_source=pyp
...
+[[tool.hatch.metadata.hooks.fancy-pypi-readme.fragments]]
+path = "README.rst"
+
+[[tool.hatch.metadata.hooks.fancy-pypi-readme.fragments]]
+text = """
+
+
+===================
+Release Information
+===================
```
Julian, line 91: I think for the same reason as the comment this should be just under-`-` unless we do change the headings in the `README`, otherwise it'll turn into an H1 (and be the only other H1), no?

hynek:
The h3 stuff is just relative, but this is the priority:

```rst
==
h1
==

h2
==

h3
--

h4
~~

```

Since you don't have `=` in your README, your `-` become h2 when rendering it. But `-` clashes with your changelog.

I'm a bit surprised by both double-= and single-= becoming H1 in index.rst, I guess it's trying to be smart there?
 
---

Which is why I use in my docs double-`=` in my own docs. It makes it possible to have sections with single-= that are sometimes h1 and sometimes h2.

pyproject.toml
hynek: yes, but your changelog consists of `-`-headings which means they'd become equal to your sections.

Making it `=` would \~probably\~ make it h2 and all your sections h3.

hynek:
Hilariously, here's the resulting readme in a gist: https://gist.github.com/hynek/378cc636308aa9fdd9eaaea550d31e12

for some reason it made double-`=` h2s but I suspect that is because it has its own h1 and tries to avoid having two.

Julian:
> I'm a bit surprised by both double-= and single-= becoming H1 in index.rst, I guess it's trying to be smart there?

Those are different, double-`=` becomes H1, and single-`=` becomes H3, see: https://python-jsonschema.readthedocs.io/en/stable/#jsonschema vs. https://python-jsonschema.readthedocs.io/en/stable/#indices-and-tables

The priority is based on what heading styles it sees first

pyproject.toml
Julian: I think if we want to make believe this problem doesn't exist :D we can just change all the headings in the Changelog to be single `=` and then change this to be single `-` (which will make the heading H2, and then each changelog entry be an H3)

hynek:
ugh yeah you're right…it's about order which is insane.

just to re-iterate: the problem is that the changelog uses the same section marker as the readme. pulling it up to `=` would solve it, because "Release Info" could become single-`=`.

There's almost nothing in the index.rst: we'd just have to switch the marker of Contents and Indices, or am I missing something again?

Julian:
There's a bunch of ways to do it, I *think* the easiest one is to leave `index.rst` exactly as is (so its sections stay H2 and H3 respectively) and then make `Release Information` single-`-` which would be H2, and then change all the headings in the Changelog to `=` which would make them H3 (under the release information).

ssbarnea:
TBH, I by fancy it means potential (ab)use of emoji, like in PR title, I would better prefer it without it. I hate seeing emoji inside commit messages, makes the git log looks my Spam folder from gmail.

Emoji are fine, for personal messages but not for changelogs. They distract user from the real content.

hynek:
> There's a bunch of ways to do it, I _think_ the easiest one is to leave `index.rst` exactly as is (so its sections stay H2 and H3 respectively) and then make `Release Information` single-`-` which would be H2, and then change all the headings in the Changelog to `=` which would make them H3 (under the release information).

Done!

I've updated https://gist.github.com/hynek/378cc636308aa9fdd9eaaea550d31e12 for preview too.

Julian:
Hooray! Thanks :D, very fancy.

PR #987, opened 2022-08-18
Use rST markers in README

hynek:
Let's not talk about it.

https://gist.github.com/hynek/378cc636308aa9fdd9eaaea550d31e12 updated


Julian:
Hah, oh, right. Good job me for not catching that already. Thanks.

PR #999, opened 2022-09-09
Make it clearer that `format: regex` validates against Python regexes

Julian:
The same applies to pattern, patternProperties, etc.

CC @hynek

----
:books: Documentation preview :books:: https://python-jsonschema--999.org.readthedocs.build/en/999/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -105,8 +105,8 @@ They can be used when installing in order to include additional dependencies, e.
 
     $ pip install jsonschema'[format]'
 
-Be aware that the mere presence of these dependencies – or even the specification of ``format`` checks in a schema – do **not** activate format checks (as per specification).
-Please read the `format validation documentation <https://python-jsonschema.readthedocs.io/en/latest/validate/#validating-formats>`_ for details.
+Be aware that the mere presence of these dependencies – or even the specification of ``format`` checks in a schema – do *not* activate format checks (as per the specification).
+Please read the `format validation documentation <https://python-jsonschema.readthedocs.io/en/latest/validate/#validating-formats>`_ for further details.
 
 .. start cut from PyPI
 


docs/validate.rst
```diff
@@ -187,70 +187,31 @@ By default, no validation is enforced, but optionally, validation can be enabled
...
-which will install all of the below dependencies for all formats.
+        $ pip install jsonschema[format-nongpl]
 
 Or if you want to install MIT-license compatible dependencies only:
 
-.. code-block:: sh
+    .. code-block:: sh
 
-   $ pip install jsonschema[format_nongpl]
+        $ pip install jsonschema[format_nongpl]
```
hynek, line 205:
```suggestion
        $ pip install jsonschema[format-nongpl]
```

(same indent problem)

docs/validate.rst
```diff
@@ -187,70 +187,31 @@ By default, no validation is enforced, but optionally, validation can be enabled
...
-(if any,) are listed below.
+The more specific list of formats along with any additional dependencies they have is shown below.
 
-.. note::
+.. warning::
 
-    If the following packages are not installed when using a checker
-    that requires it, validation will succeed without throwing an error,
-    as specified by the JSON Schema specification.
+    If a dependency is not installed when using a checker that requires it, validation will succeed without throwing an error, as also specified by the specification.
```
hynek, line 214: This is so terrible. Can't you add an unofficial strict mode that fails when

1. an format is added without a format checker
2. and unknown format is specified
3. an format is specified that you lack the deps for?

docs/validate.rst
```diff
@@ -173,7 +173,7 @@ Validating Formats
 ------------------
 
 JSON Schema defines the :kw:`format` keyword which can be used to check if primitive types (``string``\s, ``number``\s, ``boolean``\s) conform to well-defined formats.
-By default, no validation is enforced, but optionally, validation can be enabled by hooking in a format-checking object into a `Validator`.
+By default, as per the specification, no validation is enforced. Optionally however, validation can be enabled by hooking a `format-checking object <jsonschema.FormatChecker>` into a `Validator`.
```
hynek, line 176:
```suggestion
By default, as per the specification, no validation is enforced.
Optionally however, validation can be enabled by hooking a `format-checking object <jsonschema.FormatChecker>` into a `Validator`.
```

docs/validate.rst
```diff
@@ -187,70 +187,31 @@ By default, no validation is enforced, but optionally, validation can be enabled
...
-available if an appropriate package is available for use. The easiest way to
-ensure you have what is needed is to install ``jsonschema`` using the
-``format`` or ``format_nongpl`` collection of optional dependencies -- e.g.
-
-.. code-block:: sh
-
-   $ pip install jsonschema[format]
-
-which will install all of the below dependencies for all formats.
+        $ pip install jsonschema[format-nongpl]
```
hynek, line 199: Given the narrative before, I think this should be:

```suggestion
        $ pip install jsonschema[format]
```

Also the indentation is off I think. If I interpret it correctly, Sphinx is rendering it as code in a blockquote.

README.rst
```diff
@@ -105,6 +105,8 @@ They can be used when installing in order to include additional dependencies, e.
 
     $ pip install jsonschema'[format]'
 
+You will also want to read the `format validation documentation <https://python-jsonschema.readthedocs.io/en/latest/validate/#validating-formats>`_.
```
hynek, line 108: This is too ominous IMHO. Also I think a relative link with a ref is better? I just accidentally clicked myself into the official docs and was surprised why it's old.

You'd add something like `.. _validating-formats:` _right before the heading_ and then you can use `:ref:BACKTICKvalidating-formatsBACKTICK` or something similar.

```suggestion
Be aware that the mere presence of these dependencies – or even the specification of ``format`` checks in a schema – do **not** activate format checks (as per specification).
Please read the `format validation documentation <https://python-jsonschema.readthedocs.io/en/latest/validate/#validating-formats>`_ for details.
```

hynek (changes requested):

README.rst
Julian: This is the README, so the relative links don't work (they'll work in sphinx from RTD but not when clicking in the README itself, which doesn't descend into the doc folder). This is why I want to kill the readme entirely personally (and just link to the docs), rather than extend it :/

(Other change looks good though thanks).

Julian:
> Additionally to what I mentioned in the comments, I really think you should change the example that is on the frontpage to something that does what people expect. 

I'll consider this. Like I said, I would kind of like to actually neuter the README more rather than extend it :( but can't do so until the quickstart is up. But will consider.

PR #1003, opened 2022-09-14
don't install tests and benchmarks

jcgruenhage:
`jsonschema` currently installed the whole `jsonschema` subfolder of this
repository, which includes tests and benchmarks. From my testing, those don't
need to be installed, so here's a patch to not install them.

----
:books: Documentation preview :books:: https://python-jsonschema--1003.org.readthedocs.build/en/1003/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,12 @@ build-backend = "hatchling.build"
 [tool.hatch.version]
 source = "vcs"
 
+[tool.hatch.build]
+exclude = [
+  "jsonschema/tests",
+  "jsonschema/benchmarks",
+]
+
 [project]
 name = "jsonschema"
 description = "An implementation of JSON Schema validation for Python"


Julian:
Hi, thanks! An argument could possibly be made for the benchmarks, so probably happy to exclude those, but generally tests should be installed with (any) package, as users may be interested in running the tests against an installed version.

jcgruenhage:
I don't really agree on that. Running tests of a library that's installed as a dependency somewhere down the stack isn't something that's commonly done by users. If a user wants to do so, they can still choose to include those files, but defaulting to having them not included does make more sense IMO. Installing the tests is not a common practice, and also not the default for the tooling that you're using. `hatch` defaults to having the tests in a separate folder in the repository root, which causes the tests not to be shipped to users by default.

jcgruenhage:
The test coverage delta is still something that I hadn't forseen and is something that means that this can't be merged here as is.

Julian:
> Installing the tests is not a common practice

Sure it is. See e.g. http://pythonchb.github.io/PythonTopics/where_to_put_tests.html which is just the first Google result, but there have essentially been two camps here for ~10 years at least, and plenty of packages do both.

> and also not the default for the tooling that you're using. 

Hatch is a packaging tool, it doesn't really have an opinion on this, though it's author of course has to make a choice one way or the other on the topic itself of where to put tests in hatch itself as a package.

Of course if you're packaging for a distro you're welcome to prune the tests if you feel you'd like to, since I agree if some end user gets jsonschema installed they're unlikely to care about it or its tests, but for normal uses (developers who've installed jsonschema while developing some Python package) this is both common and a good practice in my opinion -- we're talking tiny files-- so I'm going to thank you for the PR but decline.

PR #1006, opened 2022-10-06
Resolve references

chbrandt:
This is the old PR #419 from @kave merged into current `main` as suggested in thread/comment in https://github.com/python-jsonschema/jsonschema/issues/692#issuecomment-938988243 .

Pytest is running fine. Tox is complaining about the docs somewhere.
I am pushing the PR anyways to not let it pass (code-wise).

----
:books: Documentation preview :books:: https://python-jsonschema--1006.org.readthedocs.build/en/1006/


chbrandt:
Yes, I see. Well, maybe we could put that (i.e, dynamicRef) as a "known issue/limitation"? 

Programmatically, the functionality could be enabled for when Draft07 validator was being use, or (ii) as a function ("util") outside Resolver.

Julian:
Happy to take another look at this if/when it handles dynamic references. Appreciated!

PR #1019, opened 2022-11-26
Add some typing to the ``exceptions.py`` module

DanielNoord:
After getting the go-ahead in https://github.com/python-jsonschema/jsonschema/issues/1017 I thought I would test the waters both in size and scope of potential PRs.

Feel free to ask me to separate some of the commits into separate PRs. I thought this was quite manageable, especially if you go commit by commit.

----------------------

I thought I would start with a fairly self contained module (_and a module I interact with quite a lot with at work myself_).
I think most of the changes are pretty straight forward. The first commit is hardest to review and for now I focused on the attributes that I was most sure about, keeping harder ones for later.

The second and third commit "fix issues". I understand that these are not actual issues and that through the dynamicness of Python these code paths actually currently work. So: what is the desired action here? Add ignores to tell `mypy` we know this works? Or add some `isinstance` checks and attribute definitions to make the code actually type safe. I lean towards the latter but don't want to impose my personal preference on a project I am not involved with until now. I'm more than happy to change those commits to do something differenty!

Happy reviewing 😄 

Edit: Sorry for all the force pushes, after a helpful [comment](https://github.com/python/typeshed/issues/8662#issuecomment-1328039199) from @ AlexWaygood in the `typeshed` repo I found one issue with my initial PR.

----
:books: Documentation preview :books:: https://python-jsonschema--1019.org.readthedocs.build/en/1019/

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -6,13 +6,16 @@ from __future__ import annotations
 from collections import defaultdict, deque
 from pprint import pformat
 from textwrap import dedent, indent
-from typing import ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar
 import heapq
 import itertools
 
 import attr
 
 from jsonschema import _utils
+
+if TYPE_CHECKING:
+    from jsonschema import _types
 
 WEAK_MATCHES: frozenset[str] = frozenset(["anyOf", "oneOf"])
 STRONG_MATCHES: frozenset[str] = frozenset()
@@ -66,10 +69,10 @@ class _Error(Exception):
         for error in context:
             error.parent = self
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"<{self.__class__.__name__}: {self.message!r}>"
 
-    def __str__(self):
+    def __str__(self) -> str:
         essential_for_verbose = (
             self.validator, self.validator_value, self.instance, self.schema,
         )
@@ -99,11 +102,11 @@ class _Error(Exception):
         )
 
     @classmethod
-    def create_from(cls, other):
+    def create_from(cls, other: _Error):
         return cls(**other._contents())
 
     @property
-    def absolute_path(self):
+    def absolute_path(self) -> deque[str | int]:
         parent = self.parent
         if parent is None:
             return self.relative_path
@@ -113,7 +116,7 @@ class _Error(Exception):
         return path
 
     @property
-    def absolute_schema_path(self):
+    def absolute_schema_path(self) -> deque[str | int]:
         parent = self.parent
         if parent is None:
             return self.relative_schema_path
@@ -123,7 +126,7 @@ class _Error(Exception):
         return path
 
     @property
-    def json_path(self):
+    def json_path(self) -> str:
         path = "$"
         for elem in self.absolute_path:
             if isinstance(elem, int):
@@ -132,7 +135,11 @@ class _Error(Exception):
                 path += "." + elem
         return path
 
-    def _set(self, type_checker=None, **kwargs):
+    def _set(
+        self,
+        type_checker: _types.TypeChecker | None = None,
+        **kwargs: Any,
+    ) -> None:
         if type_checker is not None and self._type_checker is _unset:
             self._type_checker = type_checker
 
@@ -188,7 +195,7 @@ class RefResolutionError(Exception):
 
     _cause = attr.ib()
 
-    def __str__(self):
+    def __str__(self) -> str:
         return str(self._cause)
 
 
@@ -197,10 +204,10 @@ class UndefinedTypeCheck(Exception):
     A type checker was asked to check a type it did not have registered.
     """
 
-    def __init__(self, type):
+    def __init__(self, type: str) -> None:
         self.type = type
 
-    def __str__(self):
+    def __str__(self) -> str:
         return f"Type {self.type!r} is unknown to this type checker"
 
 
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -4,6 +4,7 @@ Validation errors, and some surrounding helpers.
 from __future__ import annotations
 
 from collections import defaultdict, deque
+from collections.abc import Iterable, Mapping
 from pprint import pformat
 from textwrap import dedent, indent
 from typing import TYPE_CHECKING, Any, ClassVar
@@ -31,17 +32,17 @@ class _Error(Exception):
     def __init__(
         self,
         message: str,
-        validator=_unset,
-        path=(),
-        cause=None,
+        validator: str | _utils.Unset = _unset,
+        path: Iterable[str | int] = (),
+        cause: Exception | None = None,
         context=(),
         validator_value=_unset,
-        instance=_unset,
-        schema=_unset,
-        schema_path=(),
-        parent=None,
-        type_checker=_unset,
-    ):
+        instance: Any = _unset,
+        schema: Mapping[str, Any] | bool | _utils.Unset = _unset,
+        schema_path: Iterable[str | int] = (),
+        parent: _Error | None = None,
+        type_checker: _types.TypeChecker | _utils.Unset = _unset,
+    ) -> None:
         super(_Error, self).__init__(
             message,
             validator,
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -154,10 +154,14 @@ class _Error(Exception):
         )
         return dict((attr, getattr(self, attr)) for attr in attrs)
 
-    def _matches_type(self):
+    def _matches_type(self) -> bool:
         try:
-            expected = self.schema["type"]
+            # We ignore this as we want to simply crash if this happens
+            expected = self.schema["type"]  # type: ignore[index]
         except (KeyError, TypeError):
+            return False
+
+        if isinstance(self._type_checker, _utils.Unset):
             return False
 
         if isinstance(expected, str):


DanielNoord:
I have spent way too much time getting the documentation to build 😅 

`autoapi` and `autodoc` really don't work well with signature typing. They can't follow imports to ignored files so anything from `_utils` is unfound. This could be solved by also including those files, but then we get warnings about `_RaisesType` not being found:
https://github.com/python-jsonschema/jsonschema/blob/46fdb982a26aac211bc4d20b2ef6d8089a520f77/jsonschema/_format.py#L15

It seems silly that this should be so difficult but whatever I tried I couldn't get it to find all references successfully...

jsonschema/exceptions.py
```diff
@@ -4,35 +4,44 @@
...
+
 WEAK_MATCHES: frozenset[str] = frozenset(["anyOf", "oneOf"])
 STRONG_MATCHES: frozenset[str] = frozenset()
 
 _unset = _utils.Unset()
 
 
 class _Error(Exception):
+
+    _word_for_schema_in_error_message: str | None = None
```
Julian, line 29: This shouldn't be `None`, just `str`, though I guess I see you're adding it to the (never-instantiated `_Error`). But yeah can we just leave it at `str`?

jsonschema/exceptions.py
```diff
@@ -142,12 +159,18 @@ def _contents(self):
         )
         return dict((attr, getattr(self, attr)) for attr in attrs)
 
-    def _matches_type(self):
+    def _matches_type(self) -> bool:
+        if isinstance(self.schema, _utils.Unset):
```
Julian, line 163: I'd prefer adding `type: ignore`s rather than any runtime things here yeah.

It's a programmer error to get here if `schema` is somehow unset.

jsonschema/exceptions.py
```diff
@@ -61,14 +70,18 @@ def __init__(
...
 
-    def __str__(self):
+    def __str__(self) -> str:
         essential_for_verbose = (
             self.validator, self.validator_value, self.instance, self.schema,
         )
-        if any(m is _unset for m in essential_for_verbose):
+        if (
+            any(m is _unset for m in essential_for_verbose)
+            or self._word_for_schema_in_error_message is None
```
Julian, line 82: (Same here, I'd rather not add this, `_Error` is not actually meant to be instantiated, it's just an interface, though a messy one clearly with concrete behavior in it. Some day maybe it should instead be a function.)

jsonschema/exceptions.py
```diff
@@ -4,35 +4,44 @@
...
+    _word_for_instance_in_error_message: str | None = None
+
     def __init__(
         self,
-        message,
-        validator=_unset,
-        path=(),
-        cause=None,
+        message: str,
+        validator: str | _utils.Unset = _unset,
```
Julian, line 35: Apparently I don't have ReadTheDocs preview builds enabled (ugh), but I'm worried all these will show up in documentation, when they shouldn't -- the user facing type of `validator` is `str`. It's an internal implementation detail that it can be `unset` for the time validation is still ongoing, but I don't want users to see that, and it's not a part of its public API. Will build the docs to see what shows up (and probably enable preview builds).

jsonschema/exceptions.py
```diff
@@ -4,35 +4,44 @@
...
+        path: Iterable[str | int] = (),
+        cause: Exception | None = None,
         context=(),
         validator_value=_unset,
-        instance=_unset,
-        schema=_unset,
-        schema_path=(),
-        parent=None,
-        type_checker=_unset,
+        instance: dict[str, Any] | _utils.Unset = _unset,
```
Julian, line 40: `instance` is type `Any` (any Python object can be validated).

jsonschema/exceptions.py
```diff
@@ -4,35 +4,44 @@
...
+        schema: dict[str, Any] | _utils.Unset = _unset,
```
Julian, line 41: `schema` is at the minute type `Mapping[str, Any] | bool`.

docs/spelling-wordlist.txt
```diff
@@ -46,10 +48,37 @@ validator
...
 
 Zac
 HD
 
 Berman
 Libera
 GPL
+
+# json properties
+additionalProperties
```
Julian, line 61: I assume this is somehow autodoc pulling things in again? Basically none of these should really be here, because none of them are really words, they should always either be surrounded in backticks (which will disable spell-checking them) or be used in `:kw:`foo`` roles. But autodoc can be funny there and find these words in documents we don't control...

DanielNoord: I can try one more time to fix this 😄 

I have an idea that might work, but I can't promise much. It is indeed `autodoc` and `autoapi` pulling more stuff in.

jsonschema/exceptions.py
DanielNoord: I used `None` as for most `_Errors` this is indeed `None`.

Ideally we would create a private subclass of `_Error` that has this defined as `str` and which can then be subclassed by the actual errors that have this attribute. But since I didn't want to introduce new subclasses without approval I thought I would propose this first. Would new subclasses for the classes that use this attribute be okay?

jsonschema/exceptions.py
DanielNoord: With this PR the typing no longer shows up as adding the typing annotations completely breaks the documentation builds.

I understand that it should be `str` but there is really no way to satisfy a type checker when you are assigning `_unset` to something that is annotated with `str`. Is it really that important that `Unset` doesn't leak? I found it quite easy to reason about when I first saw it and understood that `validator` should be `str` unless I'm breaking the package 😄

jsonschema/exceptions.py
DanielNoord: See my comment above!

jsonschema/exceptions.py
Julian: > With this PR the typing no longer shows up as adding the typing annotations completely breaks the documentation builds.

Hm. As in totally disables it displaying in documentation at all?

I definitely want it showing up there, it's really the only reason I care about typing at all, though I definitely feel the tooling pain with these tools, but yeah that's really the primary motivation for typing info, so I'd be worried if that's totally disabled now?

DanielNoord: The issue is that `autoapi` doesn't understand typing aliases... So even if I get all other references to resolve correctly it will still complain about:
https://github.com/python-jsonschema/jsonschema/blob/46fdb982a26aac211bc4d20b2ef6d8089a520f77/jsonschema/_format.py#L15

This is used in a signature and based on the documentation that `autoapi` builds any reference to `_RaisesType` will cause a warning. That's why in the end I disabled the typing from showing up in the docs.

Julian: https://github.com/readthedocs/sphinx-autoapi/issues/224 it looks like is maybe the/a relevant issue -- but yeah to be upfront without trying to waste any of your time I definitely would consider not showing typing info in the docs to be a non-starter unfortunately, as I say, it's the main thing I care about :/ so it may require some thinking if indeed stuff doesn't work there before merging additional changes.

DanielNoord: That's indeed the issue. Would just using `autodoc` instead of `autoapi` be okay? You'd need to manually rewrite the index whenever you write a new module, but I think `autodoc` should be able to find all relevant classes needed to create the documentation.

Julian: I'm open to the idea (it's how I used to do things before autoapi). If `sphinx-apidoc` produces the right output with minimal modifications I'm even more open to it. But yeah :) if the diff is reasonable then yes.

jsonschema/exceptions.py
DanielNoord: Done!

jsonschema/exceptions.py
Julian: Hm. Can you elaborate on what you mean by most `_Error`s having this be none?

I could be misremembering my own library, but every "encountered" instance of `_Error` should be of a concrete subclass (i.e. `ValidationError` or `SchemaError`), both of which define these class attributes appropriately, so no end-user should ever see an `_Error` where it's anything but `str`. Am I missing what you mean?

DanielNoord: Well the `__str__` implementation that uses these attributes is on `_Error`, even though the attribute will only ever be defined on two of its subclasses. This would always throw up any typechecker as `_Error.__str__` uses attributes that can not always be accessed.
This can be avoided by either adding the attribute to `_Error` as being `None` unless they are set in the two subclasses that do, or have an inbetween subclass that uses the attributes and remove the attributes from `_Error.__str__`

DanielNoord: I can make an example PR for the sub classes example. However, as long as the documentation doesn't build I don't think it makes a lot of sense to pursue these changes anyway. Some help with that PR would be appreciated (but I don't want to rush you, everybody has other priorities than OSS).

DanielNoord:
Rebased on `main` and ready for review again!

jsonschema/exceptions.py
```diff
@@ -28,17 +32,17 @@ class _Error(Exception):
     def __init__(
         self,
         message: str,
-        validator=_unset,
-        path=(),
-        cause=None,
+        validator: str | _utils.Unset = _unset,
```
Julian, line 35: I think these `_unsets` still need removing / `# type: ignore'ing` as per the above, I still don't want them appearing in docs (I don't want end-users knowing `_unset` exists at all even).

DanielNoord: Sorry for never getting back on this, life became busy.

I don't really know how to resolve this as I think it would start to become really hard to add typing to the package itself if we were to add `type: ignore` everywhere. If it is not part of the public API then it will be hard to "hide" it from users if it is also being used in parts that are exposed to the public API (such as the exception classes).

I'd be very happy to help add typing and refactor some code so that internal details don't get exposed, but adding incorrect typing and then adding `type: ignore` seems to more harm than good.

Julian: Thanks for the update and no worries on the delay, sorry for my own.

To be direct, kind of like https://github.com/python-jsonschema/jsonschema/pull/1027#discussion_r1042505168 I'm not sure I'm ready to merge this as-is in that case -- I have to kind of stew on it a bit. I've hopefully said a few times before, I think for documentation and editor support that I'm quite happy to merge or add typing related fixes, but beyond the two I really honestly think it's all overall more annoyance than it helps still, so while I'm slightly sympathetic, I will always still go with the former (docs) over anything else -- I want to get some `$ref` related functional fixes out, which will likely be another few weeks, but then I can perhaps think a bit more about whether it's possible to get an improvement that I'm happier with. Just being transparent to not waste any of your time.

DanielNoord:
Is there any interest in this PR? Or shall I just close it?

Julian:
Apologies for letting this falling off the map. I'll have a look today. I'll likely merge what I'm comfortable with (regarding what we were discussing about not leaking `Unset`), even if I'm not sure if it gets you all of the things you were hoping for. I'm happy to keep discussing if there's still a large gap afterwards, though I really think we need a type checker plugin which allows differentiating between public and private type signatures...

But let me have another look and really appreciate your effort.

Julian:
(Done)

Please feel encouraged to give more feedback (or certainly propose further work); I hope to take less than 2 years for any such future discussion :(

But yeah really appreciated, and hopefully even though it's not exactly as proposed, I hope what's merged is at least slightly better for you than previous.

dangotbanned:
@DanielNoord @Julian 
I'm not sure where to open an issue for this, as the [cause](https://github.com/vega/altair/pull/3475#issuecomment-2225376524) for https://github.com/vega/altair/pull/3475 seems to be `typeshed` conflicting with the aims set out in this PR - to not leak `Unset | _unset`
- https://github.com/python-jsonschema/jsonschema/pull/1019#issuecomment-2148958004
- https://github.com/python-jsonschema/jsonschema/commit/0024b584c6b6f332f0f5fa5452488901b73328f9
- https://github.com/python-jsonschema/jsonschema/pull/1019#discussion_r1055672054

Following `v4.23.0` the new stubs now expose `Unset` https://github.com/python/typeshed/pull/12301/files

I'm very aware `jsonschema` did not author the stubs, but I imagine this change was not anticipated - as a result of https://github.com/python-jsonschema/jsonschema/releases/tag/v4.23.0

Julian:
I actually honestly don't know either what the right specific thing to do is if typeshed did something different.

From an end-user perspective as I tried to say in this PR, it's just plain incorrect -- the public type of that attribute is str, will always be str, and it's an implementation detail that in the course of assembling the object we still do this temporary `_Unset` nonsense which we'd never do if it were authored today. But no one downstream using the attribute should be seeing anything but str, if they did it would be a bug in this library.

It sounds perhaps at least like we would return to the status quo if a PR to typeshed removed the change to that line and returned it to how they had it before?

dangotbanned:
> I actually honestly don't know either what the right specific thing to do is if typeshed did something different.
> 
> From an end-user perspective as I tried to say in this PR, it's just plain incorrect -- the public type of that attribute is str, will always be str, and it's an implementation detail that in the course of assembling the object we still do this temporary `_Unset` nonsense which we'd never do if it were authored today. But no one downstream using the attribute should be seeing anything but str, if they did it would be a bug in this library.
> 
> It sounds perhaps at least like we would return to the status quo if a PR to typeshed removed the change to that line and returned it to how they had it before?

Thanks for the quick response @Julian and confirming my suspicion

I agree, that would make the most sense in a PR to `typeshed`

DanielNoord:
As PR author I agree with patching `typeshed`. I think packages should strive towards typing their code "as is" and let dependencies deal with any "nonsense". This allows type checkers to actually verify the code of the package which in the long run probably provides more benefit to downstream packages than avoiding their nonsense.

However, while `jsonschema` is not fully typed (and therefore the type checker provides no benefit for downstream), `typeshed` should help users a little and only expose the types as they are at runtime.

PR #1021, opened 2022-11-28
Use only ``autodoc`` to generate the API documentation

DanielNoord:
As discussed in https://github.com/python-jsonschema/jsonschema/pull/1019 `autoapi` has some issues with typing annotations.

This creates almost parity with the current documentation. There are slight differences in lay-out and some additional classes get documented (they could be excluded by hand).
I tested this with the other PR and could build with `make html` without any validation errors. I'll rebase that PR on this one as a final test.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -1,0 +1,7 @@
+API Reference
+=============
+
+.. toctree::
+   :titlesonly:
+
+   /api/jsonschema/index
--- a/docs/api/jsonschema/exceptions/index.rst
+++ b/docs/api/jsonschema/exceptions/index.rst
@@ -1,0 +1,6 @@
+:py:mod:`jsonschema.exceptions`
+===============================
+
+.. automodule:: jsonschema.exceptions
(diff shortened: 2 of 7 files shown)


DanielNoord:
I really have no idea how to fix this anymore. `autodoc` just seems unable to pick up `TypeVars` or aliases from imported modules.  The documentation of a class from a private module in the index of the API reference (`FormatChecker` from `_format`) which uses a private alias from the module (`_F`) seems to completely throw off Sphinx and `autodoc`. I have searched online but wasn't able to find much help as to how this could be fixed...

As far as I can see now, I don't think any additional typing can be included into the documentation without completely restructuring it.

Julian:
I can try to have a look in a bit, thanks for the efforts so far.

Julian:
OK, the commit I just pushed I believe should move this forward.

DanielNoord:
@Julian This looks perfect! Well done figuring this out, I wasn't able to...

I'm fine with merging!

PR #1027, opened 2022-12-07
Harmonize annotations with python/typeshed#8608

sirosen:
I haven't had time to focus on `jsonschema` typing much recently, but I want to get things back on track. The first thing to do is to correct any drift between `jsonschema` and `typeshed`.
I'm going to try to produce a steady stream of very small PRs which get things back in sync.

This is therefore hopefully the first of a few changes.

---

python/typeshed#8608 introduced annotations for `create` which are not fully reflected here.

In order to reflect that state into `jsonschema`, a new module, `jsonschema._typing` is introduced. The purpose of `_typing` is to be a singular place for the library to define type aliases and any typing-related utilities which may be needed. This will let us use aliases like `_typing.JsonValue` in many locations where any valid JSON datatype is accepted.
The definitions can be refined over time as necessary.

Initial definitions in `_typing` are:
- Schema (any JSON object)
- JsonObject (any JSON object)
- JsonValue (any JSON value, including objects or sequences)

`Schema` is just another name for `JsonObject`. Perhaps it is not needed, but the name may help clarify things to a reader. It is not obvious at present whether or not it is a good or bad idea to notate it as such, but a similar Schema alias is defined in typeshed and seems to be working there to annotate things accurately.

These types are using `Mapping` and `Sequence` rather than `dict` or `list`. The rationale is that `jsonschema`'s logic does not dictate that the objects used *must* be defined in stdlib types or subtypes thereof. For example, a `collections.UserDict` could be used and should be accepted by the library (`UserDict` wraps but does not inherit from `dict`.)

----
:books: Documentation preview :books:: https://python-jsonschema--1027.org.readthedocs.build/en/1027/

### DIFF
--- a/jsonschema/_typing.py
+++ b/jsonschema/_typing.py
@@ -1,0 +1,14 @@
+"""
+Type aliases and utilities to help manage `typing` usage and type annotations.
+"""
+
+from collections.abc import Mapping, Sequence
+import typing
+
+Schema = Mapping[str, typing.Any]
+
+JsonObject = Mapping[str, typing.Any]
+
+JsonValue = typing.Union[
+    JsonObject, Sequence[typing.Any], str, int, float, bool, None
+]
--- a/jsonschema/validators.py
+++ b/jsonschema/validators.py
@@ -23,12 +23,18 @@ from jsonschema import (
     _format,
     _legacy_validators,
     _types,
+    _typing,
     _utils,
     _validators,
     exceptions,
 )
 
 _UNSET = _utils.Unset()
+
+_ValidatorCallback = typing.Callable[
+    [typing.Any, typing.Any, _typing.JsonValue, _typing.JsonObject],
+    typing.Iterator[exceptions.ValidationError],
+]
 
 _VALIDATORS: dict[str, typing.Any] = {}
 _META_SCHEMAS = _utils.URIDict()
@@ -114,13 +120,15 @@ def _store_schema_list():
 
 
 def create(
-    meta_schema,
-    validators=(),
+    meta_schema: _typing.Schema,
+    validators: Mapping[str, _ValidatorCallback] | tuple[()] = (),
     version=None,
-    type_checker=_types.draft202012_type_checker,
-    format_checker=_format.draft202012_format_checker,
-    id_of=_id_of,
-    applicable_validators=methodcaller("items"),
+    type_checker: _types.TypeChecker = _types.draft202012_type_checker,
+    format_checker: _format.FormatChecker = _format.draft202012_format_checker,
+    id_of: typing.Callable[[_typing.Schema], str] = _id_of,
+    applicable_validators: typing.Callable[
+        [_typing.Schema], typing.Iterable[tuple[str, _ValidatorCallback]]
+    ] = methodcaller("items"),
 ):
     """
     Create a new validator class.
@@ -186,7 +194,7 @@ def create(
     @attr.s
     class Validator:
 
-        VALIDATORS = dict(validators)
+        VALIDATORS: dict[str, _ValidatorCallback] = dict(validators)
         META_SCHEMA = dict(meta_schema)
         TYPE_CHECKER = type_checker
         FORMAT_CHECKER = format_checker_arg
@@ -339,7 +347,7 @@ def create(
     if version is not None:
         safe = version.title().replace(" ", "").replace("-", "")
         Validator.__name__ = Validator.__qualname__ = f"{safe}Validator"
-        Validator = validates(version)(Validator)
+        Validator = validates(version)(Validator)  # type: ignore[misc]
 
     return Validator
 


Julian:
Thanks! No worries on the time shortage, I hear you!

Very much appreciate the PR, this is a great plan overall. Will leave some minor comments.

(There's also #1022 from @DanielNoord outstanding to look at, which is driven by typing-related concerns, just in case you haven't seen it).

jsonschema/_typing.py
```diff
@@ -0,0 +1,14 @@
+"""
+Type aliases and utilities to help manage `typing` usage and type annotations.
```
Julian, line 2: (Definitely support adding this, thanks!)

jsonschema/_typing.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+
+from collections.abc import Mapping, Sequence
+import typing
+
+Schema = Mapping[str, typing.Any]
```
Julian, line 8: Schema is tricky to nail down for multiple reasons. Its precise type depends on the specific dialect of JSON Schema involved, and (speaking off the top of my head without double checking, which counts for lots of the comments I'm about to make --)
`applicable_validators` means that *technically* its type is literally `Any` (or some dependent type), because someone can decide they're inventing a JSON Schema dialect where the number `17` is a schema and means something, and `applicable_validators` will give the library what it needs to deal with that as a schema. I'm honestly not sure what to do about that.
* Ignoring the above, the "internal" type to all JSON Schema "official" drafts is -- for later drafts, `bool | Mapping[str, typing.Any]`, and for earlier ones `Mapping[str, typing.Any]`

jsonschema/_typing.py
```diff
@@ -0,0 +1,14 @@
...
+
+JsonObject = Mapping[str, typing.Any]
+
+JsonValue = typing.Union[
```
Julian, line 12: This is incomplete -- someone can come along and invent validator callbacks which only work if someone deserialized into `Decimal`, and as long as they ensure they've done so in their loading of JSON, their code works. In other words, I'm somewhat afraid this is really just `Any` in "real life".

jsonschema/validators.py
```diff
@@ -114,13 +120,15 @@ def _store_schema_list():
 
 
 def create(
-    meta_schema,
-    validators=(),
+    meta_schema: _typing.Schema,
```
Julian, line 123: I have to double check (which I will be lazy and not do here) but Unlike `_typing.Schema`, meta-schemas (in the library, but possibly also according to the spec, I forget) *must* be objects I think, not `bool | thing`, so you'll need two types I suspect?

jsonschema/validators.py
```diff
@@ -114,13 +120,15 @@ def _store_schema_list():
...
     version=None,
-    type_checker=_types.draft202012_type_checker,
-    format_checker=_format.draft202012_format_checker,
-    id_of=_id_of,
-    applicable_validators=methodcaller("items"),
+    type_checker: _types.TypeChecker = _types.draft202012_type_checker,
+    format_checker: _format.FormatChecker = _format.draft202012_format_checker,
+    id_of: typing.Callable[[_typing.Schema], str] = _id_of,
+    applicable_validators: typing.Callable[
+        [_typing.Schema], typing.Iterable[tuple[str, _ValidatorCallback]]
```
Julian, line 130: (This has to do with the above, the type for `applicable_validators` is really `Any -> tuple(...)`)

jsonschema/validators.py
```diff
@@ -114,13 +120,15 @@ def _store_schema_list():
 
 
 def create(
-    meta_schema,
-    validators=(),
+    meta_schema: _typing.Schema,
+    validators: Mapping[str, _ValidatorCallback] | tuple[()] = (),
```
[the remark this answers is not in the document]
sirosen: Yep, I'm cool with that way of handling these!

I had some other ideas which I like less, but another option just occured to me.

With `_typing` we could do this:
```python
# in the definition
validators: Mapping[str, _ValidatorCallback] = _typing.EmptySequence

# in _typing
EmptySequence = typing.cast(typing.Any, ())
```

Because `Any` is a subtype of everything, the rules get wonky, and `mypy` and other checkers should be okay with this.
That way, `_typing.EmptySequence` can be used in any location where the empty tuple is used today.

jsonschema/validators.py
sirosen: Okay, that makes sense. I need to think harder about what to do with schemas more broadly, but we now have a clean place where we can define `_typing.MetaSchema`. I'll make that tweak.

jsonschema/validators.py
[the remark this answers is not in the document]
Julian: That seems also fine with me (although I think the fact that `()` works is more a quirk of `dict` than a generic empty sequence type, but I'll leave pedantry out until/unless it comes back to rear its head I think).

jsonschema/_typing.py
sirosen: I'm definitely going to have to circle back to try to better grasp what's sensible here.

In line with my other comment about maybe accepting some technical inaccuracy in the annotations as a pragmatic approach, I'm instinctively inclined towards `bool | Mapping[...]`.

DanielNoord:
Nice to see you have found some more time this!

One question though, why did you choose to make the module private? A lot of the typing ends up in the public API as well, so users would then need to import from a private module to make their own annotations be correct.

jsonschema/validators.py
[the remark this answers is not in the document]
sirosen: Does it help that I also considered
```
EmptyMap = pyrsistent.freeze({})
```
?
😁 

I just know that this pattern occurs several times in the code, with `()` as an empty default. So it seems like it might be nice to sweep any messiness associated with it into `_typing`.

jsonschema/_typing.py
Julian: I have to think about what I think honestly.

To me, all I am "convinced" is useful of at the minute (and I'm quite convinced it's very useful indeed) is that typing forces documentation to have correctly documented types. That's the reason I'm quite picky about making sure that things which work "accidentally" because we abuse some type internally do not in fact get represented as the type -- because later down the line I will *not* support those internal abuses if we need to change them for whatever reason. They're not part of the public API, they don't have stability guarantees.

Here, the opposite is true -- *not* documenting the type here actually means "regressing" something that is indeed public API, and which I very much do intend folks to rely on (by which I don't mean I have no reservations that it will bite me in the ass, but I do mean it's something I think people *should* consider public API at this point, type annotation or not, and I don't intend to break it).

And I know we don't actually regress, since I'm pretty sure there are tests covering the `Any` behavior at least in part -- but if the goal (at least the goal in my head) is "be able to say every object in `jsonschema` has its type documented explicitly", it complicates that goal if the documentation again needs tweaking to *broaden* types into their real public API.

(So... "I don't know yet" :D)

sirosen:
I'm going to close this, since I don't think there's any chance we would merge it (with how the code has evolved, `referencing`, etc). I was just poking at the `jsonschema` source for something and realized how much this has drifted out of sync.

TBH, I can't remember what my full plan was here, but it probably didn't involve my past couple of super-busy years (hey, have you heard of [that PEP I wrote](https://peps.python.org/pep-0735/)? Didn't plan on that! :joy: ).
Right now, I accept that whatever the plan was, it didn't work, and we should continue to make progress in other ways. I might source some of this work to try to make other typing improvements.

PR #1050, opened 2023-02-22
Raise for status when fetching w/ requests

gadomski:
In the case of (e.g.) 404 errors that still return a response, this will let the 404 error propagate instead of a JSON decode error (if the response isn't JSON). See the last paragraph of https://requests.readthedocs.io/en/latest/user/quickstart/#json-response-content for some text about this.

### DIFF
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -2323,5 +2323,8 @@ class _ReallyFakeJSONResponse:
 
     _response = attr.ib()
 
+    def raise_for_status(self):
+        return None
+
     def json(self):
         return json.loads(self._response)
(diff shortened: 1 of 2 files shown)


Julian:
Thanks for the PR, it's appreciated. This entire interface is changing however in #1049, where you're going to be entirely in control of how you make HTTP requests ([this](https://python-jsonschema--1049.org.readthedocs.build/en/1049/referencing/#automatically-retrieving-resources-over-http) in particular is a preview of what sounds like the equivalent of what you have now, and you can certainly raise for status there).

Going to close this given it'd not see the light of day given the changeover, but feedback is definitely welcome on the new APIs if they do or don't suit you, and again, thanks for the PR regardless.

Issue #1068, opened 2023-03-24
$dynamicRefs not evaluated on error


ikonst:
If there are validation errors, a `$dynamicRef` appears not to be followed and thus not evaluated.

If such schema is set to disallow unevaluated properties, then validation would result in an unevaluated properties error in addition to the "culprit" error. In particular when using `jsonschema.validate`, the "best match" is the unevaluated property "error".

# Example

```python
jsonschema.validate(
    instance={
        'spam': 'spam',
    },
    schema={
        "$id": "file:///spam/ham.json",
        '$schema': 'https://json-schema.org/draft/2020-12/schema#',
        'unevaluatedProperties': False,
        "$dynamicRef": '#hovercraft',
        '$defs': {
            'does_not_matter': {
                "$dynamicAnchor": 'hovercraft',
                'properties': {
                    'spam': {'type': 'string'},
                },
            },
        },
        'required': ['ham'],
    }
)
```

# Actual Result

```
jsonschema.exceptions.ValidationError: Unevaluated properties are not allowed ('spam' was unexpected)
```

# Expected Result

```
jsonschema.exceptions.ValidationError: 'ham' is a required property
```

and if not for the missing `ham` property, the `spam` property would've been evaluated by the "hovercraft" definition, and thus there'd be no errors.

Julian:
Thanks for the report -- I'll have a look at your example more carefully (probably not until after the release, so few weeks). Just to be sure immediately though, have you seen the discussions on the spec repo about `unevaluatedProperties`-in-the-face-of-other-errors? I don't recall whether this specific kind of example was discussed, or even what the spec says about it, but at the minute from my recollection a bunch of this behavior is somewhere between "not specified" and "contradictory" in the spec -- specifically, I don't think there are any examples where it changes the validation result (let me know if that's your intention but seems not) -- but there definitely are examples where it's either unspecified or worse as to whether one should get 1 error or 2 in cases using `unevaluatedProperties` -- so while I would hope to have as clear of an error message as possible out of `best_match`, which I think it sounds like is what you're asking for, that may not be fully feasible, needs thinking.

[...]

PR #1073, opened 2023-03-27
Treat invalid pattern property as unevaluated

ikonst:
This addresses an "error message" issue, not a validation issue, so no need to update the test suite.

For regular properties, a property which failed validation would be also deemed "unevaluated", thus resulting in two messages, e.g.
```python
validate(
    schema={"properties": {"foo": {"type": "string"}}},
    instance={"foo": 42},
)
```
would result in 2 errors:
- 42 is not of type 'string'
- unevaluated property 'foo'

With this change, pattern properties would have the same behavior:
```python
validate(
    schema={"patternProperties": {"^foo": {"type": "string"}}},
    instance={"foo": 42},
)
```

----
:books: Documentation preview :books:: https://python-jsonschema--1073.org.readthedocs.build/en/1073/

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -291,10 +291,10 @@ def find_evaluated_property_keys_by_schema(validator, instance, schema):
 
     if "patternProperties" in schema:
         for property, value in instance.items():
-            for pattern, _ in schema["patternProperties"].items():
+            for pattern, subschema in schema["patternProperties"].items():
                 if re.search(pattern, property) and validator.evolve(
-                    schema=schema["patternProperties"],
-                ).is_valid({property: value}):
+                    schema=subschema,
+                ).is_valid(value):
                     evaluated_keys.append(property)
 
     if "dependentSchemas" in schema:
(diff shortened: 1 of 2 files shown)


ikonst:
@Julian 

I'm confused what it does right now, though. We're validating instance `{"foo": 42}` against schema `{"^foo": {"type": "number"}}` that doesn't look like a valid schema at all and valid for all inputs. For sake of argument, why call `is_valid` at all?

Julian:
Oh I see this isn't related to #1068 -- I hadn't looked at the example carefully and assumed it was what we were previously discussing. OK, I'll still probably not be able to focus on this until after 4.18, but if there's some actual bug and this doesn't add new performance impact that's a lot more likely to be mergeable.

ikonst:
I think it's a pretty straightforward change (what's there right now looks like a mix-up).

Then again, if you want to validate less, maybe instead of this PR you'd just want to remove the `is_valid`s since the tests seem to pass even in their absence:
https://github.com/python-jsonschema/jsonschema/pull/1075

ikonst:
Replaced by #1075

PR #1075, opened 2023-03-28
Do not validate for unevaluatedProperties

ikonst:
It appears that without checking `is_valid` the test suite still passes. The invalid properties already produce an error, so that the "unevaluated property" error is never _needed_ to achieve a validation failure. On the contrary, when all property validation succeeds, "unevaluated property" could still produce an error.

Arguably, producing an error for expected but invalid properties is also confusing, being a "double jeopardy" of validation failures.

----
:books: Documentation preview :books:: https://python-jsonschema--1075.org.readthedocs.build/en/1075/

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -275,12 +275,13 @@ def find_evaluated_property_keys_by_schema(validator, instance, schema):
         "properties", "additionalProperties", "unevaluatedProperties",
     ]:
         if keyword in schema:
-            if validator.is_type(schema[keyword], "boolean") and schema[keyword]:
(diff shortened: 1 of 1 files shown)


Julian:
Nice! Indeed looks like this is a decent (~40%) speedup for the (unevaluatedProperties-specific) microbenchmark that I've been referring to for the new release. 

Obviously I can believe there's gaps in the precise output for `unevaluatedProperties` (especially given the spec ambiguities), so this is potentially in need of revisiting later when `unevaluatedProperties` gets another look, but certainly seems like a win for now. Well done, merging.

PR #1120, opened 2023-07-10
perf(_validators.type): Use for-loop instead of slower any() with generator expression

jpmckinney:
I did some before/after comparisons here: https://github.com/OpenDataServices/lib-cove/pull/66

But in short it is not surprising that a simple for-loop is faster than a more indirect any() + generator.

### DIFF
--- a/jsonschema/_validators.py
+++ b/jsonschema/_validators.py
@@ -291,7 +291,10 @@ def dynamicRef(validator, dynamicRef, instance, schema):
 def type(validator, types, instance, schema):
     types = ensure_list(types)
 
-    if not any(validator.is_type(instance, type) for type in types):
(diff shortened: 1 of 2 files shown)


jpmckinney:
Aha, I'll look into that.

jpmckinney:
Benchmark results:

```
any generator: Mean +- std dev: 37.3 us +- 0.4 us
for loop: Mean +- std dev: 35.8 us +- 0.3 us
```

The `type` validator runs on pretty much every node of my very large JSON objects, so any improvement is good.

jpmckinney:
Hmm, yeah, pretty small and it wouldn't be great if the faster PyPy was slowed down via this change.

Issue #1008, opened 2022-10-24
Support standard output formats


flying-sheep:
The standard draft 2020-12 defines four standard output formats: https://json-schema.org/draft/2020-12/json-schema-core.html#name-output-formatting

- Flag - A boolean which simply indicates the overall validation result with no further details.
- Basic - Provides validation information in a flat list structure.
- Detailed - Provides validation information in a condensed hierarchical structure based on the structure of the schema.
- Verbose - Provides validation information in an uncondensed hierarchical structure that matches the exact structure of the schema.

`jsonschema` should provide ways to emit them and integrate them with its error structure:

- there should be a `Validator` method with a `format` keyword argument that just emits a Python dict conforming to the chosen format
- the ValidationErrors emitted by `iter_errors` should match the errors in the `Basic` format
- the ErrorTree structure should match the `Detailed`/`Verbose` format

Julian:
Hi there! This is certainly welcome/desired, though the implementation is a bit different.

The output formats are actually being restructured heavily still (see https://github.com/json-schema-org/json-schema-spec/pull/1249 and https://github.com/json-schema-org/json-schema-spec/issues/1320) for the next release of the specification.

What this means here is that certainly for now, though implementing them is very much desired and will be done, that the *mechanism* is going to follow more closely the API that exists here and has existed forever -- meaning -- I want this library, in its `ValidationError` [details](https://python-jsonschema.readthedocs.io/en/stable/errors/) to essentially support all of the data needed to *generate* those output formats, or any changes to them.

My quick pass previously was that that's indeed either fully possible or very nearly possible to do already today, simply by taking a bunch of `ValidationError`s and generating the specified JSON for each format. Any gaps here though are 100% stuff that should be added.

Someone (if no one else, eventually me) then should package functions into a package which depends on this one and generates the specific output format (and I'm happy to host+promote that package in this org).

So, a `jsonschema-output2020` package, which, in pseudocode supports:

```python
from jsonschema.validators import Draft2020Validator
import jsonschema_output2020
errors = Draft2020Validator(schema).iter_errors(123)
output = jsonschema_output.detailed(errors=errors)
print(json.dumps(output))
```

is what I believe is the right next step here, along with collecting any missing things from the `ValidationError` API which would support such a thing.

If/when things stabilize upstream, it's possible that the functionality can fully move into this library, though it's something we can independently think about.

> the ValidationErrors emitted by iter_errors should match the errors in the Basic format
[...]

PR #1144, opened 2023-07-27
Preserve $ref history.

atsuoishimoto:
In large schemas, understanding which $ref is being referenced when an error occurs can be helpful in debugging. We used to refer to the history of $ref with code like this:

```python
for error in validator.iter_errors(instance):
    print(validator.resolver._scopes_stack)
    print(error)
```

However, with the current version of jsonschema, validator.resolver is deprecated, and it will be not possible to refer to it.

In this PR, I have made changes so that the history of $ref is recorded in the exception object when an error occurs, and the $ref from the root to the part where the error occurred can be referred to with `error.refs`. I believe this feature will be useful not only for us, but for many users as well.

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -51,6 +51,7 @@ class _Error(Exception):
         schema_path=(),
         parent=None,
         type_checker=_unset,
+        refs=()
     ):
         super().__init__(
             message,
@@ -63,6 +64,7 @@ class _Error(Exception):
             schema,
             schema_path,
             parent,
+            refs,
         )
         self.message = message
         self.path = self.relative_path = deque(path)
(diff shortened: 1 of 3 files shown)


atsuoishimoto:
I mistakenly created this pull request. My apologies for any confusion caused.

Julian:
Your original code clearly touched something private :)

But for this one, I'm willing to consider such a thing, especially given it's required to support #1008, but not precisely int his way.

We'd also need to handle `$dynamicRef` and `$recursiveRef`, so some extra thought is needed.

But the objects used within the `referencing` library were written partially to support this (`referencing`'s `Resolver.dynamic_scope` essentially contains this information).

I'm going to close this PR given the above but please open an issue and we can discuss some other way to do this.

atsuoishimoto:
Thank you for the comment!

I filed the #1146. Thanks!

PR #1145, opened 2023-07-28
Faster pprint

atsuoishimoto:
jsonschema can sometimes be very slow when validation errors occur. This is due to the slow formatting of large schemas or data with pprint.pformat(). In this PR, I propose replacing pprint.pformat() with json.dumps(), which is expected to perform more than 10 times faster.

The following samples measure the time taken to format 300 GitHub PRs.

### pprint.pformat()
```sh
$ python3 -m timeit -s "import requests, pprint;L=requests.get('https://api.github.com/repos/python/cpython/pulls?state=all').json() * 100" "pprint.pformat(L)"
1 loop, best of 5: 14.2 sec per loop
```

### json.dumps()
```sh
$ python3 -m timeit -s "import requests, json;L=requests.get('https://api.github.com/repos/python/cpython/pulls?state=all').json() * 100" "json.dumps(L, ensure_ascii=False, indent=2, default=repr)"
1 loop, best of 5: 1.19 sec per loop
```

(Memo: If it's possible to add as a dependency, using orjson would make it more than 10 times faster.)

```sh
python3 -m timeit -s "import requests, orjson;L=requests.get('https://api.github.com/repos/python/cpython/pulls?state=all').json() * 100" "orjson.dumps(L, option=orjson.OPT_INDENT_2|orjson.OPT_SORT_KEYS, default=repr)"
5 loops, best of 5: 79.2 msec per loop
```

json.dumps() does not perform wordwrapping or other formatting pformat() does. However, such formatting can insert unnecessary newline characters into the data, sometimes making debugging hard. I believe it's more appropriate not to perform such modifications.

----
:books: Documentation preview :books:: https://python-jsonschema--1145.org.readthedocs.build/en/1145/

### DIFF
--- a/docs/errors.rst
+++ b/docs/errors.rst
@@ -216,8 +216,18 @@ easier debugging.
     3 is not valid under any of the given schemas
 
     Failed validating 'anyOf' in schema['items']:
-        {'anyOf': [{'maxLength': 2, 'type': 'string'},
-                   {'minimum': 5, 'type': 'integer'}]}
+        {
+          "anyOf": [
+            {
+              "maxLength": 2,
+              "type": "string"
+            },
+            {
+              "minimum": 5,
+              "type": "integer"
+            }
+          ]
+        }
 
     On instance[1]:
         3
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -4,7 +4,7 @@ Validation errors, and some surrounding helpers.
 from __future__ import annotations
 
 from collections import defaultdict, deque
-from pprint import pformat
+import json
 from textwrap import dedent, indent
 from typing import ClassVar
 import heapq
@@ -32,6 +32,9 @@ def __getattr__(name):
         return _RefResolutionError
     raise AttributeError(f"module {__name__} has no attribute {name}")
 
+
+def _format_json(obj):
+    return json.dumps(obj, ensure_ascii=False, indent=2)
 
 class _Error(Exception):
 
@@ -104,10 +107,10 @@ class _Error(Exception):
             {self.message}
 
             Failed validating {self.validator!r} in {schema_path}:
-                {indent(pformat(self.schema, width=72), prefix).lstrip()}
+                {indent(_format_json(self.schema), prefix).lstrip()}
 
             On {instance_path}:
-                {indent(pformat(self.instance, width=72), prefix).lstrip()}
+                {indent(_format_json(self.instance), prefix).lstrip()}
             """.rstrip(),
         )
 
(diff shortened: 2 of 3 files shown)


Julian:
Hi there, thanks for the PR.

> jsonschema can sometimes be very slow when validation errors occur. This is due to the slow formatting of large schemas or data with pprint.pformat().

This sounds slightly odd -- exceptions are only formatted when they're *printed* (or otherwise when `repr` ends up getting called) -- your application repeatedly *prints* large numbers of exceptions coming from jsonschema?

Regardless though, this PR likely can't go forward, because it isn't the case that exception contents will be JSON serializable! The instance or schema may have been deserialized using a custom `JSONDecoder` which the library won't know about, or may not have originated in JSON ini the first place!

atsuoishimoto:
Hi, thank you for comment!

> your application repeatedly prints large numbers of exceptions coming from jsonschema?

Yes, we output errors as log entries. Generally, they don't cause harm, but they can be time-consuming when outputting a JSON with numerous elements involved.

> Regardless though, this PR likely can't go forward, because it isn't the case that exception contents will be JSON serializable!

Agreed. So the [`default=repr`](https://docs.python.org/3/library/json.html#json.dump) option will handle such non-serializable objects. This is also the default behavior when pprint outputs user-defined objects.

Julian:
I still wouldn't want to use `json.dump`s, as it introduces the possibility that a schema is invalid for some instance *before* dumping but appears to be valid afterwards. This is the case if e.g. as mentioned the user used a custom deserializer, but also in cases like coercing numeric keys (properties) to strings.

Is there a reason in your logs you don't simply format exceptions in whatever way you'd like by pulling off whatever attributes you're interested in?

You might also be interested in #243 which is about making this precise function (`pformat`) injectable, and would allow you to inject `json.dumps` if you so preferred, without touching the internals. That'd be more like a solution I'd be comfortable with.

atsuoishimoto:
> I still wouldn't want to use json.dumps, as it introduces the possibility that a schema is invalid for some instance before dumping but appears to be valid afterwards. This is the case if e.g. as mentioned the user used a custom deserializer, but also in cases like coercing numeric keys (properties) to strings.

Points taken.

> Is there a reason in your logs you don't simply format exceptions in whatever way you'd like by pulling off whatever attributes you're interested in?

No, we can format our log ourselves. But I thought using json.dumps() is a reasonable default to improve jsonschema. 

I'm closing the PR. Thank you!

Julian:
Fair enough, I appreciate the offer(s) regardless! Thanks for giving it a shot.

PR #1164, opened 2023-09-15
Allow single label hostname in format string

swaeberle:
When validating hostname format strings with module fqdn, only FQDNs were accepted, as the minimum label length of the FQDN class is 2. Therefore, single label hostnames like "localhost" were rejected.

Fixed this by validating with a minimum label length of 1.

- Tests: json-schema-org/JSON-Schema-Test-Suite#685
- Fixes: python-jsonschema/jsonschema#1162

----
:books: Documentation preview :books:: https://python-jsonschema--1164.org.readthedocs.build/en/1164/

### DIFF
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -274,7 +274,7 @@ with suppress(ImportError):
     def is_host_name(instance: object) -> bool:
         if not isinstance(instance, str):
             return True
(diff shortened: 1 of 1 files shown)


Julian:
Merged, really appreciated! Will be released shortly.

Issue #1170, opened 2023-09-28
Warn when using schema features annotated with `"deprecated": true`


edgarrmondragon:
> **New in draft 2019-09** The `deprecated` keyword is a boolean that indicates that the instance value the keyword applies to should not be used and may be removed in the future.<sup>[[1]]</sup>

Usage example:

```python
from jsonschema import validators

schema = {
    "properties": {
        "id": {"type": "integer"},
        "name": {"type": "string", "deprecated": True},
    },
}
instance={"id": 1, "name": "abc"}
validator = validators.Draft202012Validator(schema, check_deprecated=True)

validator.validate(instance)  # raises ValidationError
```

I can start working on a PR if the maintainers are interested.

[1]: https://json-schema.org/understanding-json-schema/reference/generic.html#annotations

sirosen:
Just to pull in some of the broader context:
This comes from a `check-jsonschema` ticket, to which I said basically "cool but I don't know how to do this!" 🙂

It may very well be possible today. If so, a doc example would be awesome.

My first thought on the proposed interface is that it's a little odd that this would produce a validation error when the keyword's meaning is "warning like".
I think this problem, in the JSON Schema parlance, is a matter of collecting annotations (I'd love to get confirmation that I'm correct about this). In that context, it's not necessary that jsonschema raises exceptions or emits warnings, but rather that it has a way of communicating back to the caller which deprecations were hit.

I'm also not sure that this merits a special keyword argument. My ideal is that this fits in as part of a more generic mechanism like... (Just spitballing)
```python
annotations = []
validator.validate(schema, instance, collect_annotations=(annotations, "deprecated"))
print(annotations)  # [{...}]
```

Julian:
> My first thought on the proposed interface is that it's a little odd that this would produce a validation error when the keyword's meaning is "warning like".

This definitely immediately discards any solution or implementation where a `ValidationError` is raised or returned. The interpretation of getting a `ValidationError` is "the specification says your instance is invalid". For the deprecated keyword, that simply isn't the specified behavior.

But indeed I think this at least relies on or benefits from annotation support (which I think has no current open issue because no one has ever asked for its functionality! even though it's been on my radar).

If we did have such an interface, I do indeed think this would be implementable on top of it.

> I'm also not sure that this merits a special keyword argument. My ideal is that this fits in as part of a more generic mechanism like... (Just spitballing)

(I mentioned this on the PR, but indeed, agree.)

[...]

PR #1171, opened 2023-09-28
Check for the `"deprecated"` keyword

edgarrmondragon:
Closes https://github.com/python-jsonschema/jsonschema/issues/1170

----
:books: Documentation preview :books:: https://python-jsonschema--1171.org.readthedocs.build/en/1171/

### DIFF
--- a/jsonschema/_keywords.py
+++ b/jsonschema/_keywords.py
@@ -447,3 +447,8 @@ def prefixItems(validator, prefixItems, instance, schema):
             schema_path=index,
             path=index,
         )
+
+
+def deprecated(validator, deprecated, instance, schema):
+    if validator.check_deprecated and deprecated is True:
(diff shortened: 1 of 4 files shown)


Julian:
Let's discuss this on #1170, which I may not have time to write a longer response on now but will try to tomorrow. Broadly though, I wouldn't merge this PR for a few reasons (likely additional to what I'll write there)

* the Validator protocol is for general non-dialect-specific behavior. Meaning things that go there should apply to any conceivable JSON Schema dialect -- including ones that have no `deprecated` keyword whatsoever.
* I won't add boolean flags to any API
* There's an existing `extend` API for adding behavior to a dialect (which could be used to control whether any behavior is assigned to the keyword or not) without needing to add something specifically new

I'll put some additional thoughts on the issue.

And regardless, thanks for the suggestion!

PR #1208, opened 2024-01-08
Wrong behaviour for enum keyword

otto-ifak:
The following schema accepts the instance although it should be rejected:

```python
schema = {
    "enum": [
        [False]
    ],
    "$schema": "https://json-schema.org/draft/2020-12/schema"
}
instance = [0]
validate(instance, schema)  # erroneously accepts the instance
```

I guess, this is caused by the following line:
https://github.com/python-jsonschema/jsonschema/blob/256dadd3539861ae696c03805923eb4097b871f9/jsonschema/_keywords.py#L271

(`[0] in [[False]]` is true, so `[0] not in [[False]]` is false)

----
📚 Documentation preview 📚: https://python-jsonschema--1208.org.readthedocs.build/en/1208/

### DIFF
--- a/jsonschema/_keywords.py
+++ b/jsonschema/_keywords.py
@@ -8,7 +8,6 @@ from jsonschema._utils import (
     find_additional_properties,
     find_evaluated_item_indexes_by_schema,
     find_evaluated_property_keys_by_schema,
-    unbool,
     uniq,
 )
 from jsonschema.exceptions import FormatError, ValidationError
--- a/jsonschema/_keywords.py
+++ b/jsonschema/_keywords.py
@@ -264,11 +264,7 @@ def dependentSchemas(validator, dependentSchemas, instance, schema):
 
 
 def enum(validator, enums, instance, schema):
-    if instance == 0 or instance == 1:
-        unbooled = unbool(instance)
-        if all(unbooled != unbool(each) for each in enums):
-            yield ValidationError(f"{instance!r} is not one of {enums!r}")
-    elif instance not in enums:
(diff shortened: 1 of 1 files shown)


otto-ifak:
There is currently no test for this PR, maybe you could give me a hint where to put it?
Also, I am not sure if this change has a performance impact.

Julian:
Thanks for the PR! I haven't looked carefully at it yet, but as for:

> There is currently no test for this PR, maybe you could give me a hint where to put it?

The first step here would be a PR to the [official JSON Schema test suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite), specifically in the `enum.json` test file for each draft.

Once that's merged, we can bring the update in here, and then should be able to consider the PR.

Can you have a look there and let me know if you run into any trouble?

otto-ifak:
Thanks for your reply. I have created the PR:

https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/712

gregsdennis:
I'm not sure how implicit casting works in python, but is it possible that `[0]` and `[False]` are the same thing in python?

(I get that they're not the same in JSON.)

otto-ifak:
@gregsdennis yes, that is right:
In Python `0 == False` holds, so that `[0] == [False]` holds and so does `[0] in [[False]]` hold, which is the cause for this issue.

Julian:
Thanks again @otto-ifak!

otto-ifak:
@Julian you're welcome - thank you for this great library :smile:

Issue #1223, opened 2024-02-16
validating nan in an enum is no longer supported


shinnar:
As a result of the change in a9bf13ca04044e8b7320dc76a49c4dd287951686

`equal` (which uses ==) is now used for all enum (https://github.com/python-jsonschema/jsonschema/blob/main/jsonschema/_keywords.py#L269)  validations.

As a result, if an enum has a `nan` in it, it will no longer validate a `nan` instance, since `nan == nan` returns False.
The previous behaviour used `in`, which calls the containers `__contains__` method.  For standard containers, that uses both `is` and `==` to check for comparison, which is why this worked before this commit.

One possible fix would be to check if `one is two` at the beginning of the `equal` method.

Julian:
Similar to #1222 `NaN` doesn't exist in JSON, and therefore not in JSON Schema so it isn't really supported by this library and there's certainly no guarantees about how it behaves. That being said, I'm happy nonetheless to merge your PR, but only really because an `is` check might help in a tiny way for *other* comparison speedups. There again would be no promises about behavior of `nan` in particular.

PR #1224, opened 2024-02-16
Check for identity before equality when comparing objects with `equal` (used in a number of keywords)

shinnar:
This addresses #1223 

----
📚 Documentation preview 📚: https://python-jsonschema--1224.org.readthedocs.build/en/1224/


Julian:
Thanks!

PR #1253, opened 2024-05-03
fix: Ignore validators set to 'None'

dodslaser:
Fixes a regression in 4.22.0 where validators set to 'None' are not ignored, causing a `TypeError: 'NoneType' object is not callable` on validation. In version 4.21.1 and earlier, any validator set to None would be ignored.

----
📚 Documentation preview 📚: https://python-jsonschema--1253.org.readthedocs.build/en/1253/

### DIFF
--- a/jsonschema/tests/test_validators.py
+++ b/jsonschema/tests/test_validators.py
@@ -46,7 +46,7 @@ class TestCreateAndExtend(TestCase):
         )
 
         self.meta_schema = {"$id": "some://meta/schema"}
-        self.validators = {"fail": fail}
+        self.validators = {"fail": fail, "none": None}
         self.type_checker = TypeChecker()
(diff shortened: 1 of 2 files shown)


Julian:
Hey there.

This isn't something I intentionally changed, but it's also not a regression really, the type of validators is (always was) [documented to be callables with 4 arguments](https://python-jsonschema.readthedocs.io/en/stable/creating/) -- if you want one that does nothing just don't add it at all to `extend`, or add it and make it be a function which does nothing, but yeah I don't make any guarantees on what any sentinel value will do here, this all may change.

Julian:
Definitely trust the docs for what's public API yeah.

Not including the keyword at all should be even faster if you want to micro-optimize, but yeah I doubt it will show up in any case.

(Thanks regardless for the PR).

Issue #1257, opened 2024-05-15
Mitigate undesired side effect of new `best_match` behaviour with alternative proposal


ilia1243:
In continuation to #1250. The fix has seemingly undesired side effect.

```python
from jsonschema import Draft202012Validator as Validator, exceptions

schema = {'oneOf': [
    {'properties': {'run': {'type': 'string'}}, 'required': ['run']},
    {'properties': {'uses': {'type': 'string'}}, 'required': ['uses']},
]}
instance = {'uses': 1, 'run': 1}

error = exceptions.best_match(Validator(schema).iter_errors(instance))
print(schema, "\n\n", error)
```
After the fix it produces:
```
 1 is not of type 'string'
On instance['run']:
    1
```
Conceptually, it is not clear why `run` has priority (I understand that technically it has priority due to the alphabetical order).

Based on real example https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-workflow.json

As an alternative proposal, the relevance function could not be changed, but `best_match` could distinguish errors having different error.path in the same subscheme, and choose first of them (maybe having even not the minimal relevance, but the maximum one for this particular subscheme provided that it is still minimal among different subschemes).

ilia1243:
Opened pull request with possible improvement of `best_match`.

nfriedl1:
Hi, I think I am running into the same thing, but I wanted to share my example just in case.

```python
from jsonschema import validate
from jsonschema.exceptions import ValidationError

schema = {
    "type": "object",
    "properties": {
        "tuple_or_tuple_array": {
            "anyOf": [
                {
                    "type": "array",
                    "items": {
                        "$ref": "#/$defs/int_tuple",
                    },
                },
                {
                    "$ref": "#/$defs/int_tuple",
                },
            ]
        },
    },
    "$defs": {
        "int_tuple": {
            "type": "array",
            "items": {
                "type": "integer",
            },
            "minItems": 2,
            "maxItems": 2,
        },
    },
}


def try_validate(instance):
    try:
        validate(instance, schema)
    except ValidationError as e:
        print(str(e))
    else:
        print("No error.\n")
    finally:
        print("-" * 80)


# Case 1
control = {"tuple_or_tuple_array": [0, 0]}
try_validate(control)

# Case 2
control2 = {"tuple_or_tuple_array": [[0, 0], [0, 0]]}
try_validate(control2)

# Case 3
[...]

PR #1258, opened 2024-05-19
Pick the deepest error among the most relevant ones in each subschema

ilia1243:
Improves `best_match` in the presence of `anyOf` / `oneOf`. Calculate the most relevant error in each separate subschema and choose the deepest one.

In particular, for `anyOf` / `oneOf` keywords with the only subschema, the best error is resolved as if the subschema was not enclosed by these keywords.

To reproduce:

```python
from jsonschema import Draft202012Validator as Validator, exceptions

for applicator in "anyOf", "oneOf":
    # Should match {"properties": {"foo": {"minProperties": 2}}
    schema = {
        applicator: [
            {
                "properties": {
                    "foo": {
                        "minProperties": 2,
                        "properties": {"bar": {"type": "object"}},
                    },
                },
            },
        ],
    }
    instance = {"foo": {"bar": []}}
    error = exceptions.best_match(Validator(schema).iter_errors(instance))
    print(error)
```

Revert main code changes in commit https://github.com/python-jsonschema/jsonschema/commit/b20234e86c4dadf5d691400383a6fc0a1e9afc34 preserving the tests.

Closes: #1257

----
📚 Documentation preview 📚: https://python-jsonschema--1258.org.readthedocs.build/en/1258/

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -397,7 +397,6 @@ def by_relevance(weak=WEAK_MATCHES, strong=STRONG_MATCHES):
         validator = error.validator
         return (                        # prefer errors which are ...
             -len(error.path),           # 'deeper' and thereby more specific
-            error.path,                 # earlier (for sibling errors)
             validator not in weak,      # for a non-low-priority keyword
             validator in strong,        # for a high priority keyword
             not error._matches_type(),  # at least match the instance's type
@@ -427,7 +426,8 @@ def best_match(errors, key=relevance):
     since they indicate "more" is wrong with the instance.
 
     If the resulting match is either :kw:`oneOf` or :kw:`anyOf`, the
-    *opposite* assumption is made -- i.e. the deepest error is picked,
+    *opposite* assumption is made -- i.e. the deepest error is picked
+    among the most relevant errors in each separate subschema,
     since these keywords only need to match once, and any other errors
     may not be relevant.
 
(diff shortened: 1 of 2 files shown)


Julian:
Thanks! I'll have a look at this in the next day or two but initially looks reasonable!

jsonschema/exceptions.py
```diff
@@ -464,9 +464,19 @@ def best_match(errors, key=relevance):
     best = max(itertools.chain([best], errors), key=key)
 
     while best.context:
+        # Calculate the most relevant error in each separate subschema
+        best_in_subschemas = [None] * len(best.validator_value)
```
Julian, line 468: I know I'm still behind on looking at this, but just to mention why I haven't yet merged it, I want to stare at this closer, as I'm somewhat suspicious of the types here -- specifically, it's likely true that `validator_value` is always a container for any builtin JSON Schema keywords, but certainly not in general (i.e. someone can invent some other one).

And similarly below I'm slightly suspicious of the `schema_path` business.

Not to say any of it looks wrong, but I want to convince myself if it needs any additional hardening.

And of course again thanks for the thought and PR, I'll get to it sometime soon I hope (days not weeks).

ilia1243: If assuming that only `oneOf` and `anyOf` can have non-empty context, then schema_path is seemingly always an integer because it was produced from enumerating https://github.com/python-jsonschema/jsonschema/blob/v4.22.0/jsonschema/_keywords.py#L340
This also implies that yes, `validator_value` may be something that can be enumerated, but without `len`.

Alternatively, assuming that best.context has errors in strict order of `.schema_path` (see also the same algorithms `_keywords.anyOf | oneOf`):

```python
        best_in_subschemas = []
        for error in best.context:
            index = error.schema_path[0]
            if index == len(best_in_subschemas):
                best_in_subschemas.append(error)
            else:
                prev = best_in_subschemas[index]
                best_in_subschemas[index] = max(prev, error, key=key)
```

ilia1243:
Superseded by #1300 taking into account https://github.com/python-jsonschema/jsonschema/pull/1258#discussion_r1620718054 and with test covering a case by @nfriedl1

PR #1282, opened 2024-07-08
Declare support for 3.13

rominf:
Fixes: https://github.com/python-jsonschema/jsonschema/issues/1264.

----
📚 Documentation preview 📚: https://python-jsonschema--1282.org.readthedocs.build/en/1282/


noxfile.py
```diff
@@ -35,7 +35,7 @@
     "The Unlicense (Unlicense)",
 ]
 
-SUPPORTED = ["3.8", "3.9", "3.10", "pypy3.10", "3.11", "3.12"]
+SUPPORTED = ["3.8", "3.9", "3.10", "pypy3.10", "3.11", "3.12", "3.13"]
 LATEST = SUPPORTED[-1]
```
Julian, line 39: 3.13 is still not released, so we should "leave this" pointing to "3.12" (by hardcoding it instead of doing `[-1]` I guess).

That also presumably explains the CI failure(s), some things don't support 3.13 wheels yet.

Otherwise this looks good!

rominf: Thank you! The line is now fixed as suggested. I noticed `rpds-py` had no wheels because of this line: https://github.com/crate-py/rpds/blob/0b6ea677974fdf2856d975e9386e16bec441424f/.github/workflows/CI.yml#L164. I will open the PR for building the wheels in https://github.com/crate-py/rpds/.

Julian:
Perfect, thanks again!

noxfile.py
rominf: Here is the PR for building wheels for Python 3.13: https://github.com/crate-py/rpds/pull/80.

PR #1326, opened 2025-01-20
docs: add link to check-jsonschema for how to handle generic yaml parsing

bkueng:
I was running into this and thought it would be useful to have a link to how check-jsonschema handles it.

----
📚 Documentation preview 📚: https://python-jsonschema--1326.org.readthedocs.build/en/1326/


Julian:
Thanks!

Issue #1338, opened 2025-03-05
Mutating validators are difficult to debug


stephenfin:
I saw strange issues while adding schemas to a project I was working on. Despite knowing I was passing a valid instance in, I was getting errors like `{} is not valid under any of the given schemas`. I eventually traced this down to one of the in-tree validators ([here](https://github.com/openstack/nova/blob/master/nova/api/validation/validators.py#L182), fwiw), which has the unfortunate side-effect of mutating the input `instance`. When the schema is using `oneOf`, the first iteration ended up deleting all keys from the `instance`, causing the subsequent runs to all fail.

This is 100% not an issue with `jsonschema` but it is something I think it might be able to help mitigate. I could see us either (a) always doing a `deepcopy` of the instance before passing it to the validator, or (b) doing a deepcopy and comparing against `instance` after the call, with an exception raised if they're not identical. There might also be other approaches possible. In any case, I'm happy to help resolve this, assuming we agree it's a thing we'd like to address.

PR #1339, opened 2025-03-05
Raise exception if validator mutates instance

stephenfin:
There are test failures which I have yet to investigate in depth. In addition, this may not be the direction we want to go (if there is a direction). Hopefully it can kickstart the conversation though.

Closes: #1338

----
📚 Documentation preview 📚: https://python-jsonschema--1339.org.readthedocs.build/en/1339/

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -292,6 +292,20 @@ class UnknownType(Exception):
         )
 
 
+class MutatingValidator(Exception):
+    """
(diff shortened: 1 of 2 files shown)


Julian:
Hey! Thanks for raising. Closing as per https://github.com/python-jsonschema/jsonschema/issues/1338#issuecomment-2754710342 but still appreciated and willing to continue discussing there!

PR #1366, opened 2025-06-03
1365 evaluation order

### LINKED ISSUE #1365: `unevaluatedProperties` do not treat "in-place applicators" per the spec
`unevaluatedProperies` should be applied _after_ validating all subschemas, such that `ValidationErrors` from the subschemas are raised/yielded _before_ raising errors for `unevaluatedProperies`

From https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#name-unevaluatedproperties:
> "properties", "patternProperties", "additionalProperties", and all in-place applicators MUST be evaluated before this keyword can be evaluated. 

The reference to "in-place applicators" includes logical keywords like `allOf` and conditional keywords like `if` and `then`.

From this documentation, it proceeds that:
- Properties from _valid_ subschemas (recursively) are considered "evaluated"
- Properties from _invald_ subschemas are evaluated _first_, raising/yielding any ValidationErrors, _before_ being considered "unevaluated".

This makes a significant difference when failing on the first error encountered, where the "unevaluated properties are not allowed" message is misleading for a property that is present in the schema but fails validation - especially when the validation error is nested deeply within the "unevaluated" property.

## Environment

```bash
$ pip list
```
```
Package                   Version
------------------------- --------
attrs                     25.3.0
jsonschema                4.24.0
jsonschema-specifications 2025.4.1
pip                       25.1.1
referencing               0.36.2
rpds-py                   0.25.1
setuptools                80.3.1
typing_extensions         4.14.0
wheel                     0.45.1
```

## Example

The (simplified) example schema attempts to define a generic "signal" schema, which may have different properties depending on it's "data type". This is arguably a bad idea but, good or bad, is possible in draft 2020-12:
```json
{
  "$id": "signal.json",
  "unevaluatedProperties": false,
  "required": ["source","destination","data type"],
  "properties": {
    "data type": {"enum": ["float","string"]},
    "source": true,
    "destination": true,
  },
  "allOf": [
    {
      "if": {"properties": {"data type": {"const": "float"}}},
      "then": {"properties": {"precision": {"type": "number"}}}
    },
    {
      "if": {"properties": {"data type": {"const": "string"}}},
      "then": {"properties": {"encoding": {"enum": ["ascii", "utf-8"]}}}
    }
  ]
}
```

This works as expected for well-formed objects:
```python
validator = jsonschema.Draft202012Validator(schema)
validator.validate({"source": "A/C", "destination": "Store", "data type": "string", "encoding": "utf-8"})
validator.validate({"source": "A/C", "destination": "Store", "data type": "float", "precision": 64})
```

When a sub-schema gives unexpected results, the `unevaluatedProperties` error is wrongly printed, instead 
```python
validator.validate({"source": "A/C", "destination": "Store", "data type": "string", "encoding": "utf-16"})
```
```
ValidationError: Unevaluated properties are not allowed ('encoding' was unexpected)

Failed validating 'unevaluatedProperties' in schema:
    {'$id': 'signal.json',
     '$schema': 'https://json-schema.org/draft/2020-12/schema',
     'unevaluatedProperties': False,
     'required': ['source', 'destination', 'data type'],
     'properties': {'data type': {'enum': ['float', 'string']},
                    'source': True,
                    'destination': True},
     'allOf': [{'if': {'properties': {'data type': {'const': 'float'}}},
                'then': {'properties': {'precision': {'type': 'number'}}}},
               {'if': {'properties': {'data type': {'const': 'string'}}},
                'then': {'properties': {'encoding': {'enum': ['ascii',
                                                              'utf-8']}}}}]}

On instance:
    {'source': 'A/C',
     'destination': 'Store',
     'data type': 'string',
     'encoding': 'utf-16'}
```

This is not incorrect behavior, and a somewhat misleading message. The real problem is that `utf-16` is not an allowed value, not that `encoding` is unevaluated. Only when we print _all_ errors, we find the real (or at least original) error:
```python
import traceback
for error in validator.iter_errors({"source": "A/C", "destination": "Store", "data type": "string", "encoding": "utf-16"}):
    traceback.print_exception(error)
```
```
jsonschema.exceptions.ValidationError: Unevaluated properties are not allowed ('encoding' was unexpected)

Failed validating 'unevaluatedProperties' in schema:
    {'$id': 'signal.json',
     '$schema': 'https://json-schema.org/draft/2020-12/schema',
     'unevaluatedProperties': False,
     'required': ['source', 'destination', 'data type'],
     'properties': {'data type': {'enum': ['float', 'string']},
                    'source': True,
                    'destination': True},
     'allOf': [{'if': {'properties': {'data type': {'const': 'float'}}},
                'then': {'properties': {'precision': {'type': 'number'}}}},
               {'if': {'properties': {'data type': {'const': 'string'}}},
                'then': {'properties': {'encoding': {'enum': ['ascii',
                                                              'utf-8']}}}}]}

On instance:
    {'source': 'A/C',
     'destination': 'Store',
     'data type': 'string',
     'encoding': 'utf-16'}
jsonschema.exceptions.ValidationError: 'utf-16' is not one of ['ascii', 'utf-8']

Failed validating 'enum' in schema['allOf'][1]['then']['properties']['encoding']:
    {'enum': ['ascii', 'utf-8']}

On instance['encoding']:
    'utf-16'
```

hackowitz-af:
Closes #1365

## jsonschema/tests/test_validators.py
Adds a test method for the expected behavior, and reverences the issue being solved. The method is slightly longer than others to avoid `TestValidationErrorMessages.message_for`, which asserts that there be exactly one error.

## jsonschema/_keyword.py
The `Keyword` class can define additional tags/behavior for the implementation of the keyword. For now, I have only added a "needs" label, which names other keywords that it cannot be executed before. The name "needs" may be inaccurate, and I'd be happy to change it.

The `keyword` decorator makes it trivial to annotate a function to be used as a validator keyword.

`keyword` and `Keyword` are both optional and completely back-compatible. They would only be necessary for new behavior. There is no change to existing validator function behavior.

## jsonschema/validators.py
The only changed behavior is that the `iter_errors` evaluates `Keyword`s _after_ their `needs`. The logic is not as clean as I'd like, but is reasonably straightforward:
- Keep a memo of `todo` keywords.
- As keywords are evaluated, they are no longer `todo`.
- Any `Keyword` that `needs` a `todo` keyword is labeled `dependent` and skipped
- If _all_ attempted keywords were `dependent` of each other, there is a circular dependency and an error is raised.
- `dependent` keywords are re-attempted.

I'd be happy to add comments or a docstring to make the code more self-descriptive if necessary.

----
📚 Documentation preview 📚: https://python-jsonschema--1366.org.readthedocs.build/en/1366/

### DIFF
--- a/jsonschema/_keywords.py
+++ b/jsonschema/_keywords.py
@@ -28,7 +28,7 @@ class Keyword:
 
     def __init__(
             self,
-            func: typing.Optional[callable] = None, /,
+            func: typing.Optional[typing.Callable] = None, /,
             needs: typing.Optional[typing.Iterable[str]] = (),
     ):
         self.func = func
@@ -40,7 +40,7 @@ class Keyword:
         return self.func(validator, property, instance, schema)
 
 
-def keyword(func: typing.Optional[callable] = None, /, **kwargs):
+def keyword(func: typing.Optional[typing.Callable] = None, /, **kwargs):
     """
     Syntactic sugar to decorate a function as a keyword.
 
--- a/jsonschema/_keywords.py
+++ b/jsonschema/_keywords.py
@@ -1,5 +1,8 @@
+from __future__ import annotations
+
 from fractions import Fraction
 import re
+import typing
 
 from jsonschema._utils import (
     ensure_list,
@@ -11,6 +14,58 @@ from jsonschema._utils import (
     uniq,
 )
 from jsonschema.exceptions import FormatError, ValidationError
+
+
+class Keyword:
+    """
+    The Keyword wraps a function to maintain full back-compatibility.
+
+    The class adds the ability to annotate the keyword implementation while
+    behaving like a function, i.e. `Keyword(func)` behaves like `func` itself
+    in most cases, and allows a more formal description of the behavior and
+    dependencies.
+    """
+
+    def __init__(
+            self,
+            func: typing.Optional[callable] = None, /,
+            needs: typing.Optional[typing.Iterable[str]] = (),
+    ):
+        self.func = func
+        self.needs = needs
+
+    def __call__(self, validator, property, instance, schema):
+        if self.func is None:
+            raise NotImplementedError(f"{self} has no validation method.")
+        return self.func(validator, property, instance, schema)
+
+
+def keyword(func: typing.Optional[callable] = None, /, **kwargs):
+    """
+    Syntactic sugar to decorate a function as a keyword.
+
+    The function is not modified, but wrapped by a Keyword object. This allows
+    the same function to be reused as by multiple validators with different
+    annotations.
+
+    Examples
+    --------
+    Functionally equivalent to the un-decorated function.
+    >>> @keyword
+    ... def myKeyword(validator, property, instance, schema):
+    ...     ...
+
+    Mark the keyword "else" to be evaluated after "if"
+    >>> @keyword(needs=["if"])
+    ... def else_(validator, property, instance, schema):
+    ...     ...
+
+    >>> kw = keyword(else_, needs="if")
+
+    """
+    if func is None:
+        return lambda fun: Keyword(fun, **kwargs)
+    return Keyword(func, **kwargs)
 
 
 def patternProperties(validator, patternProperties, instance, schema):
@@ -389,6 +444,21 @@ def if_(validator, if_schema, instance, schema):
         yield from validator.descend(instance, else_, schema_path="else")
 
 
+@keyword(
+    needs=[
+        "prefixItems",
+        "items",
+        "contains",
+        "allOf",
+        "anyOf",
+        "oneOf",
+        "not",
+        "if",
+        "then",
+        "else",
+        "dependentSchemas",
+    ],
+)
 def unevaluatedItems(validator, unevaluatedItems, instance, schema):
     if not validator.is_type(instance, "array"):
         return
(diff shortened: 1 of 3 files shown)


Julian:
Thanks sincerely for giving this a shot! Like I said on the issues, I'd like this done in a more systematic way. Specifically the concerns include:

* this code is quite performance sensitive so adding extra indirection here is quite possibly harmful on CPython and needs careful benchmarking
* more importantly, I don't want to bolt on the notion of dependence order to the existing APIs -- doing this right I think needs to involve a rewrite of the validator protocol (which I'm referring to as "dialects" as that's the modern name for the concept in JSON Schema) in order to support annotations (== state + dependency order). Such an API should allow users to specify ordering, to modify ordering when extending validators, and then should likely do things like topologically sort the order at validator-build time in order to ensure the order is well defined and known ahead of time when compiling a schema. Such a thing is a larger project, which is why it hasn't happened yet, but it's the "right" next step to me.

If you'd like to help, which is very much appreciated, you could review the Dialect v2 label, help by speccing out what a new protocol might look like (I have ideas here clearly but haven't tried to write them out), or help investigate our existing benchmarks and thereby help ensure we're as good or better when we introduce this dependency notion.

I'm going to close this as I wouldn't merge it as an incremental change as I said, but again, I do appreciate your work!

hackowitz-af:
All good, thanks for the feedback! I'll see about speccing out a new protocol in the next few months, and getting familiar with the related issues.

PR #1367, opened 2025-06-04
Drop python<3.9 backports, since the package already requires python>=3.9

hackowitz-af:


hackowitz-af:
Since the package already requires python>=3.9, I think there is no harm here

PR #1390, opened 2025-07-16
Unambiguously quote and escape properties in JSON path rendering

### LINKED ISSUE #1389: `_Error.json_path` produces invalid or questionable JSON paths
The `json_path` property on the `_Error` class does not escape or otherwise account for property names that are either blank or have non-alphanumeric characters.

For example, a property named `.` currently renders to JSON path as `$..`. It should instead be `$['.']`.

Here is a script that demonstrates the `json_path` rendering of various valid property names, which result in invalid or questionable JSON paths:

```python
import jsonschema


# Keys that currently produce invalid or questionable JSON paths.
keys = ["", "[", ".", "\\", "'", " ", "\"", "a[0]"]

schema = {
    # Require that each key be a string.
    "properties": {key: {"type": "string"} for key in keys},
}

instance = {
    # Set each key to a non-string value.
    key: None for key in keys
}

validator = jsonschema.Draft7Validator(schema)
for error in validator.iter_errors(instance):
    print(f"property={error.path[-1]!r}\t\tjson_path={error.json_path!r}")
```

And the output:

```shell
$ python demo.py 
property=''             json_path='$.'
property='['            json_path='$.['
property='.'            json_path='$..'
property='\\'           json_path='$.\\'
property="'"            json_path="$.'"
property=' '            json_path='$. '
property='"'            json_path='$."'
property='a[0]'         json_path='$.a[0]'
```

kurtmckee:
This PR introduces the following change:

* Unambiguously quote and escape properties in JSON path rendering

Using the script from #1389, this is the new output with the changes in this branch:

```shell
$ python demo.py 
property=''             json_path="$['']"
property='['            json_path="$['[']"
property='.'            json_path="$['.']"
property='\\'           json_path="$['\\\\']"
property="'"            json_path="$['\\'']"
property=' '            json_path="$[' ']"
property='"'            json_path='$[\'"\']'
property='a[0]'         json_path="$['a[0]']"
```

Fixes #1389

----
📚 Documentation preview 📚: https://python-jsonschema--1390.org.readthedocs.build/en/1390/

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -8,6 +8,7 @@ from pprint import pformat
 from textwrap import dedent, indent
 from typing import TYPE_CHECKING, Any, ClassVar
 import heapq
+import re
 import warnings
 
 from attrs import define
@@ -22,6 +23,8 @@ if TYPE_CHECKING:
 
 WEAK_MATCHES: frozenset[str] = frozenset(["anyOf", "oneOf"])
 STRONG_MATCHES: frozenset[str] = frozenset()
+
+JSON_PATH_COMPATIBLE_PROPERTY_PATTERN = re.compile("^[a-zA-Z][a-zA-Z0-9_]*$")
 
 _unset = _utils.Unset()
 
@@ -152,8 +155,11 @@ class _Error(Exception):
         for elem in self.absolute_path:
             if isinstance(elem, int):
                 path += "[" + str(elem) + "]"
-            else:
+            elif JSON_PATH_COMPATIBLE_PROPERTY_PATTERN.match(elem):
                 path += "." + elem
+            else:
+                escaped_elem = elem.replace("\\", "\\\\").replace("'", r"\'")
+                path += "['" + escaped_elem + "']"
         return path
 
     def _set(
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -24,7 +24,7 @@ if TYPE_CHECKING:
 WEAK_MATCHES: frozenset[str] = frozenset(["anyOf", "oneOf"])
 STRONG_MATCHES: frozenset[str] = frozenset()
 
-JSON_PATH_COMPATIBLE_PROPERTY_PATTERN = re.compile("^[a-zA-Z][a-zA-Z0-9_]*$")
+_JSON_PATH_COMPATIBLE_PROPERTY_PATTERN = re.compile("^[a-zA-Z][a-zA-Z0-9_]*$")
 
 _unset = _utils.Unset()
 
@@ -155,7 +155,7 @@ class _Error(Exception):
         for elem in self.absolute_path:
             if isinstance(elem, int):
                 path += "[" + str(elem) + "]"
-            elif JSON_PATH_COMPATIBLE_PROPERTY_PATTERN.match(elem):
+            elif _JSON_PATH_COMPATIBLE_PROPERTY_PATTERN.match(elem):
                 path += "." + elem
             else:
                 escaped_elem = elem.replace("\\", "\\\\").replace("'", r"\'")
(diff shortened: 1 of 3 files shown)


Julian:
Thanks for the PR (and issue).

I think we should consider and look at [jsonpath-ng](https://github.com/h2non/jsonpath-ng) at very least to start with for testing. It's possible in the future we should deprecate the attribute and replace it with one that returns an object from that library, but even before we do that we can at least use it in our test cases to:

* ensure all the tests you're adding actually are guarded to be valid JSON Paths
* ensure we have covered at least the cases that that library itself tests

Could you have a look at doing that? As I say I'm good with just adding it as a test dependency for the moment and evaluating the deprecation/changing the actual interface down the line.

jsonschema/exceptions.py
```diff
@@ -23,6 +24,8 @@
 WEAK_MATCHES: frozenset[str] = frozenset(["anyOf", "oneOf"])
 STRONG_MATCHES: frozenset[str] = frozenset()
 
+JSON_PATH_COMPATIBLE_PROPERTY_PATTERN = re.compile("^[a-zA-Z][a-zA-Z0-9_]*$")
```
Julian, line 27: Please make sure this is private even if we don't use `jsonpath-ng` in the real code yet. Though if it turns out that a regex isn't good enough for fully accurate generation of paths then I think we should think carefully about adding the dep even now perhaps.

jsonschema/tests/test_exceptions.py
```diff
@@ -700,3 +700,122 @@ class TestHashable(TestCase):
     def test_hashable(self):
         {exceptions.ValidationError("")}
         {exceptions.SchemaError("")}
+
+
+class TestJsonPathRendering(TestCase):
+    def test_str(self):
+        e = exceptions.ValidationError(
```
Julian, line 707: Let's add a `self.error_with_path(...) -> ValidationError` test helper here so the duplication is removed when all we really care about is the one attribute essentially.

And as I said in the top-level comment I think adding an assertion that `jsonpath-ng` considers our output valid is a good immediate safeguard.

kurtmckee: I'm also dissatisfied with the repetitive code, so I'm glad you called this out.

I'm accustomed to using pytests test parametrization feature, and found something similar in unittest just now: [subTest](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests).

I can introduce an `error_with_path()` method, but would you be open to something like:

```python
class TestJsonPathRendering(TestCase):
    def test_strange_properties(self):
        parameters = (
            ("", "$[' ']"),
            (".", "$['.']"),
            # ...
        )
        for property, expected_path in parameters:
            with self.subTest(property=property):
                e = exceptions.ValidationError(
                    path=[property],
                    message="1",
                    validator="foo",
                    instance="i1",
                )
                self.assertEqual(e.json_path, expected_path)
```

_(Wrote that in this comment, so I don't know if it'll actually run as-is.)_

Let me know if that's appealing, otherwise I'll go with the test helper you proposed.

Either way, I'll add an assertion that jsonpath-ng can parse the resulting JSON path. :+1:

kurtmckee: Super, I'll put together those changes.

jsonschema/exceptions.py
kurtmckee: This is addressed.

jsonschema/tests/test_exceptions.py
kurtmckee: The unittest `subTest` feature doesn't appear to be on par with pytest's parametrization feature -- when I induced deliberate failures I only saw the first one show up. It also didn't appear to increase the count of tests that were running.

This may have been an issue on my part, but I went with your original suggestion to have a helper function.

I think this feedback is addressed now. :+1:

PR #1396, opened 2025-08-11
Fix `Validator` protocol init to match runtime

sirosen:
_Addresses #1382_

The runtime init signature of a validator can be seen with `inspect`:

```pycon
>>> import inspect, jsonschema
>>> inspect.signature(jsonschema.validators.Draft7Validator.__init__)
<Signature (self, schema: 'referencing.jsonschema.Schema', resolver=None, format_checker: '_format.FormatChecker | None' = None, *, registry: 'referencing.jsonschema.SchemaRegistry' = <Registry (20 resources)>, _resolver=None) -> None>
```

This aligns the protocol's declared signature with that behavior more exactly with the following changes:

-   `registry` is now keyword-only, not keyword-or-positional, and has a
    default.

-   `resolver` is added to the declared signature, so users who are using
    it won't see typing-time discrepancies with the runtime. It is marked
    as `Any` and commented inline as deprecated, since it's unclear what
    else we could do to indicate its status.
    This means that code passing a resolver will continue to type check
    (previously it would not).

-  `resolver` is the second keyword-or-positional and `format_checker` is
    the third, meaning that a positional-only caller who passes, for
    example: `Draft202012Validator(foo, None, bar)` will have `foo`
    slotted as the schema and `bar` as the `format_checker`
    This would primarily impact callers with positional-args calling
    conventions, but is more reflective of what they'll see at runtime.

In order to remove `resolver` from the protocol signature, but match the runtime signatures well, some kind of placeholder is needed to indicate `format_checker` as positional-or-keyword.
Or else, a large number of overloads for `__init__` could be declared to try to simulate its removal.

I considered such options but decided that they add undue complexity for what is supposed to be a *hint* for correct usage.

Assuming this or something like it merges, I'll backport to typeshed.

----
📚 Documentation preview 📚: https://python-jsonschema--1396.org.readthedocs.build/en/1396/

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -183,7 +183,7 @@ def typing(session):
     session.install("mypy", "types-requests", ROOT)
     session.run("mypy", "--config", PYPROJECT, PACKAGE)
     session.run(
-        "mypy", "--config", PYPROJECT, "--warn-unused-ignores", TYPING_TESTS
+        "mypy", "--config", PYPROJECT, "--warn-unused-ignores", TYPING_TESTS,
     )
 
 


Julian:
Thanks! Merging main will fix the license check, I still haven't gotten a chance to write a new tool unfortunately :/

The change looks correct obviously, but I'd love to have some actual tests... When I tried to fix this a few weeks ago I added a `jsonschema/tests/typing_tests.py` with:

```python
"""
Static typing tests for jsonschema.
"""

from typing import assert_type

from jsonschema import Draft202012Validator
from jsonschema.protocols import Validator

assert_type(Draft202012Validator, type[Validator])
```

to at least address this specific test.

I forget what issue I ran into which made me unable to just add that immediately and fix this. Does it pass after your change here or do we run into some other broken annotation? If it passes (or if you otherwise have a recommendation for what to add to do this) could you add that as well?

Julian:
OK, cool thanks! Will wait to hear back. I guess given we have a `runtime_checkable` `Protocol`, we could just `isinstance` it too?? But will wait to hear if you come up with anything.

edgarrmondragon:
I'm curious what's the process for backporting this to typeshed, to see if I can help 🙂. Just open a PR and don't touch `METADATA.toml` since we're still in the 4.25 line?

sirosen:
@edgarrmondragon, the offer is very much appreciated! I've actually already opened https://github.com/python/typeshed/pull/14591 which I _think_ is what we need (but a review of that is welcome).

PR #1487, opened 2026-05-23
tests: fix test_uses_pprint for Python 3.15 pprint dict

traylenator:
Python 3.15 changed pprint's output for dicts with non-string keys

* https://docs.python.org/3.15/whatsnew/3.15.html#pprint
* https://github.com/python/cpython/issues/112632

### DIFF
--- a/jsonschema/tests/test_exceptions.py
+++ b/jsonschema/tests/test_exceptions.py
@@ -1,5 +1,7 @@
 from unittest import TestCase
+import sys
 import textwrap
+import unittest
 
 import jsonpath_ng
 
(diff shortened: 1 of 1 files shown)


Julian:
Thanks, we should add 3.15 as an env before fixing issues here I think.

traylenator:
I should have mentioned I realise you are not declaring support for 3.15 yet understandably.
I needed it for the 3.15 preparations in Fedora.

https://copr.fedorainfracloud.org/coprs/g/python/python3.15/package/python-jsonschema

> Thanks, we should add 3.15 as an env before fixing issues here I think.

What ever makes sense for you.

traylenator:
It was only a problem in one of the beta releases - no longer required.

PR #1495, opened 2026-06-11
Avoid mutating ErrorTree on missing child lookup

### LINKED ISSUE #1328: The return of `__iter__()` and `__contains__()` change after accessing of an index with no error
The following example illustrate that accessing of an index in an `ErrorTree` object that has no error can added the index to the collection of indices considered having errors.

```python
from jsonschema import Draft202012Validator
from jsonschema.exceptions import ErrorTree

schema = {
    "type" : "array",
    "items" : {"type" : "number", "enum" : [1, 2, 3]},
    "minItems" : 3,
}
instance = ["spam", 2]

v = Draft202012Validator(schema)
for error in sorted(v.iter_errors(instance), key=str):
    print(error.message)

print("\n====================================")
tree = ErrorTree(v.iter_errors(instance))

print(f"correct value of list(tree): {list(tree)}")
"correct value of list(tree): [0]"
print(f"correct value of `1 in tree`: {1 in tree}")
"correct value of `1 in tree`: False"

# Accessing the 1 index changes the iteration of the tree
tree[1]

print("\n ===== after accessing the 1 index of the tree =====")
print(f"incorrect value of list(tree): {list(tree)}")
"incorrect value of list(tree): [0, 1]"
print(f"incorrect value of `1 in tree`: {1 in tree}")
"incorrect value of `1 in tree`: True"
```

This behavior contradicts with the documentation at https://python-jsonschema.readthedocs.io/en/latest/errors/#errortrees and is a bug.

mconsalvo95-tech:
Fixes #1328.

This replaces `ErrorTree`'s internal `defaultdict` child storage with an explicit dictionary so reading a valid child with no errors does not persist an empty subtree. Existing error paths are still populated during construction, and invalid indices still propagate the underlying instance lookup error.

Tests:
- `uv run --group test python -m unittest jsonschema.tests.test_exceptions`
- `PYTHONUTF8=1 JSON_SCHEMA_TEST_SUITE=json uv run --group test virtue jsonschema`
- `uv run --with ruff ruff check jsonschema\exceptions.py jsonschema\tests\test_exceptions.py`

### DIFF
--- a/jsonschema/exceptions.py
+++ b/jsonschema/exceptions.py
@@ -3,7 +3,7 @@ Validation errors, and some surrounding helpers.
 """
 from __future__ import annotations
 
-from collections import defaultdict, deque
+from collections import deque
 from pprint import pformat
 from textwrap import dedent, indent
 from typing import TYPE_CHECKING, Any, ClassVar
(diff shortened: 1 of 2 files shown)


mconsalvo95-tech:
Closing this after noticing that #1457 already addresses the same ErrorTree behavior. Sorry for the duplicate.

PR #1506, opened 2026-06-24
fix: reject NaN in numeric comparison keywords (minimum, maximum, exclusiveMinimum, exclusiveMaximum)

gaoflow:
NaN (`float("nan")`) currently passes all numeric comparison checks because Python's NaN comparisons always return `False`. This has two consequences:

1. **Schema validation bypass**: `check_schema` does not reject schemas containing NaN in numeric fields. For example, `{"multipleOf": float("nan")}` passes schema validation because the meta-schema's `exclusiveMinimum: 0` fails to reject NaN (`NaN <= 0` is `False`). This then causes `validate()` to crash with `ValueError: cannot convert float NaN to integer` inside the `multipleOf` handler.

2. **Instance validation bypass**: Instances containing NaN pass any `minimum`/`maximum`/`exclusiveMinimum`/`exclusiveMaximum` constraint, even though NaN is not a valid JSON number and cannot satisfy any ordering relationship.

The fix inverts the comparison logic to use the positive (non-negated) form:
- `not (instance > minimum)` instead of `instance <= minimum`
- `not (instance < maximum)` instead of `instance >= maximum`  
- `not (instance >= minimum)` instead of `instance < minimum`
- `not (instance <= maximum)` instead of `instance > maximum`

For normal numbers the result is identical. For NaN, all four positive comparisons return `False`, so the negation correctly yields `True` (rejecting NaN). The same fix is applied to the legacy draft3/draft4 equivalents in `_legacy_keywords.py`.

### DIFF
--- a/jsonschema/_keywords.py
+++ b/jsonschema/_keywords.py
@@ -213,7 +213,7 @@ def uniqueItems(validator, uI, instance, schema):
 
 def pattern(validator, patrn, instance, schema):
     if validator.is_type(instance, "string") and not re.search(
-        patrn, instance
+        patrn, instance,
     ):
         yield ValidationError(f"{instance!r} does not match {patrn!r}")
 
--- a/jsonschema/_keywords.py
+++ b/jsonschema/_keywords.py
@@ -21,7 +21,10 @@ def patternProperties(validator, patternProperties, instance, schema):
         for k, v in instance.items():
             if re.search(pattern, k):
                 yield from validator.descend(
-                    v, subschema, path=k, schema_path=pattern,
+                    v,
+                    subschema,
+                    path=k,
+                    schema_path=pattern,
                 )
 
 
@@ -128,7 +131,7 @@ def exclusiveMinimum(validator, minimum, instance, schema):
     if not validator.is_type(instance, "number"):
         return
 
-    if instance <= minimum:
+    if not (instance > minimum):
         yield ValidationError(
             f"{instance!r} is less than or equal to "
             f"the minimum of {minimum!r}",
@@ -139,7 +142,7 @@ def exclusiveMaximum(validator, maximum, instance, schema):
     if not validator.is_type(instance, "number"):
         return
 
-    if instance >= maximum:
+    if not (instance < maximum):
         yield ValidationError(
             f"{instance!r} is greater than or equal "
             f"to the maximum of {maximum!r}",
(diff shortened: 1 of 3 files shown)


Julian:
JSON Schema is defined on top of JSON, and JSON doesn't have NaN, so neither answer is right or wrong here really as it relates to JSON Schema behavior. Do you have anything in particular which you think motivates this change?

binggao1230:
The concrete case: NaN does reach validators in Python — stdlib `json.loads` accepts `NaN` by default, and plenty of people validate Python data directly. Today `float('nan')` satisfies `minimum`, `maximum`, `exclusiveMinimum` and `exclusiveMaximum` all at once, since every comparison is False — an unordered value passing every ordering constraint felt like the more surprising side of "unspecified". But you're right that the spec doesn't decide this; if you'd rather leave it undefined, I'm fine with closing.

Julian:
My issue with this kind of thing is typically that these kinds of incremental changes outside the spec still leave other behaviors unfixed, and worse, they communicate to people that outside of the spec is in some way "supported", in a way which means that future potentially harder-to-implement changes (including ones we haven't thought of) end up on the table. For example, what do we do about:

```
⊙  ~[jsonschema:python] -c 'import jsonschema, json; jsonschema.validate([["NaN"], ["NaN"]], {"uniqueItems": True})'                                                                                                                                                            julian@Airm
  File "<string>", line 1, in <module>
    import jsonschema, json; jsonschema.validate([["NaN"], ["NaN"]], {"uniqueItems": True})
                             ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/julian/.local/share/virtualenvs/jsonschema/lib/python3.14/site-packages/jsonschema/validators.py", line 1332, in validate
    raise error
jsonschema.exceptions.ValidationError: [['NaN'], ['NaN']] has non-unique elements

Failed validating 'uniqueItems' in schema:
    {'uniqueItems': True}

On instance:
    [['NaN'], ['NaN']]
```

The alternative (the "status-quo") is "it's undefined behavior anytime you have non-JSON values in your Python objects".

Julian:
Hope the above makes sense but happy to be convinced otherwise if you or anyone feels strongly and can help think through all the cases.

PR #1514, opened 2026-07-12
Fix uniq comparisons across hashability paths

Sanjays2402:
`uniq()` could report equal values as unique when one uses the structural-hash path and the other falls back to pairwise comparison (for example, `set` and `frozenset`). Compare fallback values across both paths so the result no longer depends on input order.

### DIFF
--- a/jsonschema/_utils.py
+++ b/jsonschema/_utils.py
@@ -217,13 +217,14 @@ def uniq(container):
     brute force when necessary.
     """
     seen_keys = set()
+    supported = []
     unsupported = []
 
     for element in container:
         key = _uniq_key(element)
 
         if key is _UNSUPPORTED:
-            for previous in unsupported:
+            for previous in [*supported, *unsupported]:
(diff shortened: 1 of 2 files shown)


Julian:
Hi there -- such values aren't possible to encounter in JSON as far as I can tell from your patch. Is there some actual user-facing bug you see via this path (i.e. a schema)? If so we'd at least want to test via that.

Sanjays2402:
@Julian You're right. I traced the complete `_uniq_key()` path: every strict JSON type is structurally supported recursively (objects have string keys; arrays recurse; scalars are hashable). `_UNSUPPORTED` only covers values outside the JSON data model, such as `set`, custom mappings with unhashable keys, or NaN. The current `set`/`frozenset` regression therefore cannot be reached from a schema-valid JSON instance.

I could not produce a schema-level user-facing case, so I'm withdrawing this PR rather than expanding `uniq()` for non-JSON Python values.

PR #1523, opened 2026-07-18
Report an unrepresentable duration amount as invalid instead of raising

### LINKED ISSUE #1511: is_duration raises an uncaught decimal.Overflow on a duration with a large exponent
The `duration` format check can raise `decimal.Overflow` instead of returning a result, so validating certain strings throws an uncaught exception rather than reporting the instance as invalid.

### Versions

- jsonschema 4.25.1
- isoduration 20.11.0 (the optional dependency `is_duration` delegates to)
- Python 3.9

### Reproduction

```python
from jsonschema import Draft202012Validator as V
from jsonschema import FormatChecker

v = V({"format": "duration"}, format_checker=FormatChecker())
list(v.iter_errors("P1E1000000D"))
```

raises:

```
decimal.Overflow: [<class 'decimal.Overflow'>]
```

### Cause

`_format.is_duration` calls `isoduration.parse_duration(instance)` and is registered with `raises=isoduration.DurationParsingException`. isoduration parses each component amount with `Decimal(...)`, and for an exponent past the decimal context's `Emax` (999999), `Decimal("1E1000000")` raises `decimal.Overflow`. That is not a subclass of `DurationParsingException`, so it propagates out of the format checker uncaught.

The boundary is exact: `P1E999999D` is accepted, `P1E1000000D` overflows. It is also reachable without an exponent, via a plain digit run longer than `Emax`.

### Expected

The check should report the instance as invalid (`P1E1000000D` is not a valid RFC 3339 Appendix A duration - the grammar uses `1*DIGIT`, with no exponent), not raise.

### Possible fix

Either broaden what `is_duration` catches (add `ArithmeticError` / `decimal.Overflow` alongside `DurationParsingException`, or wrap the `parse_duration` call), or treat any exception from `parse_duration` as "not a duration".

dchaudhari7177:
## Summary

Fixes #1511. The `duration` format check could raise `decimal.Overflow` out of `iter_errors` instead of reporting the instance as invalid:

```python
v = Draft202012Validator({"format": "duration"}, format_checker=FormatChecker())
list(v.iter_errors("P1E1000000D"))   # decimal.Overflow
```

`isoduration` parses each component amount with `Decimal`, and an amount past the decimal context raises `decimal.Overflow` — which is not a subclass of `DurationParsingException`, so it escaped the `raises=` handling entirely.

## Change

Catch `ArithmeticError` alongside `DurationParsingException`. `decimal.DecimalException` derives from `ArithmeticError`, so this covers the whole class of "the decimal context cannot represent this amount" without swallowing unrelated errors. Such an instance is invalid, not un-checkable.

This is the first option the issue suggests.

## Reach

The overflow isn't limited to the reported exponent form — verified all three now report invalid rather than raising:

| instance | why |
| --- | --- |
| `P1E1000000D` | exponent past `Emax` (the reported case) |
| `PT1E1000000S` | same, in a time component |
| `P` + `"9" * 1000001` + `D` | no exponent at all, just a digit run longer than `Emax` |

Valid durations are unaffected (`P1D`, `PT1S`, `P1Y2M3DT4H5M6S` still pass), and genuinely malformed input still reports invalid.

Confirmed the boundary the issue describes is intact: `P1E999999D` is still accepted. Note that's arguably wrong too — RFC 3339 Appendix A uses `1*DIGIT` with no exponent — but that's upstream `isoduration` behaviour covered by the existing `FIXME` above this function, so I left it alone rather than widening the fix.

## Tests

Two tests in `test_format.py`: one asserting all three unrepresentable amounts raise `FormatError` (parametrized via `subTest`), one asserting ordinary durations still validate. Both skip cleanly if `isoduration` isn't installed, matching how the format is optional.

Verified they catch the real bug by reverting only `_format.py`:

```
E    decimal.Overflow: [<class 'decimal.Overflow'>]
.venv\Lib\site-packages\isoduration\parser\parsing.py:69: Overflow
```

## Verification

- `pytest jsonschema/tests/test_format.py` → 10 passed, 6 subtests
- Full suite → **7695 passed, 819 skipped**
- `ruff check .` → all checks passed

I did not touch `CHANGELOG.rst`, since fixes to `_format.py` historically don't and the changelog looks like it's assembled at release time — happy to add an entry if you'd prefer it in the PR.

### DIFF
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -518,7 +518,11 @@ with suppress(ImportError):
     @_checks_drafts(
         draft201909="duration",
         draft202012="duration",
-        raises=isoduration.DurationParsingException,
+        # isoduration parses each component amount with Decimal, which raises
+        # (an ArithmeticError) rather than DurationParsingException when the
+        # amount exceeds the decimal context, e.g. "P1E1000000D" or a digit run
+        # longer than Emax. Such an instance is invalid, not un-checkable.
+        raises=(isoduration.DurationParsingException, ArithmeticError),
     )
     def is_duration(instance: object) -> bool:
         if not isinstance(instance, str):
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -518,11 +518,18 @@ with suppress(ImportError):
     @_checks_drafts(
         draft201909="duration",
         draft202012="duration",
-        # isoduration parses each component amount with Decimal, which raises
-        # (an ArithmeticError) rather than DurationParsingException when the
-        # amount exceeds the decimal context, e.g. "P1E1000000D" or a digit run
-        # longer than Emax. Such an instance is invalid, not un-checkable.
-        raises=(isoduration.DurationParsingException, ArithmeticError),
+        # isoduration parses each amount with Decimal, which rejects one it
+        # cannot represent without raising DurationParsingException. How it
+        # rejects depends on the implementation: the C one raises
+        # decimal.Overflow (an ArithmeticError), whereas the pure Python one
+        # builds an int first, so it raises ValueError once the digits exceed
+        # the interpreter's integer string conversion limit. Either way, the
+        # instance is invalid rather than un-checkable.
+        raises=(
+            isoduration.DurationParsingException,
+            ArithmeticError,
+            ValueError,
+        ),
     )
     def is_duration(instance: object) -> bool:
         if not isinstance(instance, str):
--- a/jsonschema/tests/test_format.py
+++ b/jsonschema/tests/test_format.py
@@ -80,6 +80,40 @@ class TestFormatChecker(TestCase):
         with self.assertRaises(FormatError):
             checker.check(instance="not-an-ipv4", format="ipv4")
 
+    def test_duration_with_unrepresentable_amount_is_invalid(self):
+        """
+        An amount past the decimal context is invalid, not un-checkable.
+
+        isoduration parses each component with Decimal, which raises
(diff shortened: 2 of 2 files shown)


dchaudhari7177:
CI caught that my first commit was incomplete, and the failure is worth recording here since it widens the bug beyond what the issue describes.

The `format` jobs failed on **pypy3.11 and 3.11 only**, on my `P` + `"9" * 1000001` + `D` case:

```
File ".../_pydecimal.py", line 576, in __new__
    self._int = str(int(intpart+fracpart))
ValueError: Exceeds the limit (4300) for integer string conversion: value has 1000001 digits
```

How `Decimal` rejects an amount it cannot represent depends on the implementation:

| implementation | rejects with | caught by `ArithmeticError`? |
| --- | --- | --- |
| C `_decimal` | `decimal.Overflow` | yes |
| pure Python `_pydecimal` (PyPy) | `ValueError` from the int conversion limit | **no** |

So the same class of instance escaped the format checker on PyPy exactly as the original `decimal.Overflow` did on CPython — the issue's suggested `ArithmeticError` only covers half of it.

Now catching `ValueError` too, which incidentally also subsumes `DurationParsingException` (isoduration derives it from `ValueError`); I kept that one listed explicitly so the primary case stays self-documenting and we don't depend on it remaining a `ValueError` subclass.

Verified locally against **both** implementations by forcing `_pydecimal` via `sys.modules["_decimal"] = None` — all cases report invalid, and valid durations still pass, under each.

PR #1540, opened 2026-08-01
Catch RecursionError in the regex format checker

### LINKED ISSUE #1538: regex format checker crashes with RecursionError on deeply nested patterns
The `regex` format checker is registered with `raises=re.error`, but `re.compile` raises `RecursionError` (not an `re.error` subclass) when the pattern nesting exceeds the C stack limit. This causes an uncaught exception to propagate from the validator instead of reporting the instance as invalid.

### Reproduction

```python
from jsonschema import Draft202012Validator, FormatChecker

v = Draft202012Validator({"format": "regex"}, format_checker=FormatChecker())
list(v.iter_errors("(" * 500))
# RecursionError: maximum recursion depth exceeded
```

### Expected

The instance should be reported as invalid (not a valid regex), not raise.

### Cause

`re.compile("(" * 500)` raises `RecursionError`, which is not a subclass of `re.error`, so `FormatChecker.check` does not convert it into a `FormatError`.

This is the same class of issue as the `OverflowError` case in #1526, but a distinct exception type.

### Versions

- jsonschema 4.26.0
- Python 3.13

UlikGames:
Fixes #1538

`re.compile` raises `RecursionError` rather than `re.error` when a pattern nests deeply enough to hit the C stack limit, so the checker's `raises=re.error` didn't cover it and the exception escaped the validator instead of being reported as an invalid instance.

Added `RecursionError` to the tuple, plus a test with `"(" * 500`. Same shape as the `OverflowError` case in #1526, just a different exception type.

Full test suite passes locally.

### DIFF
--- a/jsonschema/_format.py
+++ b/jsonschema/_format.py
@@ -413,7 +413,7 @@ with suppress(ImportError):
         return is_datetime("1970-01-01T" + instance)
 
 
-@_checks_drafts(name="regex", raises=re.error)
+@_checks_drafts(name="regex", raises=(re.error, RecursionError))
 def is_regex(instance: object) -> bool:
     if not isinstance(instance, str):
         return True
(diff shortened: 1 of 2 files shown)


UlikGames:
Just closing? Anything wrong with that or what?

PR #1556, opened 2026-08-18
fix: exclude tests and benchmarks from wheels

### LINKED ISSUE #1552: Avoid shipping benchmarks and tests in wheels
*jsonschema* currently bundles the benchmarks and tests inside the wheel distributions. These usually are of no additional value outside of the source distributions and thus just bloat the distribution. For this reasons, I recommend excluding them from future wheels.

This matches the official recommendation for Python packaging as well (see https://packaging.python.org/en/latest/discussions/package-formats/#what-is-a-wheel):

> Wheels are meant to contain exactly what is to be installed, and nothing more. In particular, wheels should never include tests and documentation, while sdists commonly do.

nightcityblade:
## What Problem This Solves

The wheel currently bundles the test suite and benchmarks even though neither is needed at runtime. This adds unnecessary files and size to installed distributions.

Fixes python-jsonschema/jsonschema#1552.

## Why This Change Was Made

The Hatch wheel target now excludes jsonschema/tests and jsonschema/benchmarks while the source distribution continues to include both. The build session also inspects the generated wheel and fails if either directory is bundled again.

## User Impact

Future wheels contain only runtime package files, producing a smaller, cleaner installation without changing library behavior or source distributions.

## Evidence

- Full test suite: 8,512 tests completed (7,809 passed, 703 skipped)
- Ruff: passed
- Distribution build, wheel-content assertion, Twine strict validation, and changelog validation: passed
- Verified the source distribution still includes tests and benchmarks


nightcityblade:
Superseded by #1557. I corrected the branch history and opened the clean two-file, one-commit version there.

PR #1561, opened 2026-08-20
Support swappable regular expression engines

### LINKED ISSUE #1142: extra regex features using `regex` instead of `re`
Hi there,

this is a kind of mix between question and feature request.

I'm interested in a regex feature called [possessive quantifiers](https://www.regular-expressions.info/possessive.html). In short, these allow you to use the quantifiers `*+`, `++` and `?+`. These act like their counterparts (`*`, `+`, `?`) except that if a match was found, they will not backtrack. as an example, the pattern `a?+a` will not match the string `a` because the regex engine doesn't backtrack "out of" the first `a?`. kind of like a super-greedy match

python supports these [starting with 3.11](https://docs.python.org/3/whatsnew/3.11.html#re), but some of us are stuck on lower versions (3.9 in my case).
Is there an easy way to make the `pattern` property (or the keys of `patternProperties`) use the [`regex`](https://pypi.org/project/regex/) library instead of the builtin [`re`](https://docs.python.org/3/library/re.html)? It's compatible with the builtin, but has some additional backported features like possessive quantifiers.

Ideally, I'd like some kind of optional argument where I can enable the 3rd party module and make python-jsonschema use that instead of the builtin. I don't think that should be the default, as you probably don't want "useless" extra dependencies.
Alternatively, patching this myself at runtime is probably possible, but it's not going to be pretty. If that's your recommendation, any hints about _where_ to patch it?
As a final option, [there are ways](https://stackoverflow.com/a/13577411/9145163) to mimic the behavior of possessive quantifiers with existing regex features, but that's not pretty either.

kurtmckee:
This PR introduces a "regular expression provider".

The feature allows alternate regular expression engines -- with different regex features, or with linear complexity guarantees, or other desirable attributes -- to be swapped out in place of Python's `re` parser.

Closes python-jsonschema/jsonschema#1142

> [!NOTE]
>
> I didn't see a way to neatly address the existing `is_regex()` format checker. It's hard-coded to use Python's `re` module and has no ability to access the validator instance on whose behalf it's running. This creates to distinct failure modes.
>
> First, it's possible for `Validator.check_schema()` to pass but for `validator.validate()` to fail with an uncaught exception when the configured regex engine rejects an unsupported regex feature. No `re.*` usage in jsonschema uses `try/except` clauses to guard against this, and I didn't introduce anything to address this failure mode.
>
> The other failure mode is that a regex feature supported by the configured regex engine would be rejected as invalid when `.check_schema()` is called.
>
> I chose to address this via documentation for the time being, as the work in this PR is sufficient to address the core security need that I have, as well as the core feature need seen in #1142.

# Example usage

This example uses Google's re2 package, which allows me to close a serious denial of service issue in a web service I maintain.

```python
import jsonschema
import re2


class Re2Provider(jsonschema.protocols.RegexProvider):
    raises = (re2.error,)
    compile = staticmethod(re2.compile)
    search = staticmethod(re2.search)


SafeValidator = jsonschema.validators.extend(
    jsonschema.validators.Draft202012Validator,
    regex_provider=Re2Provider(),
)


def is_regex(_instance: str) -> bool:
    if not isinstance(_instance, str):
        return True
    return bool(re2.compile(_instance))


SafeValidator.FORMAT_CHECKER.checkers["regex"] = (is_regex, Re2Provider.raises)

# SafeValidator is now configured for use.
```

### DIFF
--- a/docs/spelling-wordlist.txt
+++ b/docs/spelling-wordlist.txt
@@ -7,6 +7,8 @@ ValidationError
 # 0th, sigh...
 th
 amongst
+backreference
+backreferences
 callables
 # non-codeblocked cls from autoapi
 cls
@@ -26,6 +28,7 @@ iterable
 iteratively
 Javascript
 jsonschema
+lookahead
 majorly
 metaschema
 online
--- a/docs/validate.rst
+++ b/docs/validate.rst
@@ -299,6 +299,15 @@ The JSON Schema specification `recommends (but does not require) <https://json-s
 
 Given that there is no current library in Python capable of supporting the ECMA 262 dialect, the ``regex`` format will instead validate *Python* regular expressions, which are the ones used by this implementation for other keywords like :kw:`pattern` or :kw:`patternProperties`.
 
+.. note::
+
+    The ``regex`` format checker currently uses Python's :mod:`re` module
+    regardless of any custom `regex provider <regex-providers>` that may be
+    in use.  A pattern accepted by Python's :mod:`re` but rejected by a
+    custom engine — such as a backreference when using RE2 — will pass the
+    format check but fail at match time.  See :ref:`regex-providers` for
+    details.
+
 email
 ^^^^^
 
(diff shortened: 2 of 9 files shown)


kurtmckee:
The CI failure is due to an existing issue with Slack blocking the link-checker. It is unrelated to this PR.

_Update: The CI error was fixed in `main`; this branch was rebased and force-pushed._

Julian:
I'm just acknowledging I've seen this, but I honestly don't know when I'll fully get through reviewing it! My hope is we're talking weeks not months, so apologies, but just making sure it's clear I'm not totally ignoring it, I need to think through whether it's possible to improve upon a bit to feel totally comfortable with it!

kurtmckee:
No worries! Thanks for the acknowledgement. :+1:

PR #1572, opened 2026-09-01
Exclude benchmarks from wheel distributions

### LINKED ISSUE #1552: Avoid shipping benchmarks and tests in wheels

stefan6419846:
Closes #1552.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,6 +106,11 @@ Release Information
 path = "CHANGELOG.rst"
 pattern = "(^v.+?)\nv"
 
+[tool.hatch.build.targets.wheel]
+exclude = [
+  "/jsonschema/benchmarks",
+]
+
 [tool.coverage.html]
 show_contexts = true
 skip_covered = false


fallenmi (approved):
This matches the maintainer feedback in [#1552](https://github.com/python-jsonschema/jsonschema/issues/1552#issuecomment-5486238017): exclude the benchmark package from wheels while preserving the installed test suite. The new entry is scoped to the wheel target and uses the root-anchored Hatch pattern, so sdists and the remaining package files are unaffected: [exact lines 109–112](https://github.com/python-jsonschema/jsonschema/blob/21e897065c855b34da81bec6abfa94921bc3508c/pyproject.toml#L109-L112), [Hatch file-selection semantics](https://hatch.pypa.io/latest/config/build/#patterns).

The exact head is based directly on current main. All 85 check runs, 3 check suites, 2 workflows, and 2 legacy statuses are successful. Looks good to me.

Disclosure: I used OpenAI Codex and Anthropic Claude to assist with exact-ref, policy, CI, and source verification; I independently verified the cited change before submission.

